PR #24, opened 2015-01-13
pep8 the thing

NorthIsUp:
fixed indentation
dict.has_key -> "key" in dict
== (None|True|False) -> is (None|True|False)
!= (None|True|False) -> is not (None|True|False)

This diff should not change the behavior at all. I did this while trying to debug why it wasn't working for me.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -10,7 +10,7 @@ import urlparse
 import HTMLParser
 from websocket import create_connection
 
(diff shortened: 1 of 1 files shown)


NorthIsUp:
bump?

rawdigits:
this wouldn't cleanly merge after some other work. i ran autopep8, which seemed to be roughly the same. thanks for the suggestion!

PR #28, opened 2015-01-20
Fix errors when colorize nicks is off

jhawthorn:
Plugin wouldn't connect because `user.color_name` was undefined


rawdigits:
Oops! thanks!

PR #37, opened 2015-02-11
Help text improvements

nijotz:


rawdigits:
Tested/works well, and much better user messaging with this. Thanks!

Issue #16, opened 2014-11-06
Buffers.pl foreground color


Keithbsmiley:
The foreground color in the buffer list has the correct background color but incorrect selected foreground color. This can lead to unreadable names and channels. This also applies to the foreground color for unread channels.

pbrisbin:
I also notice that my slack buffers don't tab complete for the `/buffer` command. Is it possible these buffers aren't "real" in some sense?

rawdigits:
possibly? i think /buffers (name) uses the long name, which would be (yourdomain).slack.com.(channelname)

rawdigits:
(finally) fixed: https://github.com/rawdigits/wee-slack/pull/39 
thanks to @ishigoemon

Keithbsmiley:
Woo!

PR #39, opened 2015-02-18
Fix buffers.pl hotlist colors

ishigoemon:
This fixes the #16 problem for me. 

buffers.pl appears to depend on the color not being set in a channel name in order to properly set the hotlist_* colors. Previously the color was being set to default, this change stops setting any color if channels_not_on_current_server_color isn't set.


rawdigits:
Ahh thanks! This seems like a sensible fix.

PR #49, opened 2015-02-26
Display name instead of ID for bots

terlar:
Add Bot entity to be able to get the name

Right now all the bots are represented as their ID:s which is not very
useful, this fixes this.

This is my first python code so please let me know if something is
weird.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -126,6 +126,7 @@ class SlackServer(object):
         self.ws = None
         self.ws_hook = None
         self.users = SearchList()
+        self.bots = SearchList()
         self.channels = SearchList()
         self.connecting = False
         self.connected = False
@@ -151,6 +152,9 @@ class SlackServer(object):
     def add_user(self, user):
         self.users.append(user, user.get_aliases())
         users.append(user, user.get_aliases())
+
+    def add_bot(self, bot):
+        self.bots.append(bot)
 
     def add_channel(self, channel):
         self.channels.append(channel, channel.get_aliases())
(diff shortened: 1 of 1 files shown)


rawdigits:
Thanks! This looks good, but going to do some testing before a merge.

pbrisbin:
FWIW, this is working well for me

tfheen:
Hi @terlar, any chance you could rebase this? Master's moved ahead in the meantime so this no longer merges cleanly.

terlar:
Updated, I hope this will be the last rebase, I have tried to keep it up to date for quite some time.

wee_slack.py
```diff
@@ -1469,7 +1496,7 @@ def get_user(message_json, server):
     elif 'service_name' in message_json:
         name = u"-{}-".format(message_json["service_name"])
     elif 'bot_id' in message_json:
-        name = u"-{}-".format(message_json["bot_id"])
+        name = u"-{}-".format(server.bots.find(message_json["bot_id"]).name)
```
tfheen: This elif stanza should be moved earlier so we prefer the bot name if one exists.

Also, I'm wondering if we should use something more distinctive than `-$Name-`. @rawdigits suggested `:]`, which kinda looks like a robot. I tried using a unicode symbol, but that leads to various warnings, so I'm somewhat leaning towards the `:]` idea too.

tfheen:
This looks pretty good and seems to work fine in testing. I've that one comment in there, but it's a fairly minor thing. Opinions welcome. :-)

terlar:
I have done your suggested changes, perhaps one could let that be configurable how to display the bot name. But that could be a later improvement. I have been running this since February. But with the latest rebase and changes only a few times. If something is broken it should be related to the rebase but I don't think so.

tfheen:
This looks good to me. @rawdigits, ok to merge?

PR #50, opened 2015-03-12
linkify_text: Add support for "@group" and "@channel".

ctiwald:
These two commands are aliases for one another, and should work in both
private and public channels in Slack. Technically, there's also an
!everyone command, but that isn't implemented here.


rawdigits:
sorry for the delay - works! thanks!

PR #51, opened 2015-03-16
Hide messages with hidden: true

tfheen:
Some message, in particular file upload messages seem to be reposted
periodically, but with hidden: true.  Don't display those, to avoid
noise.


tfheen:
This is slightly simplistic in that it fails to do the right thing for edits to existing messages, so it should probably be refined a bit.

tfheen:
This is kinda superseded by #52, so closing.

PR #55, opened 2015-03-24
Fix up formatting of join/part/topic messages

tfheen:
Somewhat fixes https://github.com/rawdigits/wee-slack/issues/54, at least they're now shown as the right type of activity.


rawdigits:
I wonder if it might be useful to allow buffer_prnt to accept a set of weechat tags as a list arg?

tfheen:
Sure, I can implement that if you'd prefer that approach instead.

rawdigits:
thanks! i'm gonna test this out and merge if all is well.

rawdigits:
This is working well for me. thank you for the fixes!

Issue #64, opened 2015-04-24
Unicode error when using /me


bchretien:
To reproduce this, just type `/me à`:

```sh
python: stdout/stderr: Traceback (most recent call last):
python: stdout/stderr:   File "~/.weechat/python/autoload/wee_slack.py", line 737, in me_command_cb
python: stdout/stderr:     message = message.encode('utf-8')
python: stdout/stderr: UnicodeDecodeError: 'ascii' codec can't decode byte 0xc3 in position 17: ordinal not in range(128)
python: error in function "me_command_cb"
```

PR #65, opened 2015-04-25
Fix unicode error in me_command_cb.

bchretien:
This seems to solve #64 (at least on my system).


rawdigits:
thanks! unicode in /me - guess i don't do that enough :)

bchretien:
No problem ;)

PR #73, opened 2015-05-20
fixed that /slack (away|back) didn't update status

scharfie:
This fixes that `/slack back` and `/slack away` were not updating user presence status by adding support for `manual_presence_change` message type from slack.  Before the fix, this would appear in the slack-debug channel:

```
DEBUG: Function not implemented: manual_presence_change
{u'event_ts': u'REDACTED', 'myserver': 'REDACTED', u'type': u'manual_presence_change', u'presence': u'active'}
```


rawdigits:
this is a much cleaner way to do this. thanks for the fix!

PR #75, opened 2015-05-23
Colored own nick as weechat.color.chat_nick_self

iasoon:


pbrisbin:
FWIW, this is working well for me

rawdigits:
works well - sorry for the delay!

PR #76, opened 2015-05-29
switch to buffer on query

scharfie:
With this PR, `/query USERNAME` will automatically switch to the newly created buffer.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -753,7 +753,10 @@ def me_command_cb(data, current_buffer, args):
 def join_command_cb(data, current_buffer, args):
     if channels.find(current_buffer) or servers.find(current_buffer):
         channel = args.split()[1]
-        servers.find(current_domain_name()).channels.find(channel).open()
+        channel = servers.find(current_domain_name()).channels.find(channel)
+        channel.open()
+        w.buffer_set(channel.channel_buffer, "display", "1")
+        
         return w.WEECHAT_RC_OK_EAT
     else:
         return w.WEECHAT_RC_OK


pbrisbin:
FWIW, this is working well for me.

rawdigits:
This seems like a reasonable thing to do, but I'd prefer for there to be a configurable option for the old behavior. I've become used to opening buffers as a reminder to talk to someone, and often don't want to jump into a buffer and back out because it will move the read marker.

pbrisbin:
That's a fair point. I turned this on mostly b/c the newly opened query buffer gets stuck in "loading..." until I go to it. Maybe that's a separate bug worth reporting?

rawdigits:
The "loading..." thing is related to keeping redraws of the buffer list to a minimum. I've just fixed that one here (I think?): https://github.com/rawdigits/wee-slack/commit/5f5885185b10da6ea1ff4c791e17803a2e0ff22c

pbrisbin:
Oh nice. Maybe that'll also fix a "stuck" typing indicator. I'll test tomorrow.

JohnMaguire:
Given that default behavior in IRC seems to be switching to the window on `/query`, perhaps that should be the default w/ a boolean to switch to old behavior?

If that sounds reasonable, I can try to add the config option at some point this week to finish out this PR. (Thanks @rawdigits for wee-slack, and @scharfie  for the PR!)

rawdigits:
@JohnMaguire having the option seems reasonable to me, if you make a PR I'll make sure it gets in.

rawdigits:
Thanks! Did this, and made it optional. (defaults to on) - going to close this, as was quicker to just type the cpl of lines as opposed to merge dance.

rawdigits:
https://github.com/rawdigits/wee-slack/commit/027a2218948558825e67c7efe50b1226107aafa8

PR #88, opened 2015-06-25
Ignore deleted users when constructing nick list

tfheen:


rawdigits:
Looks good. thanks for adding deleted user handling!

PR #90, opened 2015-07-18
Use nick lists for grouping here/away

### LINKED ISSUE #72: wee-slack doesn't create nicklist groups
wee-slack doesn't create nick groups when creating channels, which causes things like  [Glowing Bear](https://github.com/glowing-bear/glowing-bear) to not show nicklists.

From #glowing-bear:

```
<+xt> it needs to be done "properly"
<+xt> I don't remember the exact specifics but I think nicks must be added to nick groups
<+xt> tell ryan to look at my matrix.lua script for how to do it :)
...
<Nei> I think G-B requires nicks to be in groups whereas weechat has no such requiremnt
<Nei> so it is sort of a bug in GB
<+xt> sounds right
<+xt> could be fixed
```

tfheen:
Instead of using + to denote users present, use nick groups and label
them Present and Away.  Also, add all non-deleted users to the server
buffer.

I believe this should fix #72

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -49,6 +49,9 @@ SLACK_API_TRANSLATOR = {
 
 }
 
+NICK_GROUP_HERE = "0|Here"
+NICK_GROUP_AWAY = "1|Away"
+
 def dbg(message, fout=False, main_buffer=False):
     message = "DEBUG: {}".format(message)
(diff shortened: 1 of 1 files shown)


rawdigits:
This is more like how slack presents here/away users in the webapp. If anyone prefers the old way, LMK and I'll make it an option.

PR #91, opened 2015-07-20
Improve nick completion

tfheen:
Previously, weechat would not complete @-prefixed nicks, fix that, and
make it so that if you complete on an already-full nick, wee-slack
adds a prefixing @, leading to a slack highlight.


rawdigits:
This looks good to me! thanks!

PR #92, opened 2015-07-22
Backend refactor, Performance improvements, Reactions

rawdigits:
This started as a branch to support reactions, but to do so properly required a number of backend changes. In no particular order:

* Enhanced SearchList class and made much more performant. Added a hashtable to each SearchList that acts as an index for much faster lookup of objects when doing `.find()`
* Added `Message` object. When storing a message in `channel.messages`, it should be this type.
* The global `message_cache` is now a list instead of a hash. this allows a simpler format where the message json is just stored directly to disk and replayed via process_message.
* Removed `Meta` class.
* Removed `SlackThing` superclass.
* Added some documentation.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -63,101 +63,33 @@ def dbg(message, fout=False, main_buffer=False):
         if slack_debug is not None:
             w.prnt(slack_debug, message)
 
-# hilarious, i know
-
-
-class Meta(list):
-    """
-    A collection of SearchLists that allow multiple accounts to coexist nicely
-    """
-    def __init__(self, attribute, search_list):
-        self.attribute = attribute
-        self.search_list = search_list
-
-    def __str__(self):
-        string = ''
-        for each in self.search_list.get_all(self.attribute):
-            string += "{} ".format(each)
-        return string
-
-    def __repr__(self):
-        self.search_list.get_all(self.attribute)
-
-    def __getitem__(self, index):
-        things = self.get_all()
-        return things[index]
-
-    def __iter__(self):
-        things = self.get_all()
-        for channel in things:
-            yield channel
-
-    def get_all(self):
-        items = []
-        items += self.search_list.get_all(self.attribute)
-        return items
-
-    def find(self, name):
-        items = self.search_list.find_deep(name, self.attribute)
-        items = [x for x in items if x is not None]
-        if len(items) == 1:
-            return items[0]
-        elif len(items) == 0:
-            pass
-        else:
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -60,97 +63,33 @@ def dbg(message, fout=False, main_buffer=False):
...
-
-    def get_all(self, attribute):
-        items = []
-        for child in self:
-            if child.__class__ == self.__class__:
-                items += child.get_all(attribute)
-            else:
-                items += (eval("child." + attribute))
-        return items
+            if "get_aliases" in dir(child):
```
tfheen: I like `hasattr(child, "get_aliases") and callable(child.get_aliases)` better. `dir` is more of a debugging tool

wee_slack.py
```diff
@@ -363,33 +301,58 @@ def buffer_input_cb(b, buffer, data):
...
-        self.last_read = float(last_read)
         self.last_received = None
+        self.messages = []
+        self.scrolling = False
         if active:
             self.create_buffer()
             self.attach_buffer()
             self.update_nicklist()
             self.set_topic(self.topic)
+            self.create_members_table()
```
tfheen: Shouldn't this happen before you call `update_nicklist`?

wee_slack.py
rawdigits: ah yep, much better!

wee_slack.py
rawdigits: it should indeed. good catch.

wee_slack.py
```diff
@@ -1602,10 +1673,30 @@ def url_processor_cb(data, command, return_code, out, err):
...
+                cache_file.write("{}\n".format(json.dumps(message.message_json)))
     return w.WEECHAT_RC_OK
 
+def cache_load():
+    global message_cache
+    try:
+        cache_file = open("{}/{}".format(WEECHAT_HOME, CACHE_NAME), 'r')
+        for line in cache_file:
+            message_cache.append(line)
+    except:
```
tfheen: Consider just catching the file not found exception?

wee_slack.py
```diff
@@ -1683,21 +1774,39 @@ def config_changed_cb(data, option, value):
...
+            channels.find(w.current_buffer()).unset_scrolling()
+    except:
+        pass
+    return w.WEECHAT_RC_OK
+
 # END Utility Methods
 
 # Main
 if __name__ == "__main__":
+
```
tfheen: Remove those blank lines

PR #93, opened 2015-07-28
add install instructions for FreeBSD

mrtazz:
just tried it out on one of my servers and those were the dependencies I had to install.

Thanks for an awesome weechat plugin!


rawdigits:
looks good - thanks!

PR #94, opened 2015-07-31
Unload cleanly

tfheen:
Let weechat know that we unloaded successfully and prevent a tiny bit
of whining on reload


rawdigits:
Ah yes - good catch.

PR #102, opened 2015-08-13
Centralise handling of unfurling a bit and unfurl replies

tfheen:
Previously, we would not unfurl responses to our own messages, which
meant we ended up with messages like "<@U12341|foo>: hi" instead of
"@foo: hi". This fixes that.


Hywan:
:+1:

PR #103, opened 2015-08-17
Add emoji support through dependency

terlar:
2015-08-17-164949_351x74_scrot

This can be enabled / disabled by changing the
`python.slack_extension.replace_emoji` variable in `plugins.conf`.

I have a solution that would not require a dependency:

```python
EMOJI_DICT = [
    (':+1:', u'\U0001F44D'),
    (':grinning:', u'\U0001f600'),
    (':loudspeaker:', u'\U0001f4e2')
]

for k, v in EMOJI_DICT:
    message = message.replace(k, v)
```

However I don't feel it is up to this project to maintain a list of
emoji-conversions, so I figured we could use a third party library.

However we might want to introduce a CUSTOM_EMOJI_DICT that the user can
configure for custom emojis their team have. This list would be
maintained in `plugins.conf`.

If this is too intrusive change to add a dependency perhaps you know some way to check for the available package and only enable the feature in that case.

What do you guys think?

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -11,6 +11,7 @@ import urllib
 import HTMLParser
 import sys
 import traceback
+import emoji
 from websocket import create_connection
 
 # hack to make tests possible.. better way?
@@ -583,6 +584,10 @@ class Channel(object):
                     message = user.name_regex.sub(
                         r'\1\2{}\3'.format(user.formatted_name() + w.color(chat_color)),
                         message)
+
+            if replace_emoji:
+                message = emoji.emojize(message, use_aliases=True)
+
             message = HTMLParser.HTMLParser().unescape(message)
             data = u"{}\t{}".format(name, message).encode('utf-8')
             w.prnt_date_tags(self.channel_buffer, time_int, tags, data)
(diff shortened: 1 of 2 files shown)


Hywan:
:+1: for emoji support but… what if it is another weechat plugin? And wee-slack could require this plugin? This could profit to other weechat “source” like IRC, XMPP etc.

wee_slack.py
```diff
@@ -583,6 +584,10 @@ def buffer_prnt(self, user='unknown_user', message='no message', time=0):
                     message = user.name_regex.sub(
                         r'\1\2{}\3'.format(user.formatted_name() + w.color(chat_color)),
                         message)
+
+            if replace_emoji:
+                message = emoji.emojize(message, use_aliases=True)
```
tfheen: You probably want to call this when you get a reply to a message you sent too, that currently doesn't go through buffer_prnt.

terlar: Weird because I get both messages received and sent in both channels and private messages translated with emojis. What other places are there?

terlar:
I was considering that but I don't know how often people would use `:smile:` inside regular IRC etc. People tend to use the old school text based smileys.

That is why I thought it belongs to this project since these `:x:` is the way slack communicates smileys/emojis.

kattrali:
Not sure this is the solution you are looking for, @terlar, however I made the dependency-free weechat plugin for doing just this: https://github.com/kattrali/weemoji

I went through the same thought process and added it to wee-slack directly originally, though my (now deleted fork) implementation increased wee-slack complexity a bit where it could instead be self contained.

terlar:
@kattrali That seems perfect, if you already did that work I don't see any reason to add this to weeslack directly.

So since such a plugin already exists I am closing this.

PR #106, opened 2015-08-29
unfurl unit test and cleanups

tfheen:


rawdigits:
Thanks again! Good start at unfurling this unfurling madness.

PR #107, opened 2015-08-30
support server aliases, so buffer names are shorter

justone:
First off, thank you so much for wee-slack.  It's fantastic.

After I switched to wee-slack, I found that buffers.pl (at least in my configuration) put the slack subdomain in the channel name.  I wanted to trim it down, so instead of this:

full_name

I wanted this:

shortened

I made this change initially back in March when I started using wee-slack and it's been useful to me, so I thought I would share.

Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -170,6 +170,11 @@ Turn off colorized nicks:
 /set plugins.var.python.slack_extension.colorize_nicks 0
 ```
 
+Set channel prefix to something other than my-slack-subdomain.slack.com (e.g. when using buffers.pl):
+```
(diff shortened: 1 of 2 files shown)


tfheen:
Looks useful to me, but I'll leave it to @rawdigits to review and merge. (I haven't tested it.)

rawdigits:
This looks good to me. I'm running it right now and will merge if all goes well.. Thanks!

rawdigits:
Tested out so well, I forgot it was there. Thanks again!

justone:
Fantastic, thanks for merging!

PR #108, opened 2015-09-18
Add upload capability to slack buffer

kenyonj:
This commit:
- Adds the ability to upload a file to the current slack buffer using
  the `/slack upload [file_path]` command

Why?
- This is a very useful feature for people that are trying to fully
  replace the slack web app

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,8 @@ A WeeChat native client for Slack.com. Provides supplemental features only avail
 
 Features
 --------
-  * **New** Emoji reactions!
+  * **New** Upload to slack capabilities!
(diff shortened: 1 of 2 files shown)


keith:
This is great, I'm going to checkout this PR to test it locally.

keith:
This has worked for me locally with text files and images. I haven't tried this but it would be nice if we could limit the completion to file paths. Once you type `~` it works but before that it tries to complete usernames and everything. I'm also not sure where that function is being set for this command.

rawdigits:
This worked well for me too. thanks very much @kenyonj

PR #109, opened 2015-09-23
Add dockerfile

schvin:
If interested, here is a basic Dockerfile that builds the dependencies and starts up weechat+wee-slack.


rawdigits:
Thanks very much for the offer, but this is rather out of scope for our little plugin.

Issue #74, opened 2015-05-20
Glowing Bear has characters in front of queries in buffer list


sklirg:
Regular channels looks right, and the queries looks right in weechat in a terminal, but using glowing-bear produces some strange combination of numbers and digits. 

_Could it be the color code for a nick (used in buffer list in weechat)?_

Screenshot with both active ('+') and inactive (' ') users
![screenshot 2015-05-20 21 31 12](https://cloud.githubusercontent.com/assets/5422571/7734630/bb82baf8-ff37-11e4-9ae5-b758f70c32c4.png)

sklirg:
Just to verify my assumption that this was the color codes I used a users username rather than the formatted username in buffer list;
https://github.com/rawdigits/wee-slack/blob/master/wee_slack.py#L655
I tried changing
`w.buffer_set(self.channel_buffer, "short_name", new_name)`
to 
`w.buffer_set(self.channel_buffer, "short_name", self.name)`

the result is that the messages will be shown as channels, but at least the numbers are gone.
![screenshot 2015-05-20 21 38 51](https://cloud.githubusercontent.com/assets/5422571/7734773/b210c798-ff38-11e4-83c8-e0910ce3f320.png)

bascht:
@sklirg relevant issue over at Glowing bear: glowing-bear/glowing-bear#615

bascht:
\o/ Thanks! :)

PR #110, opened 2015-09-24
Stop colouring channels and dmchannels

tfheen:
The weechat short_name property is a non-coloured string (according to
discussion on #weechat), so don't add any colours to it.

Fixes: #74


rawdigits:
Looks good!

Issue #56, opened 2015-03-27
feature request: support prefix same nick


scharfie:
Is it possible to support the `weechat.look.prefix_same_nick` setting?

rawdigits:
this is a good idea, and probably pretty easy to implement. would just need to track the ID of the last person to send a message and not print the ID until that value changes.

Nicd:
This is now broken in the current master. Nicknames are always repeated.

PR #113, opened 2015-10-13
Respect weechat.look.prefix_same_nick setting

tfheen:
When that is set, use it for repeated nick lines.

Doesn't handle bots properly yet, will be easier to do once
https://github.com/rawdigits/wee-slack/pull/49 is merged.

Fixes: #56


rawdigits:
quick conflict resolve and merge

Issue #25, opened 2015-01-13
Don't show activity in muted (slack setting) channels


rawdigits:

tfheen:
Seems like this isn't a property on the channel. Any idea where it's exposed?

PR #115, opened 2015-10-13
Handle muted of channels by not showing activity for them

tfheen:
If a channel is muted, don't show any activity for it, unless you're
explicitly mentioned.

Fixes: #25


tfheen:
This is fairly lightly tested, but please review and test.

rawdigits:
this is working well for me. will test some more before merging

PR #116, opened 2015-10-18
Only decode messages if they are not already unicode

trygveaa:
Some messages are already of type unicode when sent into buffer_prnt.
This makes message.decode fail with the following error:

    UnicodeEncodeError: 'ascii' codec can't encode character u'\xe5' in
    position 76: ordinal not in range(128)


calve:
Successfully tested on my side.

PR #121, opened 2015-10-27
Rewrite ``unfurl_refs``

calve:
Digging into weird ``UnicodeDecodeException`` bugs I experienced, I ended up rewritting the ``unfurl_refs()`` function. 

Well, now itworks®

### DIFF
--- a/_pytest/test_unfurl.py
+++ b/_pytest/test_unfurl.py
@@ -24,6 +24,9 @@ unfurl_map = [
     { "input": "url: <https://example.com|example with spaces> suffix",
       "output": "url: https://example.com (example with spaces) suffix",
     },
+    { "input": "<@U2147483697|@othernick> multiple unfurl <https://example.com|example with spaces>",
+      "output": "@othernick multiple unfurl https://example.com (example with spaces)",
+    },
     ]
 
 
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1678,7 +1678,7 @@ def unfurl_refs(text, ignore_alt_text=False):
     matches = re.findall(r"(<@(?:\S*)>)", text)
     for m in matches:
         # Replace them with human readable strings
-        text = text.replace(m, unfurl_ref(m[1:-1]))
+        text = text.replace(m, unfurl_ref(m[1:-1], ignore_alt_text))
     return text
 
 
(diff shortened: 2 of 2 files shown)


tfheen:
I don't think this will work, unfurl_refs also handles other cases.  See https://github.com/rawdigits/wee-slack/blob/master/_pytest/test_unfurl.py for some inputs and outputs.

calve:
I wasn't aware of this test. Could you document how am I supposed to run it ?

tfheen:
]] calve 

> I wasn't aware of this test. Could you document how am I supposed to run it ?

On my system (with python-pytest installed), just run `py.test` in the
root of the checkout.

-- 
Tollef Fog Heen

calve:
Indeed.

I updated the code to handle all tested patterns.

calve:
Any update on it ?

rawdigits:
I'll check it out.

rawdigits:
Hi calve - I tested this out with the profiler and it is sometimes the case, the regex version is actually slightly slower than the ugly brute force method currently used. I agree the code is MUCH better to look at, and am thinking of ways we can use your code without any slowdown.

The slowdown is mostly related to reading the cache re-parsing every message on startup. Forthcoming changes where we save a copy of the 'rendered' message for future use may mean we can use this code without any measurable impact to performance.

calve:
Rebased. Could you share me your performance test script ?

Also looks like the test suite is broken. Last commit which successfully build appears to be d6652ae

rawdigits:
Ya, need to fix testing!

I'll just leave the performance stuff here @calve :

```
import cProfile

p = cProfile.Profile()
p.enable()
counter = 1

#needed because otherwise no new data comes in
def profiler_write_cb(data, remaining):
    global counter, p
    p.dump_stats('/tmp/stats/asdf_{}'.format(counter))
    p.enable()
    counter += 1
    return w.WEECHAT_RC_OK


w.hook_timer(1000 * 5, 0, 0, "profiler_write_cb", “")
```

the above will write a file in `/tmp/stats/` every 5 seconds. it can be parsed with:

```
import pstats
p = pstats.Stats('asdf_1')
p.sort_stats('tottime').print_stats(20)
p.sort_stats('cumulative').print_stats(20)
```

calve:
I just pushed a rebased patch against the current master.

rawdigits:
Thanks for this @calve - makes the code much more readable. :smile:

PR #124, opened 2015-11-13
lots of refactoring for efficiency

rawdigits:
This (mostly) stops using buffer_redraw(), which re-rendered the entire buffer by clearing it and then passing every item in channel.messages through the process_message function.

Weechat uses posix seconds in message timestamps. Slack uses posix seconds + a unique index in case multiple messages come in the same second. In the case where there is only one message for a particular second, we just use hdata pointers to modify that line in the buffer. If we are unlucky and there are multiple messages at the same second that need modification, we fall back to redrawing the buffer.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1417,7 +1417,7 @@ def render_message(message_json):
         if "reactions" in message_json:
             text += create_reaction_string(message_json["reactions"])
 
-        w.prnt("", "rendered message")
+        #w.prnt("", "rendered message")
         return text
 
 
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -295,6 +295,8 @@ class SlackServer(object):
         for item in data["ims"]:
             if "last_read" not in item:
                 item["last_read"] = 0
+            if item["unread_count"] > 0:
+                item["is_open"] = True
             name = self.users.find(item["user"]).name
             self.add_channel(DmChannel(self, name, item["id"], item["is_open"], item["last_read"]))
         for item in data['self']['prefs']['muted_channels'].split(','):
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -222,7 +222,7 @@ class SlackServer(object):
                     w.unhook(self.ping_hook)
                     self.communication_counter = 0
                 self.ping_hook = w.hook_timer(1000 * 5, 0, 0, "slack_ping_cb", self.domain)
-                if len(self.users) and 0 or len(self.channels) == 0:
+                if len(self.users) == 0 or len(self.channels) == 0:
                     self.create_slack_mappings(login_data)
 
                 self.connected = True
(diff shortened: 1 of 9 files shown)


rawdigits:
DO NOT MERGE YET - breaks a couple of things when messages are more complicated. Need to break process_message out into something that renders messages and doesn't just buffer_prnt them anymore.

tfheen:
Can we store our identifier (whether it's timestamp or timestamp+id) in a tag and use that for keying off instead? It'd allow us to have one code path instead of two.

rawdigits:
Seems like a good idea. I'll investigate whether tags (or another hdata structure) can be used here

wee_slack.py
```diff
@@ -1386,142 +1415,148 @@ def create_reaction_string(reactions):
...
-#                    w.hdata_update(hdata_line_data, data, {"message": "{}{}".format(new_message, append)})
-#                    break
-#                elif not user and (int(date) == int(time)):
-#                    w.hdata_update(hdata_line_data, data, {"message": "{}{}".format(new_message, append)})
-#                else:
-#                    pass
-#            line = w.hdata_move(hdata_line, line, -1)
-#    return w.WEECHAT_RC_OK
-
+def modify_buffer_line(buffer, new_line, time):
```
rawdigits: Return to more efficient handling of changes: edits, deletes, emoji reactions, etc are now done by modifying lines instead of redrawing the buffer completely.

wee_slack.py
```diff
@@ -1386,142 +1415,148 @@ def create_reaction_string(reactions):
...
+                    w.hdata_update(struct_hdata_line_data, data, {"message": new_line})
+                    break
+                else:
+                    pass
+            #move backwards one line and try again - exit the while if you hit the end
+            line_pointer = w.hdata_move(struct_hdata_line, line_pointer, -1)
+    return w.WEECHAT_RC_OK
 
-def process_message(message_json, cache=True):
+def render_message(message_json, force=False):
```
rawdigits: the new render_message function, which takes over for process_message() in many instances.

wee_slack.py
```diff
@@ -1386,142 +1415,148 @@ def create_reaction_string(reactions):
...
+                channel.cache_message(message_json)
 
     except Exception:
         if channel and ("text" in message_json) and message_json['text'] is not None:
             channel.buffer_prnt('unknown', message_json['text'])
         dbg("cannot process message {}\n{}".format(message_json, traceback.format_exc()))
 
-def unwrap_message(message_json):
+
+def process_message_changed(message_json):
```
rawdigits: simplified process_message() - most important change is that we call render_message to do the manipulation of data and now cache the rendered version of the message so we don't need to call render_message again if we have already done it. Trades some memory for speed.

rawdigits:
This involves some pretty significant changes. Please report bugs in issues or via #wee-slack on freenode.

Caching support and fixes
More efficient buffer list updates (track changes without w.buffer_get calls
More efficient nicklist updates
More efficient typing updates (only apply to buffers that aren't hidden)
New (underscore) prepend for items added to json objects + test support
Reimplement line-by-line message changes
Remove unused stuff

PR #127, opened 2015-11-19
check for minimum weechat version + error

rawdigits:
Adds a version check right after script registration (must be registered to check the version).
Print an error if running version is < 1.3


tfheen:
Looks good to me.

PR #128, opened 2015-11-20
add code to do changes of topic.

rawdigits:
This allows a user to use either

`/slack topic [<#channel>] [topic]`
or
`/topic [<#channel>] [topic]` in a slack buffer to change the topic.


tfheen:
🚓

PR #142, opened 2015-12-04
always update the channels hashtable when we attach or detach a buffer

### LINKED ISSUE #141: DMs to wrong person.
As they say:
"There are only two hard things in Computer Science: cache invalidation and naming things."

This bug is related to the hashtable not being updated properly when a buffer goes away. Because many other things update the entire hashtable, this was infrequent as it required 1) a buffer to be created with a pointer that collides with a previous buffer's pointer and 2) no update_hashtable calls in the meantime.

rawdigits:
fixes #141


tfheen:
Looks reasonable: ✈

PR #147, opened 2015-12-21
Format more fields from attachments

tfheen:
This makes the formatting of attachments somewhat more verbose and
makes it more similar to the web client. Downside is that it takes
more vertical space.


tfheen:
Mostly putting this here so people can look and comment on it. Happy to take improvements and suggestions (both in form of code and textual descriptions)

PR #149, opened 2015-12-21
Allow dot in nicks

tfheen:


rawdigits:
:thumbsup: thanks @tfheen

PR #151, opened 2015-12-23
Cleanup message_json branch

keith:
This removes some duplicate fallback code

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1525,11 +1525,8 @@ def render_message(message_json, force=False):
 
         if "fallback" in message_json:
             text = message_json["fallback"]
-        elif "text" in message_json:
-            if message_json['text'] is not None:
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -1525,11 +1525,8 @@ def render_message(message_json, force=False):
 
         if "fallback" in message_json:
             text = message_json["fallback"]
-        elif "text" in message_json:
-            if message_json['text'] is not None:
-                text = message_json["text"]
-            else:
-                text = u""
+        elif message_json['text'] is not None:
```
tfheen: Don't you need `message_json.get('text')` here to avoid a possibly KeyError?

keith: Ah yep, thought this would handle that. My bad, thanks for catching it!

PR #152, opened 2015-12-23
Add mpdm channel

keith:
This adds a new channel type for mpdms which formats as `"#first,second,third"`

This includes the current user's username, which we may want to exclude.

Closes https://github.com/rawdigits/wee-slack/issues/133


rawdigits:
thanks @keith !

keith:
Thanks!

PR #156, opened 2016-01-13
Fix bug that fail to modify previous message if contain non-ascii characters

odknt:
Fail to modify previous message if contain non-ascii characters.

`s/test/てすと/`
```
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe3 in position 0: ordinal not in range(128)
```


rawdigits:
Tested out and works well. Thank you @odknt

PR #157, opened 2016-01-13
Change to add ' (edited)' only when rendering

odknt:
Current version is adds `(edited`) to tail when modifing previous message again.

```
test (edited)
```

`s/test/test/`

↓

```
test (edited) (edited)
```


rawdigits:
nice! this fixed an annoying little bug. thanks yet again @odknt :smile:

PR #160, opened 2016-01-27
Add unicode emojis support

aranega:
Currently, emojis are rendered as string: (:simple_smile: ...). When a user overly uses them, it becomes very annoying to read even small messages.
The idea here is to use unicode emojis when slack-emoji patterns are found. This option is enabled by default, but can be disabled this way:

    /set plugins.var.python.slack_extension.enable_emojis 0

This commit only propose a small dict of emojis, but it is easy to add new ones.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ In Development
 
 Dependencies
 ------------
-  * WeeChat 1.3+ http://weechat.org/ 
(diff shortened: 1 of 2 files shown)


aranega:
Closes this pull request because of this request I missed:
https://github.com/rawdigits/wee-slack/pull/103

Sorry!

PR #165, opened 2016-01-31
Pass system default ssl options to websocket.create_connection

### LINKED ISSUE #122: websocket connection error: cafile, capath and cadata cannot be all omitted
I got the following error when I try to load the slack plugin:

```
python: stdout/stderr: websocket connection error: cafile, capath and cadata cannot be all omitted
python: stdout/stderr: websocket connection error: cafile, capath and cadata cannot be all omitted
```

After this it just infinitely tries to reconnect to slack without success.

I think this might have been due to some upgrade of libraries but not sure. Any pointers in the right direction would be appreciated or of this is actually a bug in wee-slack.

daurnimator:
Fixes #122

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -14,6 +14,10 @@ import sys
 import traceback
 import collections
 from websocket import create_connection,WebSocketConnectionClosedException
+import ssl
+
+ssl_defaults = ssl.get_default_verify_paths()
+sslopt_ca_certs = {'ca_certs': ssl_defaults.cafile}
 
 # hack to make tests possible.. better way?
 try:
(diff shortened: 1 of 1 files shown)


bchretien:
:+1:  Works for me!

etu:
Works here as well, no issues at all. :+1:

KellerFuchs:
@daurnimator Ideally, that should honour weechat-wide settings like `weechat.network.gnutls_ca_file`.

daurnimator:
> Hi @daurnimator - Thank you for fixing this for lots of folks. my irc VM must be rather outdated, as ssl.get_default_verify_paths doesn't exist for me. I'm happy to implement your changes once I know how to make them backward compatible. Would a try: be best, or a version check? Thanks for any advice you can share!

I'm not python expert.
My guess would be to leave `sslopt_ca_certs` as an empty dictionary.
But I don't have an old VM to test with.

trygveaa:
This patch makes it compatible with older versions of python:

```
index 256fead..442fee9 100644
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -16,8 +16,11 @@ import collections
 from websocket import create_connection,WebSocketConnectionClosedException
 import ssl
 
-ssl_defaults = ssl.get_default_verify_paths()
-sslopt_ca_certs = {'ca_certs': ssl_defaults.cafile}
+if "get_default_verify_paths" in dir(ssl):
+    ssl_defaults = ssl.get_default_verify_paths()
+    sslopt_ca_certs = {'ca_certs': ssl_defaults.cafile}
+else:
+    sslopt_ca_certs = {}
 
 # hack to make tests possible.. better way?
 try:
```

PR #171, opened 2016-03-05
Pass system default ssl options to websocket.create_connection

tfheen:
Thanks to @daurnimator and @trygveaa for patches this is built on.

Fixes: #122 
Obsoletes: #165


rawdigits:
Thanks for fixing this for lots of people @trygveaa @daurnimator and @tfheen!

PR #172, opened 2016-03-05
Implement support for /msg user message

tfheen:


rawdigits:
Tested + works well for me.

megalithic:
ya'll typically update the version number somewhere? i see the version hasn't changed in the script itself.

rawdigits:
@megalithic yep! we are doing a PR merge round. i'm testing the changes now and will bump the version in a few mins.

scotte:
FYI @tfheen 6b53c203c50d44add1e33facb4159a4e6d573b2e has an issue due to this code:

```
        if channel:
             channel.open()
```

Since channel is None here, the open() method is guaranteed to fail. I noticed someone else already suggested a fix in PR #234 but thought I would comment here as well. Thanks!

Issue #158, opened 2016-01-13
/query doesn't work


adamgibbins:
I try to initiate a query with someone so I go to a channel on the Slack network, and /query person - where person is tab completed.  and I get this error:
```
20:19:25     weechat     | python: stdout/stderr: Traceback (most recent call last):
20:19:25     weechat     | python: stdout/stderr:   File "/home/aeg/.weechat/python/autoload/wee_slack.py", line 933, in join_command_cb
20:19:25     weechat     | python: stdout/stderr:     if command_talk(current_buffer, args.split()[1]):
20:19:25     weechat     | python: stdout/stderr:   File "/home/aeg/.weechat/python/autoload/wee_slack.py", line 1001, in command_talk
20:19:25     weechat     | python: stdout/stderr:     w.buffer_set(channel.channel_buffer, "display", "1")
20:19:25     weechat     | python: stdout/stderr: AttributeError: 'NoneType' object has no attribute 'channel_buffer'
20:19:25     weechat =!= | python: error in function "join_command_cb"
20:19:25     weechat =!= | irc: command "query" must be executed on irc buffer (server or channel)
```

bkmediagroup-josh:
I had the same issue when I had a server_alias set. With the alias /query and /slack text would only work from the server's buffer.

tfheen:
Good catch, this does indeed seem related to `server_alias`, I have a proposed fix in PR #173.  Can you test and see if it resolves the problem for you?

bkmediagroup-josh:
Sorry for the delay. I can verify the issue has been resolved. I will use that branch for a few days and report any problems.

PR #173, opened 2016-03-05
Make /slack talk and friends work when server_alias is set

tfheen:
We need to add the short name as an alias so current_domain_name is
able to figure out which server the current buffer belongs to.

Fixes: #158


rawdigits:
:thumbsup: works for me

PR #175, opened 2016-03-05
Support adding and removing reactions

### LINKED ISSUE #155: Add ability to react to a post
wee-slack shows when posts are reacted to, but does not allow you to react to them.
It'd be nice if it supported the same syntax as Slack itself, e.g. `+:+1:` to add a `+1` to the post above.

trygveaa:
A reaction can be added to the last message in the channel by writing

    +:reaction:

And removed by writing

    -:reaction:

Where reaction is the name of the reaction. This is the same syntax that
Slacks web interface uses.

Additionally, it supports sending a reaction to an earlier message by
prefixing the text with a number. The number is the number of the
message to add a reaction to, counting from the bottom. E.g. 1 is the
last message (same as without a number), 2 is the one before etc.

This fixes #155.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,6 +160,11 @@ Delete previous message:
 s///
 ```
 
+Add a reaction to the nth last message. The number can be omitted and defaults to the last message. The `+` can be replaced with a `-` to remove a reaction instead.
+```
+3+:smile:
+```
+
 Turn off colorized nicks:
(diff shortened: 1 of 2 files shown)


rawdigits:
can now send reactions! \o/ - thanks @trygveaa

PR #176, opened 2016-03-05
Show nicks for each reaction

trygveaa:
This adds an option of showing who added each reaction. Instead of e.g. `[:smile:1]`, it now shows `[:smile:(@trygveaa)]`.

In case someone thinks this takes up too much space, I made it optional with the config option `show_reaction_nicks`. That defaults to false, which makes the reactions appear just like before.


rawdigits:
Tested it out and happy with the result. Thanks!

PR #177, opened 2016-03-05
Return from update_nicklist if no channel_buffer

### LINKED ISSUE #150: `afk` referenced before assignment
I started noticing this new exception in the weechat buffer:

```
python: wrong arguments for function "nicklist_search_nick" (script: slack_extension)
python: wrong arguments for function "nicklist_remove_nick" (script: slack_extension)
python: stdout/stderr:   File "/Users/ksmiley/dev/dotfiles/weechat/python/autoload/wee_slack.py", line 1274, in slack_websocket_cb
python: stdout/stderr:     proc[function_name](message_json)
python: stdout/stderr:   File "/Users/ksmiley/dev/dotfiles/weechat/python/autoload/wee_slack.py", line 1332, in process_presence_change
python: stdout/stderr:     server.users.find(identifier).set_inactive()
python: stdout/stderr:   File "/Users/ksmiley/dev/dotfiles/weechat/python/autoload/wee_slack.py", line 803, in set_inactive
python: stdout/stderr:     channel.update_nicklist(self.identifier)
python: stdout/stderr:   File "/Users/ksmiley/dev/dotfiles/weechat/python/autoload/wee_slack.py", line 441, in update_nicklist
python: stdout/stderr:     w.nicklist_add_nick(self.channel_buffer, afk, user.name, user.color_name, "", "", 1)
python: stdout/stderr: UnboundLocalError: local variable 'afk' referenced before assignment
python: error in function "slack_websocket_cb"
```

Looks like that's getting hit here:

https://github.com/rawdigits/wee-slack/blob/master/wee_slack.py#L441

Which must mean that the first condition [here](https://github.com/rawdigits/wee-slack/blob/master/wee_slack.py#L422-L432) evaluates to false and `afk` never gets set.

trygveaa:
All code paths in this method requires channel_buffer, so we can return
in the start if it is falsy.

This fixes a bug where the here and afk variables were not set, but used
later.

This fixes #150.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -422,17 +422,19 @@ class Channel(object):
         self.server.channels.update_hashtable()
 
     def update_nicklist(self, user=None):
-        if self.channel_buffer:
-            w.buffer_set(self.channel_buffer, "nicklist", "1")
+        if not self.channel_buffer:
+            return
 
-            #create nicklists for the current channel if they don't exist
-            #if they do, use the existing pointer
-            here = w.nicklist_search_group(self.channel_buffer, '', NICK_GROUP_HERE)
-            if not here:
-                here = w.nicklist_add_group(self.channel_buffer, '', NICK_GROUP_HERE, "weechat.color.nicklist_group", 1)
-            afk = w.nicklist_search_group(self.channel_buffer, '', NICK_GROUP_AWAY)
-            if not afk:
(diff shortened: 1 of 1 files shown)


rawdigits:
LGTM :thumbsup:

(bye silly errors!)

Issue #143, opened 2015-12-09
NPE in slack_websocket_cb


Quentin-M:
Reporting an error:

```
python: error in function "slack_websocket_cb"
python: wrong arguments for function "buffer_set" (script: slack_extension)
python: stdout/stderr: TypeError: must be string, not None
python: error in function "slack_websocket_cb"
```

I will fix it if I find the time.

tfheen:
]] Quentin Machu 

> python: error in function "slack_websocket_cb"
> python: wrong arguments for function "buffer_set" (script: slack_extension)
> python: stdout/stderr: TypeError: must be string, not None
> python: error in function "slack_websocket_cb"

There should be a traceback too. Can you provide that?

-- 
Tollef Fog Heen
UNIX is user friendly, it's just picky about who its friends are

Quentin-M:
Nah, it did not appear in my buffer. I just enabled slack-debug. I will see if it appears again.

tfheen:
Hmm, are you running an up-to-date version? I don't see any buffer_set in the websocket callback function at all.

tfheen:
Hmm, I'm wondering if it's a duplicate of #138

etu:
This started happening when I did `/buffer merge 1` on the slack status buffers:

```
23:41:25                        weechat =!= | python: wrong arguments for function "nicklist_search_nick" (script: slack_extension)
23:41:25                        weechat =!= | python: wrong arguments for function "nicklist_remove_nick" (script: slack_extension)
23:41:25                        weechat     | python: stdout/stderr: Traceback (most recent call last):
23:41:25                        weechat     | python: stdout/stderr:   File "/home/etu/.weechat/python/autoload/wee_slack.py", line 1282, in slack_websocket_cb
23:41:25                        weechat     | python: stdout/stderr:     proc[function_name](message_json)
23:41:25                        weechat     | python: stdout/stderr:   File "/home/etu/.weechat/python/autoload/wee_slack.py", line 1338, in process_presence_change
23:41:25                        weechat     | python: stdout/stderr:     server.users.find(identifier).set_active()
23:41:25                        weechat     | python: stdout/stderr:   File "/home/etu/.weechat/python/autoload/wee_slack.py", line 801, in set_active
23:41:25                        weechat     | python: stdout/stderr:     channel.update_nicklist(self.identifier)
[...]

PR #178, opened 2016-03-05
Don't call buffer_set in im_marked if buffer is None

### LINKED ISSUE #138: Error closing buffer
If I `/close` a buffer representing a direct message, it closes successfully but I get the following error in the main window:

```
weechat   ⚠ │ python: wrong arguments for function "buffer_set" (script: slack_extension)
weechat     │ python: stdout/stderr: TypeError: must be string, not None
weechat   ⚠ │ python: error in function "slack_websocket_cb"
```

trygveaa:
process_im_marked may be called after process_im_close (e.g. when you do
/close in an im window). After process_im_close has run,
channel.channel_buffer is None, which makes the buffer_set fail.

This fixes #138 and #143.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1444,7 +1444,8 @@ def process_im_open(message_json):
 def process_im_marked(message_json):
     channel = channels.find(message_json["channel"])
     channel.mark_read(False)
-    w.buffer_set(channel.channel_buffer, "hotlist", "-1")
+    if channel.channel_buffer is not None:
+        w.buffer_set(channel.channel_buffer, "hotlist", "-1")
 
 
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -1444,7 +1444,8 @@ def process_im_open(message_json):
 def process_im_marked(message_json):
     channel = channels.find(message_json["channel"])
```
trygveaa: When doing this, I noticed that this method uses `channels` directly, while the two methods above does `server = servers.find...` and then uses `server.channels`. Should this method do the same?

(This comment is not really related to this PR though.)

PR #182, opened 2016-03-12
Coloring 'prefix_same_nick' prefix the same as the preceding username…

gilbertw1:
… it is replacing if 'colorize_nicks' is turned on.

Example:

Colorized Prefixes

IMO, this makes conversations a bit easier to follow when a prefix is defined.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ In Development
 
 Dependencies
 ------------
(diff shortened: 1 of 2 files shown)


gilbertw1:
Sorry accidentally included all three of my commits in this pull request. I'll close this one and re-open one specifically for this commit.

PR #183, opened 2016-03-12
Added option to colorize messages the same color as the sending nick …

gilbertw1:
…(defaults to off)

Added new option 'colorize_messages' that defaults to off, that when turned on will result in all messages being colored the same as the nick of the sender. Doesn't interfere with coloring usernames in the messages, and I've added corresponding doc to the README.

Example:
Colorized Messages

I missed this a lot when moving to wee-slack from using the irc gateway + a bouncer. This option effectively mimics the behavior of the 'colorize_lines.pl' plugin.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,7 +44,7 @@ In Development
 
 Dependencies
 ------------
-  * WeeChat 1.3+ http://weechat.org/ 
+  * WeeChat 1.3+ http://weechat.org/
   * websocket-client https://pypi.python.org/pypi/websocket-client/
 
 Setup
@@ -101,7 +101,7 @@ If you don't want to store your API token in plaintext you can use the secure fe
 ```
 
 ##### Optional: If you would like to connect to multiple groups, use the above command with multiple tokens separated by commas. (NO SPACES)
-    
+
 ```
 /set plugins.var.python.slack_extension.slack_api_token [token1],[token2],[token3]
 ```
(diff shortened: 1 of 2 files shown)


tfheen:
I think having this would be good, but as you mention, there's a plugin that does this already. Do you know what it'd take for that one to just be able to do what it wants directly, rather than effectively reimplementing it in weeslack?

gilbertw1:
I agree, that would me a much more ideal solution. The main reason I took this route is because wee-slack seems to do a lot on it's own (colorizing nicks, hard-coding mention highlights), and doesn't seem to play nice with number of plugins & and features (colorized_lines, custom highlight words, etc) built into wee-slack. I imagine this is mostly due to the fact that the buffer frequently changes and behaves in different ways than normal irc, and all data is being provided by an external source.

If you look at all three of my pull requests, they're essentially adding / fixing behavior that's already implemented elsewhere in weechat. It'd be great to hook into this functionality directly in wee-slack, and maybe implement some buffer related hooks (if they exist), so that other normal buffer altering plugins will play nicely.

I'm a very new weechat user and not familiar at all the how plugins work or tie into weechat in a technical sense, the above is just based on observation. All that being said, when I get a chance I'll try and dig a little deeper to see if there's a good way to hook into the functionality for 'colorized_lines.pl' and if possible, work on finding a way so that buffer related plugins can play nicely with wee-slack's buffers.

rawdigits:
This works fine for me. Merging - thanks!

PR #184, opened 2016-03-13
Added option 'slack_highlight_words' that when turned on will highlig…

gilbertw1:
…ht messages containing @channel and @everyone.

This setting defaults to off, however when it's on allows weechat to highlight any messages containing @channel or @everyone. This way the user can use something like highmon to capture these messages or a notification plugin to get desktop notifications.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -617,6 +617,10 @@ class Channel(object):
             set_read_marker = True
         elif message.find(self.server.nick.encode('utf-8')) > -1:
             tags = "notify_highlight"
+        elif slack_highlight_words and message.find("!channel".encode('utf-8')) > -1:
+            tags = "notify_highlight"
+        elif slack_highlight_words and message.find("!everyone".encode('utf-8')) > -1:
+            tags = "notify_highlight"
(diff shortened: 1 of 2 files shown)


rawdigits:
This has a merge conflict because of a change to on line. I played with it a bit, but it was preventing some messages from being displayed at all. Mind taking a look @gilbertw1 ?

wee_slack.py
```diff
@@ -617,6 +617,10 @@ def buffer_prnt(self, user='unknown_user', message='no message', time=0):
             set_read_marker = True
         elif message.find(self.server.nick.encode('utf-8')) > -1:
             tags = "notify_highlight"
+        elif slack_highlight_words and message.find("!channel".encode('utf-8')) > -1:
```
tfheen: slack_highlight_words isn't marked as a global (and we should avoid globals anyway). Can you just grab the setting from weechat near the top of the function?

rawdigits:
Wee-ng adds support for highlight words. If you set them in slack, they will be properly highlighted by weechat!

PR #194, opened 2016-03-25
Set the user to "--" for new user and new channel messages

tfheen:
This also fixes a bug where the server buffer was passed as the
message and the user was set to "New user joined: foo".


bendem:
This is customisable (`--` is just the weechat default), you might want to read the user config instead.

Issue #145, opened 2015-12-11
Trello issue comments don't render well: pretext


FauxFaux:
For the cheapo Trello integration, cards work well, but comments miss a load of context.  This bug affects the IRC gateway too, and they're also not fixing it there.  I guess `pretext` is a new feature.

New card:

    11/12:59:16 Trello :] │ New card "https://trello.com/c/afafa (test)" added to list "https://trello.com/b/afafa (Inbox)"

Comments are a bit broken, though:

    11/12:59:41 Trello :] │ test

In the web client, this renders as:

    New comment on card "<a href="https://trello.com/c/afafa">test</a>" by John Johninson
     | test

The `message_json` comes through as:

```json
{
  "attachments": [
    {
      "color": "8daec2",
      "text": "test",
      "fallback": "test",
      "id": 1,
      "pretext": "New comment on card \"<https://trello.com/c/afafa|test>\" by John Johninson"
    }
  ],
  "text": "",
  "ts": "1449838781.001147",
  "subtype": "bot_message",
  "mrkdwn": true,
  "_server": "afafa.slack.com",
  "type": "message",
  "channel": "afafa",
  "bot_id": "afafa"
}
```

Adding:

```python
if "pretext" in attachment:
    t.append(attachment["pretext"])
```

... to `unwrap_attachments` isn't very effective, as it causes the thing to read backwards; it looks like that message unpacker would need heavy adjustments?

    11/13:19:20 Trello :] │ Here's the comment: New comment on card "https://trello.com/c/awawawa (test)" by John Johninson

FauxFaux:
Aha, I was being dumb.  All the content is in the attachment, so you can actually do it with just `unwrap_attachment`.

tfheen:
I did some changes to formatting here recently, did that make the problem go away?

FauxFaux:
This is great; thanks.

Issue #200, opened 2016-03-26
Author name is not shown for attachments anymore


trygveaa:
After https://github.com/rawdigits/wee-slack/pull/147, the attachments doesn't show the author name anymore. @tfheen has a patch to fix this.

trygveaa:
Fixed by #201.

PR #201, opened 2016-03-27
Add author_name to attachment formatting

tfheen:
This should hopefully make some of the trello and other formatting a
bit better.

Fixes: #145


tfheen:
This probably fixes #200 too.

trygveaa:
I can confirm that this fixes #200.

PR #204, opened 2016-03-28
Set buffer localvars for channel and server

### LINKED ISSUE #188: Localvars for server and channel
Lots of scripts and logger settings and relay clients look at the buffer local variables server and channel.
Another useful localvar is short_name.

I think slack should set those on each buffer it creates.

trygveaa:
These are the same as what the irc plugin sets, and can be used in the
logger masks among other things.

This fixes #188.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -419,6 +419,8 @@ class Channel(object):
                 w.buffer_set(self.channel_buffer, "localvar_set_type", 'private')
             else:
                 w.buffer_set(self.channel_buffer, "localvar_set_type", 'channel')
+            w.buffer_set(self.channel_buffer, "localvar_set_channel", self.name)
+            w.buffer_set(self.channel_buffer, "localvar_set_server", self.server.server_buffer_name)
             w.buffer_set(self.channel_buffer, "short_name", self.name)
             buffer_list_update_next()
 
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -421,8 +421,11 @@ class Channel(object):
                 w.buffer_set(self.channel_buffer, "localvar_set_type", 'private')
             else:
                 w.buffer_set(self.channel_buffer, "localvar_set_type", 'channel')
+            if self.server.alias:
+                w.buffer_set(self.channel_buffer, "localvar_set_server", self.server.alias)
+            else:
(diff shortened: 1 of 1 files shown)


tfheen:
Looks fine to me. @rawdigits, ok to merge?

trygveaa:
Hm, with this, the server var is set to `<team>.slack.com`. If my suggestion in https://github.com/rawdigits/wee-slack/issues/144#issuecomment-202512332 is done, it would be set to `slack.<team>`.

Maybe it should be set to only the team name instead? This would make it more like to the irc plugin, where it is only set to the server name (e.g. "EFNet").

tfheen:
Just the team name would be fine, it'll be namespaced by the plugin name anyway, won't it? (For logging.)

trygveaa:
Yeah, it makes more sense for logging as well. That way, you can choose if you want slack as part of the file name or not. I changed it.

wee_slack.py
```diff
@@ -419,6 +421,8 @@ def create_buffer(self):
                 w.buffer_set(self.channel_buffer, "localvar_set_type", 'private')
             else:
                 w.buffer_set(self.channel_buffer, "localvar_set_type", 'channel')
+            w.buffer_set(self.channel_buffer, "localvar_set_channel", self.name)
+            w.buffer_set(self.channel_buffer, "localvar_set_server", self.server.team)
```
trygveaa: Should this take the server alias into account?

tfheen: ]] Trygve Aaberge 

> In wee_slack.py:
> 
> > @@ -419,6 +421,8 @@ def create_buffer(self):
> >                  w.buffer_set(self.channel_buffer, "localvar_set_type", 'private')
> >              else:
> >                  w.buffer_set(self.channel_buffer, "localvar_set_type", 'channel')
> > +            w.buffer_set(self.channel_buffer, "localvar_set_channel", self.name)
> > +            w.buffer_set(self.channel_buffer, "localvar_set_server", self.server.team)
> 
> Should this take the server alias into account?

Yes, I think so makes sense.

-- 
Tollef Fog Heen

trygveaa: Fixed.

PR #205, opened 2016-03-29
Handle bots using the RTM API too

tfheen:
Some bots, such as lita (https://github.com/litaio/lita-slack) is
using the RTM API.  Those show up as regular users, but have is_bot
set on their user object.  Pick up on that when formatting lines.

Thanks to Whoop on IRC for pointing this out.


rawdigits:
Good catch - thanks @tfheen, the fix works for me. Also surprised to find out how many of my coworkers are bots.

PR #210, opened 2016-04-09
Print out the name of the user when enabling/disabling integrations

tfheen:
Previous:

    pagerduty :] | disabled an integration in this channel: https://…

Now:

    pagerduty :] | foo disabled an integration in this channel: https://…


rawdigits:
Handy! I didn't realize these were missing. Thanks @tfheen

Issue #137, opened 2015-11-28
Exception when using /topic on an IRC channel without one


JohnMaguire:
```
 python: stdout/stderr: Traceback (most recent call last):
 python: stdout/stderr:   File "/home/john/.weechat/python/autoload/wee_slack.py", line 1201, in topic_command_cb
 python: stdout/stderr:     if command_topic(current_buffer, args.split(None, 1)[1]):
 python: stdout/stderr: IndexError: list index out of range
 python: error in function "topic_command_cb"
```

rawdigits:
should probably use the decorator to check we are in a slack buffer. thanks!

tfheen:
We can't just use the decorator, since we want to pass through to the built-in topic if we're not in a slack buffer. We should probably get a new one and wrap /join and friends in the same.

PR #212, opened 2016-04-12
Ignore command not executed in a slack buffer

### LINKED ISSUE #191: Error with msg hook, when using /say or /msg *
I have some aliases and scripts that use the /say command, that use /msg *.
From /help msg:

`target: nick or channel (may be mask, '*' = current channel)`
My theory is that slack hooks /msg and parses it wrongly.

Error looks like this:

```

1242             weechat              │ This command must be used in a slack buffer
1242             weechat          =!= │ python: function "msg_command_cb" must return a
                                      │ valid value                                      
1242             weechat          =!= │ python: error in function "msg_command_cb"
```

bendem:
Commands query, join, part, leave, topic are not specific to slack.
Using them in an irc context is valid and should not trigger errors
or messages.

Fixes #191 and #137

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -977,12 +977,25 @@ def me_command_cb(data, current_buffer, args):
     return w.WEECHAT_RC_OK
 
 
+# Wrap command_ functions that require they be performed in a slack buffer
+def slack_buffer_required(f):
+    @wraps(f)
+    def wrapper(current_buffer, *args, **kwargs):
+        server = servers.find(current_domain_name())
+        if not server:
+            return w.WEECHAT_RC_OK
+        return f(current_buffer, *args, **kwargs)
+    return wrapper
+
+
+@slack_buffer_required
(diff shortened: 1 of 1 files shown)


tfheen:
Thanks for the PR!

This change means that we lose the error message for the commands which are specific to slack, though.  Can you split the functionality so that we keep that for those which are specific to slack, but adds your functionality to query, join, part, etc?

bendem:
I'm not sure I understand, I only added `@slack_buffer_required` to the handlers of the commands that are **not** specific to slack.

tfheen:
No, you also changed the implementation of `@slack_buffer_required` so that it no longer gives an error message.

This means that if you do `/slack upload` in a non-slack buffer today, you'd get an error. With your patch, it silently fails.

tfheen:
I ended up redoing this, so closing this PR. Thanks for the inspiration, though. :-)

PR #214, opened 2016-04-13
Add @slack_buffer_required to command_join

terminalmage:
This fixes a traceback when command_join is run outside of a slack
buffer.


rawdigits:
we can't do it this way, because it will break `/join` for normal irc or non slack things. we're going to fix the root issue via #212. thanks!

terminalmage:
Awesome, thanks.

PR #215, opened 2016-04-13
Remove unnecessary channels.join API request

### LINKED ISSUE #190: users.find(self.name) returns None on channel join
users.find(self.name) appears to return None when joining channels.

Updated: Originally I incorrectly thought this was preventing me from joining channels. My channels get joined, they just show up as inactive in weechat, for some reason. This is noticeable if buffers.log.mark_inactive is set, which puts parenthesis around inactive channels (such as those that have been /part'ed), making all slack channels appear to be parted.

Version info:

```
WeeChat 1.4 [compiled on Jan 29 2016 10:02:56]
slack_extension.py 0.99.9 (Extends weechat for typing notification/search/etc on slack.com)
```

Errors at script load:

```
python: wrong arguments for function "buffer_set" (script: slack_extension)
python: stdout/stderr: TypeError: must be string, not None
python: error in function "slack_websocket_cb"
```

Errors at channel join: 

```
python: stdout/stderr:   File "/path/to/.weechat/python/autoload/wee_slack.py", line 961, in slack_command_cb
python: stdout/stderr:     command = cmds[function_name](current_buffer, args)
python: stdout/stderr:   File "/path/to/.weechat/python/autoload/wee_slack.py", line 1081, in command_join
python: stdout/stderr:     servers.find(domain).channels.find(args).open()
python: stdout/stderr:   File "/path/to/.weechat/python/autoload/wee_slack.py", line 554, in open
python: stdout/stderr:     async_slack_api_request(self.server.domain, self.server.token, SLACK_API_TRANSLATOR[self.type]["join"], {"user": users.find(self.name).identifier})
python: stdout/stderr: AttributeError: 'NoneType' object has no attribute 'identifier'
python: error in function "slack_command_cb"
```

terminalmage:
There are two problems with this API request:
1. According to https://api.slack.com/methods/channels.join, "user" is
   not a valid argument.
2. The hash table lookup in the removed line of code is destined to fail
   anyway, because `self.name` here refers to the channel name, which
   is not present in the `users` **SearchList()** instance as it maps
   usernames to the Slack API user IDs. Indeed, this causes the call to
   `users.find(self.name)` to return `None`, causing an
   AttributeError to be raised.

Fixes #190.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -553,7 +553,6 @@ class Channel(object):
         if update_remote:
             if "join" in SLACK_API_TRANSLATOR[self.type]:
                 async_slack_api_request(self.server.domain, self.server.token, SLACK_API_TRANSLATOR[self.type]["join"], {"name": self.name.lstrip("#")})
-                async_slack_api_request(self.server.domain, self.server.token, SLACK_API_TRANSLATOR[self.type]["join"], {"user": users.find(self.name).identifier})
 
     def close(self, update_remote=True):
         #remove from cache so messages don't reappear when reconnecting


nicksergeant:
This also fixes #224.

PR #217, opened 2016-04-13
Fix two tracebacks in command_talk

terminalmage:
Both tracebacks occur when you attempt to /slack talk a username for which no
buffer currently exists, as command_talk attempts to set the buffer for a
nonexistant channel. See commit messages for further details.


rawdigits:
Thanks @terminalmage - i think the fix in #234 covered this case too.

PR #219, opened 2016-04-13
Make message editing use actual regexes

### LINKED ISSUE #197: Feature request: edit syntax should actually be regex
right now s/blah/new/ is just a string replace and is /g - it should be real regex.

terminalmage:
Resolves #197

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -340,7 +340,7 @@ class SlackServer(object):
 def buffer_input_cb(b, buffer, data):
     channel = channels.find(buffer)
     reaction = re.match("(\d*)(\+|-):(.*):", data)
-    if not reaction and not data.startswith('s/'):
+    if not reaction and not data.startswith('s/') and not data.startswith('s#'):
         channel.send_message(data)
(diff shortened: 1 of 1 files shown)


terminalmage:
@rawdigits Do you think it might be better to hide this behind a weechat option (e.g. ``/set python.slack_extension.edit_with_regex 1``)? If so, what should be the default behavior, ``str.replace()`` or ``re.sub()``?

This needs documentation in **README.md** regardless, so whatever the decision I'll need to push an additional commit before this would be merged.

rawdigits:
thanks! we are discussing this and whether it needs to be behind an option. perhaps it is fine to make it the new default. we'll test and merge it soon.

terminalmage:
Great! We should also document that the updated functionality allows for a ``#`` to be used as the regex delimiter.

trygveaa:
Any progress on this? Would be nice to have the functionality.

To voice my opinion on the question above, if you want to allow replacement without regex I think it should rather be as another syntax, instead of adding an option for it.

PR #221, opened 2016-04-20
Encode topics in utf-8, not ascii

bendem:
Utf-8 is already used for names and messages; it's now also the case for topics.


rawdigits:
thanks! merging.

PR #222, opened 2016-04-20
Fix error when sending messages to server query

bendem:
```
python: stdout/stderr:   File "/home/bendem/.weechat/python/autoload/wee_slack.py", line 344, in buffer_input_cb
python: stdout/stderr:     channel.send_message(data)
python: stdout/stderr: AttributeError: 'NoneType' object has no attribute 'send_message'
python: error in function "buffer_input_cb"
```


rawdigits:
lgtm :thumbsup: thanks!

PR #225, opened 2016-04-30
Support passing additional tags to enable notification_center.py

megalithic:
This supports adding additional tags to enable notification_center.py to push said tagged notifiers to osx' notification center.

In response to: https://github.com/rawdigits/wee-slack/issues/211

It feels like there is/could/should be a better way to do this? Could just be my complete lack of full knowledge around this plugin.

I originally had settings to enable/disable highlighting, as well as customizing the highlights, beyond the ones you set in weechat's `irc.conf`. This however, proved to be useless as the notification_center.py plugin simply hooks into those existing highlights. Interestingly enough, though; `@here`/`!here` mentions in the channel WILL notify, perhaps that was already built in from the start and I missed that.

At any rate, please let me know where changes need/should be made.

Thanks for this awesome plugin. Slack was already pretty dang awesome, and this plugin simply puts it over the top. :)

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -427,7 +427,7 @@ class Channel(object):
         if channel_buffer != main_weechat_buffer:
             self.channel_buffer = channel_buffer
             w.buffer_set(self.channel_buffer, "localvar_set_nick", self.server.nick)
-#            w.buffer_set(self.channel_buffer, "highlight_words", self.server.nick)
+            w.buffer_set(self.channel_buffer, "highlight_words", self.server.nick)
         else:
             self.channel_buffer = None
         channels.update_hashtable()
@@ -622,13 +622,13 @@ class Channel(object):
         elif message.find(self.server.nick.encode('utf-8')) > -1:
             tags = ",notify_highlight,log1"
         elif user != self.server.nick and self.name in self.server.users:
-            tags = ",notify_private,notify_message,log1"
+            tags = ",notify_private,notify_message,log1,irc_privmsg"
         elif self.muted:
             tags = ",no_highlight,notify_none,logger_backlog_end"
         elif user in [x.strip() for x in w.prefix("join"), w.prefix("quit")]:
             tags = ",irc_smart_filter"
         else:
-            tags = ",notify_message,log1"
+            tags = ",notify_message,log1,irc_privmsg"
         #don't write these to local log files
         #tags += ",no_log"
         time_int = int(time_float)


wee_slack.py
```diff
@@ -2238,8 +2238,7 @@ def scrolled_cb(signal, sig_type, data):
             if not w.config_get_plugin('switch_buffer_on_join'):
                 w.config_set_plugin('switch_buffer_on_join', "1")
 
-            if w.config_get_plugin('channels_not_on_current_server_color'):
-                w.config_option_unset('channels_not_on_current_server_color')
+            w.config_option_unset('channels_not_on_current_server_color')
```
tfheen: Why this bit? It's used for cleaning up an obsolete setting and should not be removed unless it causes problems (and then in a separate PR).

wee_slack.py
```diff
@@ -622,13 +622,13 @@ def buffer_prnt(self, user='unknown_user', message='no message', time=0):
         elif message.find(self.server.nick.encode('utf-8')) > -1:
             tags = ",notify_highlight,log1"
         elif user != self.server.nick and self.name in self.server.users:
-            tags = ",notify_private,notify_message,log1"
+            tags = ",notify_private,notify_message,log1,notify_highlight,irc_privmsg"
```
tfheen: Is the notify_highlight needed? My understanding is notify_private is a superset of that.

megalithic:
@tfheen thank you sir, resolved those 2 items.

wee_slack.py
```diff
@@ -427,7 +427,7 @@ def attach_buffer(self):
         if channel_buffer != main_weechat_buffer:
             self.channel_buffer = channel_buffer
             w.buffer_set(self.channel_buffer, "localvar_set_nick", self.server.nick)
-#            w.buffer_set(self.channel_buffer, "highlight_words", self.server.nick)
+            w.buffer_set(self.channel_buffer, "highlight_words", self.server.nick)
```
tfheen: This seems to be a left-over.

tfheen:
Added a comment. In addition, can you squash all the commits? I think a single commit saying you add those tags is fine, we don't need the entire evolution of the patch series.

megalithic:
@tfheen i've resolved all the issues we discussed; however i left uncommented, that one `w.buffer_set` line; without it, i was not getting notifications. might you be willing to try it on your end (that is, if you are on OS X and use the notification_center.py plugin)?

wee_slack.py
megalithic: i'm not 100% sure why, but, leaving this commented resulted in the notifications not working. so i've left it uncommented.

megalithic:
just checking in @rawdigits / @tfheen; this is ready to go.

PR #226, opened 2016-05-05
Added process_group_join, process_group_leave and process_group_topic

irconan:
that pass to their channel equivalents


rawdigits:
I wonder if the api changed or we just never implemented these?

either way :thumbsup: lgtm

PR #233, opened 2016-06-12
Use a regex for highlight detection. Fixes #220

### LINKED ISSUE #220: Highlighting ignores word boundaries
Wee-slack highlights me when my nick (Na) is part of another word (eg binary, native, finally).

As far as I can tell, the code responsible for this is in Channel.buffer_prnt on [line 622](https://github.com/rawdigits/wee-slack/blob/master/wee_slack.py#L622)

``` python
            tags = ",notify_highlight,log1"
```

Being able to provide a regex or list of words would probably be ideal, but requiring the nick to be its own word would work sufficiently well.

Amagineer:


rawdigits:
thanks! i like this idea, but i want to test performance on startup since every message is processed and this could get worse with regex.

rawdigits:
hi @Amagineer - i like this, but I think doing a re.compile every time we do a buffer.prnt is inefficient. We should probably do it elsewhere.

torhve:
The slack extension already sets this `w.buffer_set(self.channel_buffer, "highlight_words", self.server.nick)`, which means weechat already looks for your nick as a word to highlight, which means I don't see a reason for the original `.find` and I also do not see the reason for a new regex to look for words to highlight.

userrl:
Just a note, the `@?` in the regex might as well not be there.  The reason this works against `'a @nick b'` is because `'\bnick\b'` is getting matched (the initial `\b` is matching the change from the `\W` character `@` to the `\w` first character in the nick).  If you remove the question mark, the regex won't work because `\b` doesn't match the transition from space (a `\W` char) to `@` (another `\W` char) in `' @'`.

Similarly, the `:?` in `:?\b` never gets matched for the same reason (`': '` doesn't match `\b`).  The regex might as well just be simplified to `r'\b{}\b'.format(self.server.nick.encode('utf-8'))` to save some CPU cycles.

Finally, there's no advantage to recompiling the regex every time this function is called and then using it once a few lines later.  We might as well just use re.search(), or else make the regex a `self.mynick_regex` variable that's set at connect / nick-change (or better yet, `self.myuser` so we can re-use the `User.name_regex` code).

That said, I totally agree with @torhve that we should re-use weechat's existing nick-highlighting mechanisms - they're a lot more intelligent than anything we roll ourselves (see the fun that happens when `colorize_nicks` is enabled and you have people using single-character nicks - "I'**m** going to **a** party.").

rawdigits:
This looks to be outdated. Closing

PR #234, opened 2016-06-13
Open new queries

brimstone:
I was unable to open queries to new people until this patch. I've been using it
for a few weeks and it seems stable.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1067,16 +1067,16 @@ def command_talk(current_buffer, args):
     server = servers.find(current_domain_name())
     if server:
         channel = server.channels.find(args)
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -1067,16 +1067,16 @@ def command_talk(current_buffer, args):
     server = servers.find(current_domain_name())
     if server:
         channel = server.channels.find(args)
-        if not channel:
+        if channel is not None:
```
tfheen: Can we avoid double negation here? It's much easier to read if you swap the two arms of the if around and drop the `not`.

rawdigits:
thanks @brimstone !

PR #241, opened 2016-06-27
Fix tracebacks

jkoelker:
The two tracebacks I've been seeing pretty much every day since the last update.  I've been running this for the past 5 days and it seems to work. The only one I'm not so sure about is the ramifications of commit 051a0f5, but I haven't seen any problems.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -15,7 +15,7 @@ import traceback
 import collections
 import ssl
 
-from websocket import create_connection,WebSocketConnectionClosedException
+from websocket import create_connection, WebSocketConnectionClosedException
 
 # hack to make tests possible.. better way?
 try:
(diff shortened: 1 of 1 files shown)


tfheen:
Thanks for the fixes!

I've pulled in the two latter fixes here. We want to move closer to pep8, but some bits of pep8 are really outdated (like the max 80 character limit), so we need to provide default settings for pep8/flake so it stops warning about those.

jkoelker:
No problem.

Yea with PEP8 the 80 column thing is usually the contention. I mainly did it so syntastic would show me errors that I introduced rather than existing. Although I'm in the 80 column camp so vertical splits in vim don't line wrap or go off the buffer view and philosophically I think it leads to having to think about code organization (the real downside being getting long strings to fit); I totally get that others have different workflows or smaller fonts.

tfheen:
Yeah, I don't think unlimited line length is reasonable, but 80 characters feels a tad limiting. 132 is a nice, somewhat bigger and tradititional number. :-)

Again, thanks for the patch, much appreciated.

PR #245, opened 2016-07-15
add oauth code

rawdigits:
This adds the `/command register` action, which lets you register for an oauth token.


tfheen:
ITYM`/slack register`. Doc update too, maybe?

rawdigits:
good call on docs - will do that and then merge

PR #249, opened 2016-08-19
Adds safeguard around files handling

Succo:
This should reduce the risk of incorrect file creation.
And prevent the plugin from failing to load in case of incorrect json inside the cache.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2103,28 +2103,34 @@ def url_processor_cb(data, command, return_code, out, err):
     return w.WEECHAT_RC_OK
 
 def cache_write_cb(data, remaining):
-    cache_file = open("{}/{}".format(WEECHAT_HOME, CACHE_NAME), 'w')
-    cache_file.write(CACHE_VERSION + "\n")
-    for channel in channels:
-        if channel.active:
-            for message in channel.messages:
-                cache_file.write("{}\n".format(json.dumps(message.message_json)))
+    with open("{}/{}".format(WEECHAT_HOME, CACHE_NAME), 'w') as cache_file:
+        cache_file.write(CACHE_VERSION + "\n")
+        for channel in channels:
+            if channel.active:
+                for message in channel.messages:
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -2103,28 +2103,34 @@ def url_processor_cb(data, command, return_code, out, err):
...
-        if channel.active:
-            for message in channel.messages:
-                cache_file.write("{}\n".format(json.dumps(message.message_json)))
+    with open("{}/{}".format(WEECHAT_HOME, CACHE_NAME), 'w') as cache_file:
+        cache_file.write(CACHE_VERSION + "\n")
+        for channel in channels:
+            if channel.active:
+                for message in channel.messages:
+                    cache_file.write("{}\n".format(json.dumps(message.message_json)))
+        cache_file.close()
```
bendem: Doesn't `with` automatically closes that file when it goes out of its scope?

wee_slack.py
```diff
@@ -2103,28 +2103,34 @@ def url_processor_cb(data, command, return_code, out, err):
...
+                    message_cache[j["channel"]].append(line)
+                dbg("Completed loading messages from cache.", main_buffer=True)
+        except IOError:
+            w.prnt("", "cache file not found")
+            pass
+        except ValueError:
+            w.prnt("", "cache file invalid")
+            pass
+        finally:
+            cache_file.close()
```
bendem: Same as above.

Succo:
@bendem oh yeah, I got mixed up.

i removed them and the IO error catching.

frioux:
I think that just as important is making the program not crash if an invalid file exists.  Even this code could create partial files in the event of a system crash.  (write/close isn't atomic unless it's a single write system call.)

Succo:
I had added both.

frioux:
Ah indeed, I missed the second set of changes when reading over the diff.  Sorry about that!

Succo:
No problem, with the conflict, now it might be easier to add better file handling in a new PR.

PR #252, opened 2016-09-10
Update readme with new variable name

bronzehedwick:
Also update the button label in slack's API settings to match their new change.


tfheen:
This needs changing the extension name internally too before/at the same time as we merge.

rawdigits:
The script is _finally_ called "slack" and not wee_slack. The settings migration was added to wee-ng to move folks from the old variables to the new ones.

Issue #163, opened 2016-01-29
Print token that failed when a token fails


rawdigits:
only print the first bit, otherwise we log tokens in the console log.

PR #255, opened 2016-09-25
Print out first ten characters of erroring tokens

tfheen:
If you have multiple tokens, it can be hard to figure out which one is
causing an error, so print out the first ten characters to help
identify the problematic one.

Fixes: #163

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -253,7 +253,15 @@ class SlackServer(object):
                             self.message_buffer.pop(message_id)
             return True
         else:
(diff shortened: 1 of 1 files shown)


tfheen:
Did this as a PR so @rawdigits can comment on whether 10 is reasonable or not. I think it should be safe, but review appreciated.

rawdigits:
Looks good - the first section is actually not random, but related to the Team ID, so this is absolutely fine w/ me.

PR #257, opened 2016-09-26
fixes reconnects, which hang if we never hear back from an rtm.start

rawdigits:


wee_slack.py
```diff
@@ -206,10 +207,24 @@ def ping(self):
         request = {"type": "ping"}
         self.send_to_websocket(request)
 
+    def should_connect(self):
+        """
+        If we haven't tried to connect OR we tried and never heard back and it
+        has been 65 seconds consider the attempt dead and try again
+        """
+        if self.connection_attempt_time == 0 or self.connection_attempt_time + 65 > int(time.time()):
+            return true
```
tfheen: True, not true.

wee_slack.py
```diff
@@ -206,10 +207,24 @@ def ping(self):
...
+        else:
+            return false
```
tfheen: Ditto, False, not false.

wee_slack.py
```diff
@@ -206,10 +207,24 @@ def ping(self):
...
+
     def connect_to_slack(self):
         t = time.time()
+        #Double check that we haven't exceeded a long wait to connect and try again.
+        if self.connecting == True and should_connect():
```
tfheen: You can just do `if self.connecting`, no need to check for `== True`

PR #261, opened 2016-09-29
consistently open mpdm the same way by handling the process_

rawdigits:
We were parsing mpdm channel names during initial startup, but not when a new one was opened and we received a process_group_joined from slack. We now classify them as mpdm in both places for consistency. Also changed the seperator from "," to "|", because the comma caused problems with the distracting channels setting.


rawdigits:
Testing with this seems fine (actually much better than previous). Merging.

PR #262, opened 2016-09-29
add whitespace and anchors

### LINKED ISSUE #253: Reactions print literal message
I'm running a weechat instance on a FreeBSD server inside tmux.

When I type a reaction, e.g. `3+:smile:`, the output in Slack is +3:smile: as a new message, instead of reacting to the 3rd most recent message.

rawdigits:
(maybe) fixes #253

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -369,7 +369,7 @@ def buffer_input_cb(b, buffer, data):
     channel = channels.find(buffer)
     if not channel:
         return w.WEECHAT_RC_OK_EAT
-    reaction = re.match("(\d*)(\+|-):(.*):", data)
+    reaction = re.match("^\s*(\d*)(\+|-):(.*):\s*$", data)
     if not reaction and not data.startswith('s/'):
         channel.send_message(data)
         # channel.buffer_prnt(channel.server.nick, data)


wee_slack.py
```diff
@@ -369,7 +369,7 @@ def buffer_input_cb(b, buffer, data):
     channel = channels.find(buffer)
     if not channel:
         return w.WEECHAT_RC_OK_EAT
-    reaction = re.match("(\d*)(\+|-):(.*):", data)
+    reaction = re.match("^\s*(\d*)(\+|-):(.*):\s*$", data)
```
tfheen: I don't think we should eat leading whitespace, since it's a way for people to escape when explaining something.

rawdigits: because of the anchors, they would have to type something on a line by itself to match. if they put any non whitespace character before _or_ after, the regex won't match.

i.e.

"+:blah:" = Match
"    +:blah:"  = Match
"2+:blah:   " = Match

but

"2+:blah: isn't working for me!" = No match
"I'm trying to do +:blah:" = No match
" +:blah:??" = No match

It seems like someone typing a reaction with just leading space as a way to show the text of the reaction would be rare.

rawdigits:
I think this will do what most people expect. @bronzehedwick lmk if it fixes your issue.

bronzehedwick:
OK, thanks! I'll try building from source with the latest changes and give it a shot.

PR #266, opened 2016-10-09
use the correct colors for nick_prefix/nick_suffix

### LINKED ISSUE #265: Wee-slack can't handle colors for nick_prefix and nick_suffix
rno

What it should do is color the prefix/suffix (`<` and `>`) with colors specified in `weechat.color.chat_nick_prefix` and `weechat.color.chat_nick_suffix`

After my fix it looks like this:

irth:
Fixes #265

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -674,7 +674,12 @@ class Channel(object):
                     name = prefix_same_nick
             else:
                 nick_prefix = w.config_string(w.config_get('weechat.look.nick_prefix'))
+                nick_prefix_color_name = w.config_string(w.config_get('weechat.color.chat_nick_prefix'))
+                nick_prefix_color = w.color(nick_prefix_color_name)
+
                 nick_suffix = w.config_string(w.config_get('weechat.look.nick_suffix'))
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -683,7 +688,7 @@ def buffer_prnt(self, user='unknown_user', message='no message', time=0):
                 else:
                     name = user
                     self.last_active_user = None
-                name = nick_prefix + name + nick_suffix
+                name = nick_prefix_color + nick_prefix + name + nick_suffix_color + nick_suffix + w.color("reset")
```
tfheen: Doesn't this need a `w.color("reset")` before name in (arguably very weird) case you've disabled nick colouring, but use `weechat.color.chat_nick_prefix`?

irth: I guess it does. I'll look into that in a few minutes.

irth: 09:30 <Mithrandir_> if you could just fix up using --amend and force-push that's better for me.  
    09:30 <Mithrandir_> since it means I can just mash the merge button
In case someone was wondering why the commit mentioned here dissapeared ;)

PR #279, opened 2016-11-07
Make it possible to set timeouts on slack URL comms

GregBowyer:
On very busy slack networks the default timeout of `20000` can be a
problem. This timeout can mean that the extension is unable to download
all the necesarry data before weechat times out the connection.

This patch allows for the timeout to be adjusted so that for larger
networks the end user can specify it.

---
This change is [Reviewable](https://reviewable.io/reviews/rawdigits/wee-slack/279)

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2202,22 +2202,24 @@ def complete_next_cb(data, buffer, command):
 
 # Slack specific requests
 def async_slack_api_request(domain, token, request, post_data, priority=False):
+    global slack_timeout
     if not STOP_TALKING_TO_SLACK:
         post_data["token"] = token
         url = 'url:https://{}/api/{}?{}'.format(domain, request, urllib.urlencode(post_data))
         context = pickle.dumps({"request": request, "token": token, "post_data": post_data})
         params = {'useragent': 'wee_slack {}'.format(SCRIPT_VERSION)}
         dbg("URL: {} context: {} params: {}".format(url, context, params))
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -2202,22 +2202,24 @@ def complete_next_cb(data, buffer, command):
 
 # Slack specific requests
 def async_slack_api_request(domain, token, request, post_data, priority=False):
+    global slack_timeout
```
tfheen: You don't need global to read the variable, so this line can be dropped.

wee_slack.py
```diff
@@ -2202,22 +2202,24 @@ def complete_next_cb(data, buffer, command):
...
         url = 'url:https://{}/api/{}?{}'.format(domain, request, urllib.urlencode(post_data))
         context = pickle.dumps({"request": request, "token": token, "post_data": post_data})
         params = {'useragent': 'wee_slack {}'.format(SCRIPT_VERSION)}
         dbg("URL: {} context: {} params: {}".format(url, context, params))
-        w.hook_process_hashtable(url, params, 20000, "url_processor_cb", context)
+        w.hook_process_hashtable(url, params, slack_timeout, "url_processor_cb", context)
 
 
 def async_slack_api_upload_request(token, request, post_data, priority=False):
+    global slack_timeout
```
tfheen: Ditto, not needed.

rawdigits:
👍 looks good once the globals are out.

PR #281, opened 2016-11-08
fixes #274 multiline message rendering bug

### LINKED ISSUE #274: Odd output on multiline code messages
Been getting some odd output on multiline code-block messages

Received:
image

From:
image

scharfie:
This PR simply splits incoming messages by newline, and renders each to the buffer.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -697,7 +697,7 @@ class Channel(object):
             curr_color = w.color(chat_color)
             if colorize_nicks and colorize_messages and self.server.users.find(user):
                 curr_color = self.server.users.find(user).color
-            message = curr_color + message
+
             for user in self.server.users:
                 if user.name in message:
                     message = user.name_regex.sub(
@@ -707,6 +707,7 @@ class Channel(object):
             message = HTMLParser.HTMLParser().unescape(message)
 
             for line in message.split("\n"):
+                line = curr_color + line
                 data = u"{}\t{}".format(name, line).encode('utf-8')
                 w.prnt_date_tags(self.channel_buffer, time_int, tags, data)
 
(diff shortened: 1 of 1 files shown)


trygveaa:
Multiline messages works fine for me, I would not want to have them as separate messages.

scharfie:
@trygveaa the PR is to fix #274 - perhaps that bug is specific to OS and/or version of python?

trygveaa:
Yes, I saw that issue. It must be a problem only in some specific cases/versions, since I don't experience it (and I would guess the majority doesn't either). Could you give some info about your system in the issue? I can't see that either of you mention OS, python version or weechat version.

grubernaut:
@trygveaa this happens to me on both:
```
Arch Linux
Python: 2.7.12
Weechat: 1.6
```
And:
```
OSX 10.12.1
Python: 2.7.10
Weechat: 1.5
```

mrrsm:
I have this same problem on 
```
Ubuntu 14.04.5
Python 2.7.6
Weechat 1.6
```

scharfie:
I'm running:

    OS X:    10.12
    Python:  2.7.12
    Weechat: 1.5

drselump14:
Same issue here:
```
OS X: 10.12
Python: 2.7.13
Weechat: 1.7
``` 
This patch fixed the problem in my environment

rawdigits:
I am unable to reproduce this on my setup, which is

```
Ubuntu: 16.04
Python: 2.7.13
Weechat: 1.7
```

I'm with @trygveaa in that I would not really like to have multiline messages appearing as separate. If this makes things work better for some folks, I'd be willing to add it as a default-off configurable option, but if possible can some of you share screenshots so I can understand the issue a bit better?

mrrsm:
I see the first line of text and the following says python and the full channel path
Weechat side: http://imgur.com/a/i17iS
Slack Web side: http://imgur.com/a/Alm4C

rawdigits:
The message printing code has been changed and no longer exhibits this behavior in wee-ng.

PR #282, opened 2016-11-21
Update hashtable on buffer open

rawdigits:
This _should_ make it impossible for the wrong thing to happen with buffers. Every time a new buffer is opened, the hashtable will be updated, so messages to the wrong person via a bad cache should no longer happen.

**tested by closing lots of DM buffers and then sending messages. I was able to trigger the behavior by closing 100 DMs and then opening one, which would result in my first message going to the wrong person.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -886,7 +886,7 @@ class User(object):
         try:
             if compare_str == self.name or compare_str == self.identifier:
                 return True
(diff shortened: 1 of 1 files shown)


tfheen:
Did you mean to just do it on open and not close? It looks ok, but if so, the first commit should probably be removed/the commits should be squished.

rawdigits:
good point  - squashed and pushed it back up!

PR #284, opened 2016-11-23
remove # from channel.name so /slack openweb works

sbgood:
`/slack openweb` won't open in the right channel due to a `#` being prepended to the channel name in the url. This uses regex to strip the `#` from`channel.name` before passing it to the url.

I've never done a pull request before, so please tell me if I can improve this somehow.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+.DS_Store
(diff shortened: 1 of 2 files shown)


rawdigits:
Thanks! I'm surprised someone else is using openweb, I thought it was just me. :)

One request is to remove the .DS_Store gitignore, as it doesn't really apply broadly to many people using this repo. I'll close this PR and ask that you submit a fresh one with only your original change.

Issue #161, opened 2016-01-27
make this work with screen_away.py


rawdigits:
sets you away when you detach a screen session (and hopefully tmux)

need to make sure /away global works in slack

furai:
Thanks for looking into this.

torhve:
I recently implemented this in matrix.lua. Maybe this patch could serve as inspiration.

https://github.com/torhve/weechat-matrix-protocol-script/commit/87596c43377719c60d7041827e71cfa6f726af32

WeeChat also recently switch its bar item away to be in core instead of irc plugin, so away status is actually displayed in the input bar on non-irc-buffers too.

PR #285, opened 2016-12-19
Add support for /away

tfheen:
Set localvar which is picked up by weechat and which should be picked
up by `screen_away.py` and similar tools.

Thanks to Tor Hveem (@torhve) for inspiration.

Fixes: #161
This should also make /away D


rawdigits:
LGTM - thanks!

PR #286, opened 2016-12-22
Make nick colourization a bit more controllable

ToxicFrog:
Turning on or off colorize_nicks will now affect only whether nicks
occuring *in a message* are colourized, and not whether the nick of the
person speaking is colourized. The latter is now controlled by the
weechat.color.chat_nick_colors property, same as IRC.


rawdigits:
This seems reasonable. I haven't tested it yet. When you have a moment, please resolve the merge conflicts and I'll poke at it.

PR #287, opened 2016-12-22
Fix /me handling for sent and received messages

ToxicFrog:
- Received messages are now properly detected whether they use _
wrapping or the me_message subtype
- Sent messages are now properly rendered as actions in the buffer
- Message replies in general are now handled by process_message(), and
thus should have formatting consistent with normal messages

### DIFF
--- a/_pytest/conftest.py
+++ b/_pytest/conftest.py
@@ -69,6 +69,8 @@ class FakeWeechat():
         return "testuser"
     def buffer_new(*args):
         return "0x8a8a8a8b"
+    def prefix(self, type):
+        return ""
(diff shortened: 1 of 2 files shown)


ToxicFrog:
This one is not 100% baked, on further testing; incoming actions don't respect the `weechat.color.chat_nick_colors` setting.

ToxicFrog:
Updated for the wee-ng version of the patch. Incoming and outgoing /me should work (both me_message and legacy), and /me edits should mostly work.

rawdigits:
👍 lgtm - subtype logic move is logical.

PR #288, opened 2016-12-22
Disable cert validation (with warning) when no cafile available

ToxicFrog:
ssl.get_default_verify_paths() can return cafile (a single unified file
containing the root certificates), capath (a directory containing one
file for each root certificate), or both. The websocket-client library
does not support capath.

In cases where cafile is not available, we now log a warning and disable
cert validation rather than giving up on connecting entirely.


tfheen:
I'd rather have this behind a setting so we don't accidentially fall back to insecure mode because of a system misconfiguration.

ToxicFrog:
I'm working on a refactor of the settings to make it easier to add new ones; I'm a-ok with deferring this until that is done and then adding a new setting to control this.

rawdigits:
I work at Slack and am on the security team, so this seems like something I don't want to do at all. We should figure out if there is a way to make the websocket act properly in these cases. It is pretty dangerous to allow people to connect without doing validation here.

PR #289, opened 2016-12-22
Include - in the allowable nick characters

ToxicFrog:
I was having a problem where, on a net with users "foo" and "foo-bar",
attempting to @foo-bar would send @foo- instead. Turns out it thought -
terminated the nick.

also, it turns out nicks can end with - and probably . as well.


rawdigits:
Fixed in latest

PR #290, opened 2016-12-23
Refactor the way settings work

ToxicFrog:
All the per-setting globals are replaced with a single global object,
'config'; its attrs are the weechat settings. They're automatically
updated on config change, and the setting names and defaults are
specified as part of a map in the PluginConfig class definition.

To add a new setting, add the setting name and default to
PluginConfig.settings, and if it's not a boolean, add an appropriate
PluginConfig.get_<name> method as well.


rawdigits:
I like what you did here. Running it myself as a test now.

rawdigits:
superseded by #300

PR #296, opened 2017-01-09
Handles upload with spaces in filename

lchausmann:
in order to upload files with spaces in filename, escape the filename if spaces is found.


rawdigits:
Seems like a pretty reasonable change. Thanks!

PR #300, opened 2017-01-26
Refactor the way settings work + fixes

rawdigits:
via @ToxicFrog:

All the per-setting globals are replaced with a single global object,
'config'; its attrs are the weechat settings. They're automatically
updated on config change, and the setting names and defaults are
specified as part of a map in the PluginConfig class definition.

To add a new setting, add the setting name and default to
PluginConfig.settings, and if it's not a boolean, add an appropriate
PluginConfig.get_ method as well.


rawdigits:
@ToxicFrog I fixed the config vars in a few places and just moved this to a new PR.

PR #302, opened 2017-01-27
Full precision Slack timestamp handling

rawdigits:
Slack sends a timestamp in messages as "ts", which looks like "1473363728.000174".
The bit before the "." is an epoch timestamp and the bit after is a unique ID (not fractional seconds as I used to think).

If two messages are sent during the same second (common), the only way for me to add an emoji reaction is to work my way back up the list to find the right message, but weechat's resolution is only to the second.

Today I noticed that the hdata structure for line_data stores both the 'message time' and 'printed time' - the latter seems useless so I am going to store the .0000174 bit there. I'm worried something might use it, but so far i haven't seen evidence of that.

This allows me to search the backlog of a channel for the full time(string?) as opposed to the sloppy method. We need this resolution to make threads work well, mostly for performance.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -680,7 +680,6 @@ class Channel(object):
         writes output (message) to a buffer (channel)
         """
         set_read_marker = False
-        w.prnt("", time)
         time_int, time_id = time.split(".", 2)
         time_int = int(time_int)
         tags = "nick_" + user
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -675,16 +675,18 @@ class Channel(object):
                 self.current_short_name = new_name
                 w.buffer_set(self.channel_buffer, "short_name", new_name)
 
-    def buffer_prnt(self, user='unknown_user', message='no message', time=0):
+    def buffer_prnt(self, user='unknown_user', message='no message', time="0.0"):
         """
         writes output (message) to a buffer (channel)
         """
         set_read_marker = False
-        time_float = float(time)
+        w.prnt("", time)
+        time_int, time_id = time.split(".", 2)
+        time_int = int(time_int)
         tags = "nick_" + user
         user_obj = self.server.users.find(user)
         # XXX: we should not set log1 for robots.
-        if time_float != 0 and self.last_read >= time_float:
+        if time_int != 0 and self.last_read >= time_int:
             tags += ",no_highlight,notify_none,logger_backlog_end"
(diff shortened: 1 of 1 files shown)


tfheen:
]] Ryan Huber 

> Slack sends a timestamp in messages as "ts", which looks like "1473363728.000174".
> The bit before the "." is an epoch timestamp and the bit after is a unique ID (not fractional seconds as I used to
> think).

What is it unique by?  Channel?

> Today I noticed that the hdata structure for line_data stores both the
> 'message time' and 'printed time' - the latter seems useless so I am
> going to store the .0000174 bit there. I'm worried something might use
> it, but so far i haven't seen evidence of that.
> 
> This allows me to search the backlog of a channel for the full
> time(string?) as opposed to the sloppy method. We need this resolution
> to make threads work well, mostly for performance.

I haven't looked at the code yet, but it feels like we should be able to
just tag the message and ask weechat to search by tag instead. Maybe
suggest that to FlashCode?

-- 
Tollef Fog Heen

rawdigits:
It is indeed unique per channel.

If we tag the message with this, our tags list would grow by +1 for every new id (of which it can be 0 - 999999 I think). This would probably waste a lot of memory elsewhere.

I may be misunderstanding how tags work, but I think this is true.

Also I looked back at git history in weechat and FlashCode rejected a request for storing milliseconds as he considered it not useful for irc and wasteful for storage. (we don't need milliseconds, but we could have overloaded it).

tfheen:
]] Ryan Huber 

> If we tag the message with this, our tags list would grow by +1 for
> every new id (of which it can be 0 - 999999 I think). This would
> probably waste a lot of memory elsewhere.

I didn't necessarily mean using the current set of message tags, which I
don't think are optimised for efficient storage (though they pretty
easily could be), more just like having a uint64 or similar we can
attach to a message to find it again.

> Also I looked back at git history in weechat and FlashCode rejected a
> request for storing milliseconds as he considered it not useful for
> irc and wasteful for storage. (we don't need milliseconds, but we
> could have overloaded it).

Yeah, though I'd still suggest talking to him about it.  Maybe he's got
other ideas on how to solve this in a good manner.

-- 
Tollef Fog Heen

rawdigits:
This idea was implemented in -ng. Weechat has an open issue to add precision timestamps which we will start using when available.

PR #311, opened 2017-02-09
Display purpose in lieu of topic

thentenaar:
Seems that people tend to set the channel purpose like a topic, but never actually set a topic, so why not display it in lieu of a topic? :P

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -436,6 +436,9 @@ class Channel(object):
         self.topic = kwargs.get('topic', {"value": ""})["value"]
         self.unread_count = kwargs.get('unread_count_display', 0)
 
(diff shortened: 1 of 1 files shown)


adamgibbins:
FYI you might also want to contribute this to https://github.com/wee-slack/wee-ng which is a ground up rewrite and I think going to replace this codebase.

thentenaar:
I wouldn't mind contributing it there also, but AFAICT it doesn't handle the channel topic at all just yet. :P

rawdigits:
just added it in wee-ng here. thanks @thentenaar !

https://github.com/wee-slack/wee-ng/commit/88cafdde3337c29c94e54da143bf81db0e489448

PR #313, opened 2017-02-16
use highlight words configured in slack

jreinert:


rawdigits:
Wee-ng now implements highlight words from slack! Thanks!

PR #314, opened 2017-02-21
Force nick completion with at-sign

adsr:
Not sure if this is desirable for everyone. Currently you have to hit tab twice to get an at-sign-prefixed nick, e.g., `"br" <tab> "brad" <tab> "@brad"`. Ideally we'd (1) go straight to the at-sign, and (2) allow the end-user to tab through potential matches. This patch implements (1).

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2259,11 +2259,12 @@ def complete_next_cb(data, buffer, command):
     word = input[word_start:word_end]
     for m in channel.members:
         user = channel.server.users.find(m)
-        if user.name == word:
+        if word == user.name[:len(word)]:
             # Here, we cheat.  Insert a @ in front and rely in the @
             # nicks being in the completion list
-            w.buffer_set(buffer, "input", input[:word_start] + "@" + input[word_start:])
-            w.buffer_set(buffer, "input_pos", str(w.buffer_get_integer(buffer, "input_pos") + 1))
+            extra = len(user.name) - len(word)
+            w.buffer_set(buffer, "input", input[:word_start] + "@" + user.name + input[word_end:])
+            w.buffer_set(buffer, "input_pos", str(word_end + extra + 1))
             return w.WEECHAT_RC_OK_EAT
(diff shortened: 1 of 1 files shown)


trygveaa:
> Not sure if this is desirable for everyone.

I don't want this at least. E.g. `/query <nick>` doesn't work with the prefix, and when I talk _about_ a person (in contrast to when I'm talking _to_ a person), I don't necessarily want the prefix.

bendem:
This is neat, but it should be configurable and when enabled, depend on the context (command, PM).

mattrobenolt:
I don't want this either. I hate using @'s to mention people. Feels clunky.

tfheen:
I'd rather this didn't change, I implemented it the way it is intentionally so people can make a choice about whether they want to highlight folks or not.

trygveaa:
@tfheen: Though there is an issue that if you type some characters and press tab multiple times, it won't cycle between the nicks.

adsr:
@trygveaa Good point re: mentioning vs highlighting. I like @bendem 's idea to make it configurable. Maybe the ideal tab list is `@brad <tab> brad <tab> @bree <tab> bree` etc, or optionally `brad <tab> @brad <tab> bree <tab> @bree` (non-at-sign first). Or make it a 4-way option and support `brad <tab> bree` and `@brad <tab> @bree`. In any case I suspect most of the work will be hacking the completion code.

@tfheen I haven't confirmed this but I suspect `buffer_set` might clear the buffer's tab completion state, which prevents tabbing through the completion list from working.

tfheen:
Yup, I agree the bug in 2) should be fixed, just that the current behaviour on having to double tab is intentional. Happy to take patches for that, but I suspect you might want to write your PR against the wee-ng repo (sibling of this one); @rawdigits is in the middle of rearranging everything.

trygveaa:
Closing since this has been dormant for a long time, and it doesn't seem like the majority of the users want it.

PR #315, opened 2017-02-21
Entity decode topics

adsr:


tfheen:
This got stuck behind a lot of rework, so I took the liberty to redo it to fit into the new framework.  Thanks for the patch, and apologies about the lack of feedback.

PR #318, opened 2017-03-01
use nick_color_name instead of irc_nick_color_name

bendem:
that removes the dependency on the irc plugin for nick coloring.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1536,7 +1536,9 @@ class SlackDMChannel(SlackChannel):
 
     def update_color(self):
         if config.colorize_private_chats:
-            self.color_name = w.info_get('irc_nick_color_name', self.name)
+            self.color_name = w.info_get('nick_color_name', self.name)
+            if not self.color_name:
+                self.color_name = w.info_get('irc_nick_color_name', self.name)
             self.color = w.color(self.color_name)
         else:
             self.color = ""


devkev:
If `w.info_get('nick_color_name', self.name)` returns falsy, you should fall back to using `w.info_get('irc_nick_color_name', self.name)`, so that the plugin doesn't break for people on older versions of weechat (pre 1.5 when nick coloring was moved to core).

bendem:
@devkev done

bendem:
Rebased

ToxicFrog:
Now that I've actually had a chance to look at this (sorry for the delay) -- it looks fine as far as it goes, but it only affects `SlackDMChannel` -- there's a similar function in `SlackUser` that needs to be updated as well.

bendem:
Yeah, the change was lost in the rebase due to a conflict with https://github.com/wee-slack/wee-slack/commit/925278f11a5eb137c42b5232714c239fe7e25c3b. Should I add the fallback to `SlackUser` or remove it from `SlackDMChannel`?

ToxicFrog:
Add it to SlackUser, IMO, or factor it out into a common function used by both -- the goal here is to use the new API if available, but cleanly fall back to the old one for people on pre-1.5 weechat.

ToxicFrog:
Obsoleted by #391

PR #320, opened 2017-03-05
Update token link

oschwald:
The previous page only links to the new style application tokens, which do not work.


chreekat:
Thanks for this, I was wondering if I was crazy. :)

rawdigits:
Ah yep - it moved! thanks!

PR #322, opened 2017-03-13
SearchList improvements

KellerFuchs:
- [x] Reduces code duplication
- [x] Removes expensive list walks, responsible for weechat hanging in tab-completion on large Slack channels


rawdigits:
Thanks much for the PR @KellerFuchs - during the refactor, searchlists were removed entirely!

KellerFuchs:
Yes, I've been pointed to wee-ng by now.
I will try it out, send patches as needed.

Issue #325, opened 2017-03-21
Feature Request: Bold and italics terminal font


mcastorina:
It would be nice to display text surrounded by * or _ as bold or italic in the buffer like in Slack.
e.g. `*this text is bold* and _this text is italic_`

ToxicFrog:
Implemented in #338

auscompgeek:
I'm not seeing italics.

ToxicFrog:
IRC doesn't support italics (and neither do most terminal emulators); italics in slack get rendered as underlines in IRC, and vice versa.

chreekat:
I'm not upset with the way things are, but my terminal (Gnome Terminal) *does* support italics, and I think it would be nice to keep the option open for the future.

ToxicFrog:
It looks like weechat itself supports italics if the terminal and libcurses do, so I don't see any technical obstacles to this.

It shouldn't be hard to add a configuration setting to toggle between rendering italics as underline vs. as italics.

chreekat:
Yeah, I was just trying it out -- weechat's `/help trigger` even lists the necessary trigger for bold/underline/italic as an example[1]. :) Unfortunately it wasn't immediately successful: it did the "use reverse in place of italics" hack that I've seen elsewhere and don't immediately recall how to fix...

[1]: Copied here: 
```
/trigger add effects modifier weechat_print "${tg_tag_nick}" "==\*(\S+)\*==*${color:bold}${re:1}${color:-bold}*== ==_(\S+)_==_${color:underline}${re:1}${color:-underline}_== ==/(\S+)/==/${color:italic}${re:1}${color:-italic}/"
```

mcastorina:
Bold and italic doesn't appear to be working: `*_bold and italic_*`

ToxicFrog:
Testing: `*_bold and italic_*` renders as `_bold and italic_` in bold, and `_*italic and bold*_` renders as `*italic and bold*` in italic (well, underlined).

Looking at the code I would expect the former to work even if the latter doesn't.

ToxicFrog:
If they aren't nested side by side it works, e.g. `_italic and *bold* in the middle_` renders correctly.

auscompgeek:
> IRC doesn't support italics

IRC does have italics. weechat has the default keybind of <kbd>C-c i</kbd> to make text italics.

ToxicFrog:
[...]

PR #329, opened 2017-04-01
Bold and italics terminal font (Issue #325)

mcastorina:
Added bold, italics, and underline to buffer_prnt that matches slack's delimeters (*, _, ~ respectively).
Known issues: Replacements such as `*_bold italics_*` and `_*bold italics*_` don't work, however `*bold _bold italics_*` works.

Some examples:

2017-04-01-103932_1621x166_scrot

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -729,10 +729,10 @@ class Channel(object):
             chat_color = w.config_string(w.config_get('weechat.color.chat'))
             if type(message) is not unicode:
                 message = message.decode('UTF-8', 'replace')
-            curr_color = w.color(chat_color)
+            curr_color = w.color("resetcolor") + w.color("|{}".format(chat_color))
             if config.colorize_nicks and config.colorize_messages and user_obj:
                 curr_color = user_obj.color
-            message = curr_color + message
+            message = curr_color + convert_attributes(message)
             for user in self.server.users:
                 if user.name in message:
                     message = user.name_regex.sub(
@@ -958,7 +958,8 @@ class User(object):
                 self.color_name = w.config_string(w.config_get('weechat.color.chat_nick_self'))
             else:
                 self.color_name = w.info_get('irc_nick_color_name', self.name)
-            self.color = w.color(self.color_name)
+            # "|" keeps previous attributes (bold, underline, italic)
+            self.color = w.color("|{}".format(self.color_name))
         else:
             self.color = ""
             self.color_name = ""
@@ -998,7 +999,8 @@ class Bot(object):
     def update_color(self):
         if config.colorize_nicks:
             self.color_name = w.info_get('irc_nick_color_name', self.name.encode('utf-8'))
-            self.color = w.color(self.color_name)
+            # "|" keeps previous attributes (bold, underline, italic)
+            self.color = w.color("|{}".format(self.color_name))
         else:
             self.color_name = ""
             self.color = ""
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -2454,6 +2456,46 @@ def scrolled_cb(signal, sig_type, data):
...
+    }
+    # Find any of the characters within ` and make sure they don't match
+    codes = list(re.finditer("(`.+?`)", text))
+    effect = ""
+    def filter_match(match):
+        # Check if within a code range
+        for code in codes:
+            if match.start() >= code.start() and match.start() <= code.end():
+                # Don't substitute if within code range
+                return ("{}"*6).format(
```
kilogram: Instead of "{}"*6, perhaps "".join would be better?

wee_slack.py
```diff
@@ -2454,6 +2456,46 @@ def scrolled_cb(signal, sig_type, data):
...
+        # Perform the substitute
+        return ("{}"*6).format(
+            match.group(1),
+            w.color(effect),
+            match.group(3),
+            match.group(4),
+            w.color("-{}".format(effect)),
+            match.group(6)
+        )
+    regex_match = "(^|\s)({0})([^{0}])(.*?)(\\2)($|\W)"
```
kilogram:
```
(^|\s)({0})([^{0}])(.*?)(\\2)($|\W)
       ^ How  does this work with "*"? Shouldn't it be "\*"?
```

kilogram:
```
(^|\s)({0})([^{0}])(.*?)(\\2)($|\W)
            ^ Use forward lookahead here (?=[^{0}])
```

kilogram:
```
(^|\s)({0})([^{0}])(.*?)(\\2)($|\W)
                          ^ Probably best not to use a backreference since you can just replace the literal
```

wee_slack.py
```diff
@@ -2454,6 +2456,46 @@ def scrolled_cb(signal, sig_type, data):
...
+    for key in matches:
```
kilogram: Why not have a list of matches here directly? You can also escape them here.
```
for key in [
        "\\*",  # bold
        "_",    # underscore
        "~",    # strikethrough
        ]:
    ...
```

wee_slack.py
```diff
@@ -2454,6 +2456,46 @@ def scrolled_cb(signal, sig_type, data):
...
+        "*": "bold",
+        "_": "italic",
+        "~": "underline"    # TODO: change to unicode strikethrough
+    }
+    # Find any of the characters within ` and make sure they don't match
+    codes = list(re.finditer("(`.+?`)", text))
+    effect = ""
+    def filter_match(match):
+        # Check if within a code range
+        for code in codes:
```
kilogram: Alternatively, you can use ```(?=([^`]*`[^\`]*\`)*[^\`]*$)``` at the end of the regex. This checks to make sure that there are an even number of backticks in the remainder of the string, which is sufficient since backticks only nest to one layer. I'm not sure which is more performant, but also, that regex will become increasingly hard to read. But, if you add the re.VERBOSE flag and comment judiciously, it should be fine.

wee_slack.py
kilogram: Why do you anchor instead of ".*" ? Better yet, why anchor at all?

wee_slack.py
```diff
@@ -2454,6 +2456,46 @@ def scrolled_cb(signal, sig_type, data):
...
+            match.group(3),
+            match.group(4),
+            w.color("-{}".format(effect)),
+            match.group(6)
+        )
+    regex_match = "(^|\s)({0})([^{0}])(.*?)(\\2)($|\W)"
+    for key in matches:
+        effect = matches[key]
+        escaped_key = key.replace("", "\\")[:-1]
+        regex = re.compile(regex_match.format(escaped_key))
```
kilogram: If you compile with re.VERBOSE, you can add comments to the regex. Also, regex strings should start `r"`, but that's a nit.

wee_slack.py
kilogram: I think if you remove the anchors, and use `[\*_~]` and backreferences, you can accomplish the same goal.

kilogram: And you'd have to wrap the whole thing in a lookahead.

PR #331, opened 2017-04-09
Make tests pass on Python 3

tsroten:
This makes the tests for wee_slack pass on Python 3. This does not guarantee Python 3 support in wee_slack, but it provides a good starting point.

This does not break Python 2 support. All of these changes should be compatible with Python 2 and 3.

### DIFF
--- a/_pytest/conftest.py
+++ b/_pytest/conftest.py
@@ -18,13 +18,13 @@ class FakeWeechat():
     WEECHAT_RC_OK = True
 
     def __init__(self):
-        print "INITIALIZE FAKE WEECHAT"
+        print("INITIALIZE FAKE WEECHAT")
     def prnt(*args):
         output = "("
         for arg in args:
             if arg != None:
                 output += "{}, ".format(arg)
-        print "w.prnt {}".format(output)
+        print("w.prnt {}".format(output))
     def hdata_get(*args):
         return "0x000001"
     def hdata_pointer(*args):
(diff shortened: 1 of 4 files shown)


tfheen:
@tsroten, can you rebase this? It looks reasonable, but needs a rebase (it probably ended up conflicting after a big pep8 related cleanup).

smaeul:
I rebased these changes at https://github.com/smaeul/wee-slack/tree/python3. Several of them no longer applied, and I needed several more changes to get the plugin to load. Those I added in additional commits. I'm not opening a PR because I'm not actually able to test them (turns out my Slack workspace disables third-party app integrations), and because some of the more questionable changes probably break Python 2. But if anyone else wants to pick this up, it's a starting point.

trygveaa:
@smaeul: Nice! I tried a bit now, and it seems to work fine in weeslack (with python2), after i did one small change. I had to import `print_function` from `__future__` to make the print function with a file argument work.

If you want to test it yourself, there is one way btw. You can copy the session token from the web browser and use that. You can find it by searching for `xoxs` in the source of the Slack web page after logging in. (Note that it will expire when not used for a while).

Also, the tests don't run with python 3 for me. You have to change the print statements to print functions in the tests as well.

seitzbg:
Any plans to support Python 3 in wee-slack?

trygveaa:
Yes, it's on my todo list. We'll have to do it before July 1. when weechat will start being compiled with python 3 by default (https://weechat.org/scripts/python3/).

trygveaa:
Sorry for not following up on this sooner. I've now created a PR which should have complete support for Python 3, #686.

PR #334, opened 2017-04-12
Wee ng merge

rawdigits:
Lots of changes here:

Broadly, the new version uses something called the "eventrouter" to remove .find from normal use. This provides a massive speedup at the cost of a bit more memory. Nearly all objects now have inbuilt references to parents and associated objects.


trygveaa:
Wouldn't it be better to keep the commit history from wee-ng, i.e. merge in the whole history?

PR #336, opened 2017-04-13
Fix readme markdown

bronzehedwick:
Github changed their markdown parser, and broke a bunch of readmes. This includes not parsing heading hashes without a space between the hash and the content, which is what is happening here.


rawdigits:
ahhhh thanks!

PR #338, opened 2017-04-13
Translate slack formatting codes to/from IRC formatting codes

ToxicFrog:
This translates `*` into bold and `_` into underline (since most IRC clients don't support italics). It also properly handles outgoing `\x02` and `\x1F` characters.


rawdigits:
:thumbsup: to fancy text

PR #339, opened 2017-04-13
Defer nick colourization to weechat's built in support

ToxicFrog:
This removes the colorize_nicks setting, instead relying on weechat's built in settings for nick colours; it also moves from `irc_nick_color_name` to `nick_color_name`.


rawdigits:
👍 much cleaner

PR #342, opened 2017-04-13
do it for channels but not users, hence new elif

rawdigits:


tfheen:
Untested, but I think this is ok.

PR #344, opened 2017-04-13
PEP8 fixups

tfheen:
Lots of missing whitespace and other minor fixups.


rawdigits:
Thanks - I wonder when I broke my linter....

PR #345, opened 2017-04-13
Escape unicode when needed

immae:
This escapes unicode when needed since the weechat client doesn't handle unicode everywhere (in particular, in channel names).

It may be over-kill, in particular some escapes that are not actually needed. It would be nice to have at least the information somewhere of the true names of the channels (somewhere where it's not limited by weechat), but I don't know enough of the script/weechat for that.

Can someone review it?

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -3,6 +3,7 @@
 
 from functools import wraps
 
+import unicodedata
 import time
 import json
 import pickle
@@ -872,7 +873,7 @@ class SlackTeam(object):
             for v in self.channels.itervalues():
                 v.set_highlights()
     def formatted_name(self, **kwargs):
-        return self.domain
+        return to_ascii(self.domain)
     def buffer_prnt(self, data):
         w.prnt_date_tags(self.channel_buffer, SlackTS().major, tag("backlog"), data)
     def get_channel_map(self):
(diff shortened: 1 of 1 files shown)


trygveaa:
Weechat supports unicode in channel names, at least the version I'm using, which is 1.7.

I created a branch where I tried to do some proper handling of unicode, and at least channel names with unicode is working for me. Can you try it, and see if it's working for you? It's here: https://github.com/trygveaa/wee-slack/tree/fix/unicode

I don't think the branch handles unicode correctly everywhere though, and I would like to do it in a nicer way, so it's not ready for a PR yet.

If someone has a suggestion on how to nicely do `decode` on all the strings we receive from weechat and slack, and `encode` on all the strings we send back, that would be very helpful (see the last commit in that branch).

@immae: Btw, the `python: wrong arguments for function "buffer_set"` message you mentioned on irc happens when you send in a unicode string with non-ascii characters as one of the arguments to `buffer_set`.

@rawdigits: That branch also fixes editing non-ascii messages, the issue I mentioned on IRC.

immae:
@trygveaa : your changed work perfectly at first try, I'd definitely suggest merging them instead of my PR.

Concerning the `python: wrong arguments for function "buffer_set"`, they have become now:
```
09:50:43    weechat =!= | python: wrong arguments for function "info_get" (script: slack)
09:50:43    weechat =!= | python: wrong arguments for function "color" (script: slack)
```
(only two lines this time, not one per channel)
If you know how to fix it...

trygveaa:
> @trygveaa : your changed work perfectly at first try, I'd definitely suggest merging them instead of my PR.

Great!

> Concerning the python: wrong arguments for function "buffer_set", they have become now:

I hadn't added encoding to all the weechat calls yet, so that makes sense. I've opened a PR with a more complete fix now. Can you test #348 and see if you experience any issues with that?

tfheen:
As per IRC discussion, I'm closing this in favour of @trygveaa's more complete unicode revamp. Thanks a lot for the PR though!

PR #346, opened 2017-04-13
Add emojis to status as a DM topic

V13Axel:
This seemed like a pretty sane way to handle this, at least for the near-term. @rawdigits mentioned wanting to do the whole thing differently, but until such time... This adds in the emoji alongside topic status text in the format:

`[:status_emoji:] status_text`

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2314,8 +2314,9 @@ def process_user_change(message_json, eventrouter, **kwargs):
     profile = user.get("profile")
     team = kwargs["team"]
     team.users[user["id"]].update_status(profile.get("status_emoji"), profile.get("status_text"))
-    dmchannel = team.get_channel_map()[user["name"]]
-    team.channels[dmchannel].render_topic(topic=create_user_status_string(profile))
+    dmchannel = team.find_channel_by_members({user["id"]}, channel_type='im')
+    if dmchannel:
+        dmchannel.set_topic(create_user_status_string(profile))
 
 
 def process_user_typing(message_json, eventrouter, **kwargs):
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1060,6 +1060,12 @@ class SlackTeam(object):
     def buffer_prnt(self, data):
         w.prnt_date_tags(self.channel_buffer, SlackTS().major, tag("team"), data)
 
+    def find_channel_by_members(self, members, channel_type=None):
+        for channel in self.channels.itervalues():
+            if channel.get_members() == members and (
+                    channel_type is None or channel.type == channel_type):
+                return channel
+
     def get_channel_map(self):
         return {v.slack_name: k for k, v in self.channels.iteritems()}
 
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -1368,7 +1368,7 @@ def __init__(self, eventrouter, users, **kwargs):
         self.type = 'im'
         self.update_color()
         self.set_name(self.slack_name)
-        self.slack_topic = {"value": users[dmuser].profile.get("status_text")}
+        self.slack_topic = {"value": "[{}] {}".format(users[dmuser].profile.get("status_emoji"),users[dmuser].profile.get("status_text"))}
```
tfheen: Missing space after comma.

wee_slack.py
```diff
@@ -1932,9 +1933,9 @@ def process_user_change(message_json, eventrouter, **kwargs):
     user = message_json['user']
     profile = user.get("profile")
     team = kwargs["team"]
-    team.users[user["id"]].update_status(profile.get("status_text"))
+    team.users[user["id"]].update_status(profile.get("status_emoji"),profile.get("status_text"))
```
tfheen: Missing space after comma.

wee_slack.py
```diff
@@ -1932,9 +1933,9 @@ def process_user_change(message_json, eventrouter, **kwargs):
...
     dmchannel = team.get_channel_map()[user["name"]]
-    team.channels[dmchannel].render_topic(topic=team.users[user["id"]].profile["status_text"])
+    team.channels[dmchannel].render_topic(topic="[{}] {}".format(team.users[user["id"]].profile["status_emoji"],team.users[user["id"]].profile["status_text"]))
```
tfheen: Missing space after comma.

wee_slack.py
```diff
@@ -2400,6 +2401,11 @@ def resolve_ref(ref):
         # Something else, just return as-is
     return ref
 
+def create_user_status_string(profile):
+    status_emoji = profile.get("status_emoji") if profile.get("status_emoji") else "None"
+    status_text = profile.get("status_text") if profile.get("status_text") else "None"
+    return "[{}] {}".format(status_emoji, status_text)
```
tfheen: Won't this lead to `[None] None` as the topic if the other user has no emoji or text status? Wouldn't a blank topic be better?

V13Axel: Yep, that currently means that the status will show as `[None] None`. 

It was showing just `None` before my proposed changes, and I liked that more than having the topic empty when I tried it - An empty topic made me wonder whether the user had no status at all, or if setting the topic simply wasn't working.

So that said, my personal preference would be to at least indicate that there's no status _somehow_, even if `[None] None` isn't the best way. Any suggestions? Perhaps just stating it as "_username_ has no status", something like that?

rawdigits: I have a freshish suggestion. Make the topic:
`username [status_emoji] "status"`
if the status emoji or status are unset, just don't render them and leave it the username.

V13Axel: Great idea.

auscompgeek: What about `full_name [status_emoji] status`? Would solve the issue of not being able to see real names in wee-slack as well.

tfheen: We don't use full names elsewhere, so seems a bit weird to use it here, except we have lots of space? That said, I'm fine with either @rawdigits  or @auscompgeek's suggestion.

auscompgeek: As another suggestion, I believe Slack's mention completion presents users as `*@username*  full_name | status_emoji status_text`. We could show `full_name | status_emoji status_text`.

tfheen: @rawdigits, @V13Axel : I think @auscompgeek's suggestion sounds good. Thoughts/opinions? It'd be good if we can get this closed out.

PR #347, opened 2017-04-13
Add command for setting Slack Status

V13Axel:
Add in a command called `/slack status`, allowing users to set or clear their Slack Status. 
Usage:

`/slack status [:emoji:] [Status message]`

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2780,7 +2780,6 @@ def command_status(data, current_buffer, args):
 
         profile = {"status_text":text,"status_emoji":emoji}
 
-        team = EVENTROUTER.weechat_controller.buffers[current_buffer].team
         s = SlackRequest(team.token, "users.profile.set", {"profile": profile}, team_hash=team.team_hash, channel_identifier=channel.identifier)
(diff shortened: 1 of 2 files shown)


wee_slack.py
```diff
@@ -2760,6 +2760,32 @@ def command_away(data, current_buffer, args):
...
+            server.buffer_prnt("Usage: /slack status [status emoji] [status text].")
+            return
+
+        split_args = args.split(None, 2)
+        emoji = split_args[1] if len(split_args) > 1 else ""
+        text = split_args[2] if len(split_args) > 2 else ""
+
+        profile = {"status_text":text,"status_emoji":emoji}
+
+        team = EVENTROUTER.weechat_controller.buffers[current_buffer].team
```
rawdigits: This is redundant as line 2771 already retrieves the "team" reference.

rawdigits:
@V13Axel added one comment. looks good after that bit is changed.

V13Axel:
@rawdigits There we go! Removed the redundant team reference. Not sure how it wound up there, to be honest.

PR #348, opened 2017-04-15
Proper support for unicode

trygveaa:
This changes wee_slack to only use unicode strings internally, and encode/decode to utf-8 when it sends/receives data, which fixes some unicode errors that's been happening e.g. when editing non-ascii messages (hopefully fixes all unicode issues, but can't guarantee that). Some info from the first commit:

> This changes all strings to be unicode strings in an attempt to handle
> unicode better. The idea is to use unicode everywhere internally in
> wee_slack and decode/encode where data is received/sent to/from
> wee-slack.
> 
> That means we have to decode strings received from weechat or slack and
> encode strings we send to them. This way encoding and decoding is done
> on specific places, and if done everywhere, we should not get any
> unicode errors.

I've tested this briefly, but not much, so it probably needs some more testing before merging.

Relates to #345.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -287,7 +287,7 @@ class EventRouter(object):
         """
         try:
             # Read the data from the websocket associated with this team.
-            data = self.teams[team_hash].ws.recv()
+            data = decode_from_utf8(self.teams[team_hash].ws.recv())
             message_json = json.loads(data)
             metadata = WeeSlackMetadata({
                 "team": team_hash,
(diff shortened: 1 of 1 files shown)


immae:
Hi @trygveaa,
 It seems to work now thanks a lot!

As I said in irc, I'd suggest merging your PR even if there are missing things, since it"s a definite improvement, and we could adjust afterwards. Otherwise you'll keep having conflicts everywhere when rebasing, and it's not cool for you :/

immae:
I used it the whole day and no encoding errors appeared

PR #350, opened 2017-04-15
Add italic support, and settings to switch between italic and underline

ToxicFrog:
This adds three new settings:
- `render_italic_as` controls how italics are rendered in weechat; it defaults to `italic` but if your terminal doesn't support that `underline` is recommended
- `render_bold_as` controls how bold is rendered
- `map_underline_to` controls how underlined text is sent to slack; the default is `_` (send as italics) but you may want to change it to `*` (bold) or even `` (nothing).

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2231,9 +2231,13 @@ def render(message_json, team, channel, force=False):
         text = text.replace("&gt;", ">")
         text = text.replace("&amp;", "&")
         text = re.sub(r'(^| )\*([^*]+)\*([^a-zA-Z0-9_]|$)',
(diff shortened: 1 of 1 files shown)


ToxicFrog:
I've tested this as much as I can, but I don't have italics enabled in the tty, so I'd be more comfortable about this patch if someone who does could have a look at it and make sure they work as expected.

trygveaa:
Italics is rendering correct for me. As you mentioned in #325 both italics and bold is not working when nested side by side though.

ToxicFrog:
I added #354 to track the issue with side-by-side nesting of formatting codes.

PR #352, opened 2017-04-17
implemented support for colorizing messages

gilbertw1:
Re-implemented support for colorizing messages based on the color of the sender. Uses the existing configuration 'colorize_messages' to toggle on and off.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1790,6 +1790,7 @@ class SlackMessage(object):
         self.submessages = []
         self.thread_channel = None
         self.hash = None
+        self.color = None
         if override_sender:
             self.sender = override_sender
             self.sender_plain = override_sender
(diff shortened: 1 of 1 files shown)


ToxicFrog:
Weechat has a separate script for whole-line colourization (`colorize_lines.pl`) that works with weeslack; we removed message colourization (and the associated configuration option) on the grounds that `colorize_lines` works better.

gilbertw1:
Interesting, I use colorize_lines.pl and it works perfectly in everywhere except for wee-slack buffers which it doesn't color at all (hence the reason I wrote this pull request). 

Is there any additional configuration required to make this work properly?

ToxicFrog:
I don't think so, but I also don't remember what changes I may have made when I installed it. Here's a config that works with weeslack for me:

```
13:31 [var] (plugins.conf)
13:31   plugins.var.perl.colorize_lines.blacklist_buffers = ""
13:31   plugins.var.perl.colorize_lines.buffers = "all"
13:31   plugins.var.perl.colorize_lines.highlight = "on"
13:31   plugins.var.perl.colorize_lines.ignore_tags = "irc_ctcp"
13:31   plugins.var.perl.colorize_lines.lines = "on"
13:31   plugins.var.perl.colorize_lines.nicks = ""
13:31   plugins.var.perl.colorize_lines.own_lines = "on"
13:31   plugins.var.perl.colorize_lines.tags = ""
```
This is on Weechat 1.7.1.

gilbertw1:
I've got the exact same settings (guessing they're defaults) and I'm using Weechat 1.8. I had a friend try it, and he didn't have colorized lines either.

I'll have to do some investigating when I get a chance. For the time being, I'll just roll with my fork.

ToxicFrog:
I wonder if something changed in 1.8? The changelist from 1.7.1 to 1.8 is pretty short, though, and there's no obvious smoking gun there.

gilbertw1:
It's probably unrelated to the Weechat version. Weechat 1.8 was only recently released, and I had this problem (and had created this PR) while using 1.7.

PR #356, opened 2017-04-20
Fix channel history not appearing when an earlier message has 'text': None

ToxicFrog:
Today I learned that sometimes, rather than the key missing, it has the value
None. In this case, (key in dict) is true!

This fixes an issue where, if the backscroll of a channel contained such a
message, it would throw while fetching the backscroll and cut off without
fetching all of it.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1859,8 +1859,8 @@ class SlackMessage(object):
             self.sender, self.sender_plain = senders[0], senders[1]
         self.suffix = ''
         self.ts = SlackTS(message_json['ts'])
-        text = self.message_json.get('text', '')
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -1860,7 +1860,7 @@ def __init__(self, message_json, team, channel, override_sender=None):
         self.suffix = ''
         self.ts = SlackTS(message_json['ts'])
         text = self.message_json.get('text', '')
```
tfheen: Replace this with `….get('text')` too, then? `.get(key)` will default to None if not found.

ToxicFrog: LGTM

Issue #354, opened 2017-04-19
Adjacently nested bold and italic don't work


ToxicFrog:
Specifically, this works: `*bold with _italic_ in the middle*`
But this doesn't: `*_bold and italic combined_*`

The outer format is applied, the inner isn't. If you swap them around it fails in the same way.

This is probably a mistake in the regexes used to convert `*` and `_` into weechat formatting codes.

PR #359, opened 2017-04-23
Improve detection of incoming formatting characters.

ToxicFrog:
This replaces the trailing whitespace match with a non-consuming lookahead
so that strings like '*one* *two*' match correctly.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2458,10 +2458,10 @@ def render(message_json, team, channel, force=False):
         text = text.replace("&lt;", "<")
         text = text.replace("&gt;", ">")
         text = text.replace("&amp;", "&")
-        text = re.sub(r'(^| )\*([^*]+)\*([^a-zA-Z0-9_]|$)',
-                      r'\1{}\2{}\3'.format(w.color('bold'), w.color('-bold')), text)
-        text = re.sub(r'(^| )_([^_]+)_([^a-zA-Z0-9_]|$)',
-                      r'\1{}\2{}\3'.format(w.color('underline'), w.color('-underline')), text)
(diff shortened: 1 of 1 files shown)


tfheen:
Wouldn't it be better to use `\w` here? The existing (and new) regexes don't match non-ASCII word letters like æøå. I'd also really like some tests for this, since it's historically been a part that's brittle and hard to reason about.

ToxicFrog:
@tfheen You're right, `\w` would be better. And while working on this, I'd like to roll a fix for #354 into this PR as well, but it looks like this whole thing may need rethinking.

 After a bunch of experimentation with the slack web UI, it looks like this is more complicated than just "`*` on a word boundary toggles bold, `_` toggles italics":
- `*_*_foo_*_*` is both bold and italic
- `*_*_foo_*_` does nothing
- `*_*_foo_*_ bar*` renders `foo` bold and italic and `bar` bold
- `_*_*foo_*_` renders `*foo` bold and italic

Runs of consecutive identical formatting characters don't do anything, though -- so `_*foo*_` is bold and italic, and so is `_*_*foo*_*_`, but `__foo__` just renders the characters verbatim.

I had  a rummage around in the JS looking for the source for its formatter, but couldn't find it.

Perfectly replicating this may require an actual lexer or something rather than just re.replace(), but I don't know if we want to aim for perfection or just "good enough".

tfheen:
> Perfectly replicating this may require an actual lexer or something rather than just re.replace(), but I don't know if we want to aim for perfection or just "good enough".

If somebody wants to write a proper parser, I'm not going to stop them, but I think that having something which handles most non-pathological cases would be fine and/or enough.

PR #365, opened 2017-05-17
Add a command for opening an editor to compose a multiline message

### LINKED ISSUE #118: Send multiline messages

kattrali:
Borrows heavily from https://github.com/keith/edit-weechat

The current syntax is `/slack longmsg` but maybe there is a better name for the command. Also unsure how to keep the terms of edit-weechat's license.

Fixes #118

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -3264,6 +3264,7 @@ class PluginConfig(object):
         'debug_mode': 'false',
         'debug_level': '3',
         'distracting_channels': '',
+        'editor': '',
         'show_reaction_nicks': 'false',
         'slack_api_token': 'INSERT VALID KEY HERE!',
         'slack_timeout': '20000',
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -17,6 +17,7 @@ import collections
 import ssl
 import random
 import string
+import subprocess
 
 from websocket import create_connection, WebSocketConnectionClosedException
 
@@ -2800,6 +2801,46 @@ def msg_command_cb(data, current_buffer, args):
             channel = team.channels[cmap[channel]]
             channel.send_message(message)
     return w.WEECHAT_RC_OK_EAT
+
+@slack_buffer_required
+def command_longmsg(data, current_buffer, args):
+    """
+    Open an editor to draft a multiline message to be sent
+    as a single input
+    /slack longmsg
+    """
+    data = decode_from_utf8(data)
+    args = decode_from_utf8(args)
+    e = EVENTROUTER
+    team = e.weechat_controller.buffers[current_buffer].team
+    args = args.split(' ')
+    if len(args) > 1 and args[1].startswith('#'):
+        cmap = team.get_channel_map()
+        channel_name = args[1][1:]
+        channel = team.channels[cmap[channel_name]]
+    else:
+        channel = e.weechat_controller.buffers[current_buffer]
+
+    editor = (weechat.config_get_plugin("editor") or
+              os.environ.get("EDITOR", "vim -f"))
+    path = os.path.expanduser("~/.weechat/slack_longmsg.txt")
+    open(path, "w+")
+
+    cmd = editor.split() + [path]
+    code = subprocess.Popen(cmd).wait()
+    if code != 0:
+        os.remove(path)
+        weechat.command(current_buffer, "/window refresh")
+        return weechat.WEECHAT_RC_ERROR
+
+    with open(path) as f:
+        text = f.read()
+        channel.send_message(text)
+
+    os.remove(path)
+    weechat.command(current_buffer, "/window refresh")
+
+    return weechat.WEECHAT_RC_OK
 
 
 @slack_buffer_or_ignore


keith:
Awesome!!! +1

gilbertw1:
This is awesome! I've tested this out locally, and it works well. The one snag I'm seeing is that it's leaving behind lots of artifacts, requiring me to run ```/window refresh```.

It looks like you renamed the ```buf``` variable to ```current_buffer```, so I was able to fix it by replacing:

    ```weechat.command(buf, "/window refresh")```

with:

    ```weechat.command(current_buffer, "/window refresh")```

Also, it appears that the edit-weechat plugin is licensed under MIT, so there shouldn't be any issues there.

kattrali:
@gilbertw1 Ah! You're correct. I renamed the variable a bit incompletely.

oblitum:
Will this overcome https://github.com/weechat/weechat/issues/984??

kattrali:
@oblitum yep!

oblitum:
@kattrali cool!!

oblitum:
@kattrali ah, this will be so awesome for code snippets! Much better than slack itself.

oblitum:
@kattrali I don't know whether it's offtopic on this pull but I just recalled that I was about to request @keith to add the option of extending the edit command to accept the filetype, like `edit cpp` or `edit js`, what you think? I mean, this is for just opening the temporary file with a proper extension instead of `txt`, this avoid the need to call `:set ft js` for a `txt` file in Vim for example.

oblitum:
Plus, for `/edit js`, etc, tripple backticks for code snippets could come implicit, around the temp file contents, so that it's not necessary to add them by hand.

kattrali:
I think those are neat ideas, but maybe assume too much for general usage, @oblitum. If anything, I think that setting your editor command preference to something like `vim -c 'set syntax=markdown' -f` would make the most sense, as you can get syntax highlighting for nested code blocks (via `g:markdown_fenced_languages`) and also compose prose messages easily.

Unrelated, there should probably be a way to handle the case that sending the message fails - at minimum not deleting the cache file - though I'm not sure how to determine whether it worked yet.

oblitum:
@kattrali thanks for your reponse. I don't like the markdown solution, I still have to enter backticks and I won't get real syntax highlighting and syntax checking as when I'd be coding in the actual programming language. That's what is most powerful for me, having Vim started for sending a code snippet and it doing syntax checking, etc, and I'd not even care for those tripple backticks. It doesn't take any generality of the feature because an filetype flag is necessary to behave this way, if it's not passed then it just behaves as it is currently, raw txt that requires the backticks.

keith:
I think automatically backticks when entering this mode is making a pretty big assumption about how this is used. Even my general code usecase doesn't fit in this since I normally do something like:

    here's some block of code I want to share, what do you think:

    ```swift
    // the actual code
    ```

oblitum:
@keith I didn't know slack accepted language specific code fences, if so, this could be made more handy by assuming the implicit fence is the same as the filetype.

oblitum:
I see it may not be everyone's taste to reduce typing, having syntax check, etc, all automatically, I guess I'll have to fork the pull :)

trygveaa:
I like the idea from @oblitum, and since it just adds an additional feature, I don't think it would be a problem in regards to general usage.

For the general usage that @kattrali mentioned, and the example @keith wrote, you would just do `/slack longmsg`, which acts like now. If you want @oblitum's feature you'd rather do e.g. `/slack longmsg js`. So both use cases are preserved.

oblitum:
> @keith I didn't know slack accepted language specific code fences, if so, this could be made more handy by assuming the implicit fence is the same as the filetype.

I just tried and as far as I see, there's no support for language specific code fences in slack. Long thread here: https://twitter.com/slackhq/status/502238121493528576.

keith:
The fencing mentioned is from the vim-markdown syntax plugin 


--
Keith Smiley

oblitum:
@keith ah ok, I think you got me wrong. I don't want backticks in the openned buffer, it's assumed to be implicit, open a markdown buffer and using markdown fences was @kattrali suggestion.

keith:
Right, but like I mentioned if they're implicit this command is only useful if you _only_ use it for code blocks. And you can't have any content in the message that is not part of the block. 


--
Keith Smiley

oblitum:
Correct, and as @trygveaa said, nothing is lost, as it's behind a filetype flag.

oblitum:
I'd rather change it to `/slack edit` too than `/slack longmsg`... :P

kattrali:
This all sounds complicated compared to the initial problem - being able to type a multiline message, including long prose, bulleted lists, and code blocks. ¯\\\_(ツ)_/¯ Assuming everyone's editor is vim sounds like a stretch to begin with. If its a matter of having a syntax-checked buffer, once your editor is open, you can add a buffer with a given file type if you need those features, and copy things into the long message buffer as needed.

trygveaa:
> Assuming everyone's editor is vim sounds like a stretch to begin with

The proposal isn't specific to vim. It just says to open a file with a specified suffix, which would give any editor the ability to do syntax highlighting, and then wrap it in back ticks.

I do agree that it's a bit more complicated though, but not much really. That might be a reason not to do it, at least until this PR is in.

oblitum:
@kattrali I've created a competing pull with a screencast: https://github.com/wee-slack/wee-slack/pull/370. Also noticed that in that pull I've fixed an issue that's currently present in your pull, you must fix the `len` cheat at https://github.com/wee-slack/wee-slack/pull/365/files#diff-67e177ac444a14d6bcd7be6306ab4a20R2817 if you wish the if to pass when the channel is given.

kattrali:
> I've fixed an issue that's currently present in your pull, you must fix the len check

Oh nice catch!

Kareeeeem:
This works geat, added it to my setup. Hope it gets merged.

oblitum:
@Kareeeeem have you tried the competing pull or even check the video? The increase in code in the other pull is completely negligible as well as the changes in the interface usage.

Kareeeeem:
@oblitum. It looks very nice. But I am the odd ball that does not use syntax highlighting or autocomplete for real coding. Much less for quick snippets.

oblitum:
> @oblitum. It looks very nice. But I am the odd ball that does not use syntax highlighting or autocomplete for real coding. Much less for quick snippets.

So you're only caring about your own workflow, which is covered by the other pull anyways, and solely based on that you wish others not having the feature that's not for real coding?

IMO, this request can be ignored.

Kareeeeem:
I made no request, just expressed that *I* enjoy how this PR works as it is. That's all. 

Isn't this the place to express opinions about PRs?

trygveaa:
Thanks for the PR, but we found a more general way of supporting multi line messages. That has been implemented with #430, and now that it's merged you should be able to use the [`edit.py`/edit-weechat](https://github.com/keith/edit-weechat) plugin for the same feature that this PR implements. Therefore, I'm closing this (see that PR for the reason that was merged instead of this). Please test it, and report back if something doesn't work, or if there's something from this PR that can't be done.

PR #370, opened 2017-05-24
Fix #118: Add command to compose multiline message

### LINKED ISSUE #118: Send multiline messages

oblitum:
Borrows heavily from #365.

This adds a `/slack edit [extension] [channel]` command that allows
editing multiline messages in the external editor of choice. If
[extension] is passed, it will be used as the extension of the temporary
file to edit the message, in this mode code backticks are added
implicitly and hence are not needed to be inserted manually. Otherwise,
the default `md` file extension  (for markdown) is assumed and no
backticks are automatically generated.

The [channel] parameter can be used to direct a message to a given
channel.

This is currently competing with pull #365, here's a [preview of the functionality](https://vimeo.com/218750891) (video is outdated regarding the change from default `txt` extension to `md`).

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -17,6 +17,7 @@ import collections
 import ssl
 import random
 import string
+import subprocess
 
 from websocket import create_connection, WebSocketConnectionClosedException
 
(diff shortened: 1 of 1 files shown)


oblitum:
Notice that asciinema swallow angle brackets of the code posted on slackbot.

oblitum:
I've changed the default extension from `txt` to `md` since slack messages are markdown.

dzfweb:
when i prompt `/slack edit` nothing work for me.  I'm using WeeChat 1.9. 
(note: i'm Windows 10 bash user)

dbarrosop:
I tried it and works great :)

trygveaa:
Thanks for the PR, but we found a more general way of supporting multi line messages. That has been implemented with #430, and now that it's merged you should be able to use a modified version [`edit.py`/edit-weechat](https://github.com/keith/edit-weechat) plugin for the same feature that this PR implements. Therefore, I'm closing this (see that PR for the reason that was merged instead of this).

This also avoids the problem of some people wanting the functionality in #365 and some wanting this, as you can use different versions of the `edit.py` plugin to do either.

Please test it, and report back if something doesn't work, or if there's something from this PR that can't be done.

oblitum:
@trygveaa OK, I'll have to do some minor work forking edit-weechat 😒 but I see this was the best path to take, thanks.

oblitum:
For the ones interested, this is the edit-weechat fork that supports the features implemented in this pull (coupled with #430):

- https://github.com/oblitum/edit-weechat

My fork allows for an optional file extension argument for the commands `/edit` and `/fenced`. I've created the `/fenced` command for the automatic fences feature to be apart from the original `/edit` command.

PR #371, opened 2017-05-26
Disable formatting for messages with `mrkdwn` unset

ToxicFrog:
The Slack API spec says we should ignore markdown formatting if this key is unset or false, and I've seen a few (rare) messages where it looks like not supporting this went wrong.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2438,6 +2438,17 @@ def process_reaction_removed(message_json, eventrouter, **kwargs):
 
 ###### New module/global methods
 
+def render_formatting(text):
+    text = re.sub(r'(^| )\*([^*]+)\*([^a-zA-Z0-9_]|$)',
+                  r'\1{}\2{}\3'.format(w.color(config.render_bold_as),
+                                       w.color('-' + config.render_bold_as)),
+                  text)
+    text = re.sub(r'(^| )_([^_]+)_([^a-zA-Z0-9_]|$)',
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -2430,6 +2430,14 @@ def process_reaction_removed(message_json, eventrouter, **kwargs):
 
 ###### New module/global methods
 
+def render_formatting(text):
+    #text = re.sub(r'(^|\W)([*_]+)(\w([*_]+)
```
tfheen: Drop this commented line?

wee_slack.py
```diff
@@ -2458,10 +2466,8 @@ def render(message_json, team, channel, force=False):
         text = text.replace("&lt;", "<")
         text = text.replace("&gt;", ">")
         text = text.replace("&amp;", "&")
-        text = re.sub(r'(^| )\*([^*]+)\*([^a-zA-Z0-9_]|$)',
-                      r'\1{}\2{}\3'.format(w.color('bold'), w.color('-bold')), text)
-        text = re.sub(r'(^| )_([^_]+)_([^a-zA-Z0-9_]|$)',
-                      r'\1{}\2{}\3'.format(w.color('underline'), w.color('-underline')), text)
+        if 'mrkdwn' not in message_json or message_json['mrkdwn'] != False:
```
tfheen: `if message_json.get('mrkdown', True)` ?

ToxicFrog:
Done.

PR #372, opened 2017-05-26
Add descriptions for weeslack plugin settings.

ToxicFrog:
I'm not super happy with the structure here, and ideally I'd like to just factor all the setting-related stuff into its own reusable library (but apparently multi-file plugins cause issues with weeslack?). I don't want to let the perfect be the enemy of the good here, though.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -3226,42 +3226,113 @@ def dbg(message, level=0, main_buffer=False, fout=False):
 
 ###### Config code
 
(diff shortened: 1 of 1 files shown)


ToxicFrog:
Switched to namedtuple after conversation on IRC.

tfheen:
Looks like an improvement. +1 from me.

rawdigits:
👍

PR #384, opened 2017-07-06
Stop polling dead websockets.

### LINKED ISSUE #349: Weechat takes 100% cpu after sleep and wake
I've noticed that after sleeping and waking my MacBook Pro, weechat will take 100% CPU.  Haven't had a chance to investigate more fully, just wanted to record the issue before I forgot.

This is with the NG code.

cmatheson:
This fixes #349 me.


rawdigits:
Oh geez, nicely done! I'm going to merge this right now.

PR #387, opened 2017-07-08
When expanding attachments avoid rendering same link multiple times

dbarrosop:
When links are expanded it sometimes becomes unreadable as the same link is pasted over and over. For example:

link_old

The longer the link the worse it gets. This PR avoids repeating the same link over and over:

link_new

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2281,8 +2281,7 @@ def subprocess_message_changed(message_json, eventrouter, channel, team):
             else:
                 message_json["fallback"] = m["fallback"]
 
-    text_before = (len(new_message['text']) > 0)
-    new_message["text"] += unwrap_attachments(message_json, text_before)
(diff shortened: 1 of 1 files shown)


auscompgeek:
How does this interact with multiple unfurl attachments?

dbarrosop:
I haven't tested with multiple attachments on the same message but I have tested with individual attachments coming from different applications (github, giphy and a few others I have at $DAYJOB) and it basically changed nothing as those apps don't repeat links over and over. Although I haven't tested with multiple attachments on the same message it should behave in the same way regardless of if there is one or multiple as the variables used in the loop are properly initialized on each iteration (famous last words).

trygveaa:
Nice, this is definitively a useful change. I tested with multiple attachments in one message, and it works properly for that as well.

yeled:
@dbarrosop this was my bugbear of the week! So much happier now 👍

trygveaa:
Thanks!

PR #391, opened 2017-07-12
Fix nick coloring in weechat < 1.5

trygveaa:
This should make nick colors work in weechat < 1.5 again by using either the info_name `irc_nick_color_name` or `nick_color_name` depending on the version. It also removes the outdated docks for the colorize_nicks option which doesn't exist anymore.

Replaces #318 since the last comment there never was addressed.

It would be great if someone with weechat 1.3 or 1.4 could verify that this actually fixes the issue.

### DIFF
--- a/README.md
+++ b/README.md
@@ -225,16 +225,6 @@ Example:
 Optional settings
 -----------------
 
-Turn off colorized nicks:
-```
-/set plugins.var.python.slack.colorize_nicks 0
-```
-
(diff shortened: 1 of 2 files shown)


ToxicFrog:
LGTM, but since I don't have a pre-1.5 weechat to test on I'd appreciate a look from someone who does.

jemc:
Works for me on Weechat 1.4

PR #394, opened 2017-07-13
Don't send typing notifications for threads, since Slack doesn't like that

ToxicFrog:
Specifically, if you have debugging on you will get an exception for every single character typed in a thread, which makes debugging anything else kind of aggravating.


auscompgeek:
I think the Slack web client handles this by sending the typing notification to the parent channel. This is a decent fix though.

Issue #396, opened 2017-07-16
Channels with new messages not added to the hotlist on startup


gilbertw1:
Slack very recently updated their rtm.start endpoint to remove 'latest', 'unread_count', and 'unread_count_display' properties from all channel objects.

Wee-slack relies on 'unread_count_display' to determine if there are new unread messages in a channel. Since this value is missing, wee-slack no longer adds any channels to the hotlist on startup.

Details here: https://api.slack.com/changelog/2017-04-start-using-rtm-connect-and-stop-using-rtm-start

tardypad:
Ah! Thanks. Finally the culprit is found.
I was wondering what was happening lately...
I thought this was a problem with my local config or so.

Hopefully this can be fixed soon :pray:, since it is currently "unusable"

gilbertw1:
Hah, I thought the same thing. I even went so far as to start reproducing my config from scratch with an older version of Weechat before I realized the problem might be in wee-slack.

PR #397, opened 2017-07-17
call channel info on channel open to set unread count and update hotlist

gilbertw1:
This fixes the issue raised in #396.

On initial creation of slack channels, I've updated the code to kick off a request to the 'channels.info' for each channel. Upon completion of the info request I've added a handler that will set the unread count on the channel and if the unread count is greater than zero, it will add the channel to the hotlist. 

I've re-used the slack channel property 'unread_count_display' (which no longer exists) so that existing functionality that relied on that property will not fail. In this PR, I've only narrowly focused on the 'unread_count' property of the channel, but this channel info request could probably be used for more in the future.

Thanks!

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -56,6 +56,7 @@ SLACK_API_TRANSLATOR = {
         "join": "channels.join",
         "leave": "groups.leave",
         "mark": "groups.mark",
+        "info": "groups.info"
     },
     "thread": {
         "history": None,
@@ -2086,6 +2087,14 @@ def handle_channelsinfo(channel_json, eventrouter, **kwargs):
     team = eventrouter.teams[request_metadata.team_hash]
     channel = team.channels[channel_id]
     channel.set_unread_count_display(unread_count_display)
+
+def handle_groupsinfo(group_json, eventrouter, **kwargs):
+    unread_count_display = group_json['group']['unread_count_display']
+    group_id = group_json['group']['id']
+    request_metadata = pickle.loads(group_json["wee_slack_request_metadata"])
+    team = eventrouter.teams[request_metadata.team_hash]
+    group = team.channels[group_id]
+    group.set_unread_count_display(unread_count_display)
  
 def handle_groupshistory(message_json, eventrouter, **kwargs):
     handle_history(message_json, eventrouter, **kwargs)
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1243,6 +1243,11 @@ class SlackChannel(object):
             s = SlackRequest(self.team.token, SLACK_API_TRANSLATOR[self.type]["info"], {"channel": self.identifier}, team_hash=self.team.team_hash, channel_identifier=self.identifier)
             self.eventrouter.receive(s)
 
+        if self.type == "im":
+            if "join" in SLACK_API_TRANSLATOR[self.type]:
+                s = SlackRequest(self.team.token, SLACK_API_TRANSLATOR[self.type]["join"], {"user": self.user, "return_im": "true"}, team_hash=self.team.team_hash, channel_identifier=self.identifier)
+                self.eventrouter.receive(s)
+
     def destroy_buffer(self, update_remote):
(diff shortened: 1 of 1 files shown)


tfheen:
This looks great. Need to test it, though. :-)

tardypad:
Thanks for the PR!
I've applied the patch on top of the current master and tried it locally
It is partially working for me

After launching weechat, existing channels that receive a message afterwards are correctly added to the hotlist

But 2 problems:
- on startup, none of the existing channels which have unread messages are added
- it doesn't seem to work for direct messages (neither at startup or after launching)

I couldn't test yet what happens in case a new channel/direct message is created

(I'm using weechat 1.7.1 btw)

gilbertw1:
I've started looking into this and a big problem is that the channels info endpoint only returns successfully for true channels (eg. #general). Separate endpoints have to be hit for group, im, and mpdm channels.

Unfortunately, only group has an info endpoint that can be called identically to the the channel info endpoint. In order to retrieve unread count for im and mpdm channels, we'll need to call the history endpoint with 'unreads' set to true.

gilbertw1:
@tardypad Just added support for groups (this includes private channels). Could you please check it out.

This PR still doesn't cover IMs and MPDMs, but all the public channels, private channels, and groups should be properly added to the hotlist on startup.

tardypad:
Thanks, I've tested it a bit today.
I confirm that now public and private channels with unread messages are added to the hotlist
both at startup and later on while running

I'm not sure though what exactly is a "group" besides public/private channels
I can't find its meaning [there](https://get.slack.help/hc/en-us/articles/213817348-Slack-glossary)
I've only tested with public/private channels to be precise then

gilbertw1:
Cool, the best I can tell is that "group" is synonymous with private channel. The subtext on the api methods page has group endpoints described as "Get info on your team's private channels".

The only gaps is this functionality are now IMs and MPDMs.

gilbertw1:
Does anyone have any feedback on the best way to proceed with this pull request? Specifically, with regards to IMs and MPDMs?

The only way via the slack api to retrieve the ```unread_count_display``` for IMs and MPDMs is to call their respective history endpoints (```im.history``` and ```mpdm.history```) with the ```unreads``` flag set to true. Our two options as I see it are to:

1) Call the history endpoints for IMs and MPDMs on startup to retrieve the ```unread_count_display``` so that we can properly highlight these channels if they have activity AND pre-populate those channels with messages (might as well since we're already calling the history endpoint).

2) Move forward with only highlighting activity on startup for public and private channels, effectively accepting the lack of highlighting for IMs and MPDMs as a limitation.

tfheen:
> The only way via the slack api to retrieve the unread_count_display
> for IMs and MPDMs is to call their respective history endpoints
> (im.history and mpdm.history) with the unreads flag set to true. Our
> two options as I see it are to:

im.open and mpim.open seems to have an unread_display_count, can you use
that?

-- 
Tollef Fog Heen

tardypad:
Small precision regarding what I was saying previously

> it doesn't seem to work for direct messages (neither at startup or after launching)

It is actually working after launching, new direct messages are added to the hotlist,
but somehow it doesn't work anymore when sending a DM to myself from the slack client (which was the way I was testing before)

So in summary, the only things I noticed that are not working are indeed IMs and MPDMs highlighting at startup (plus this DM to self but we can probably ignore that)

gilbertw1:
@tfheen Your right, thanks for pointing that out!

I discounted that method without really looking into it. I'll update this PR with those calls shortly.

gilbertw1:
Okay, I just added the call to ```im.open``` to grab the unread count and handle adding IMs to the hotlist on startup. While debugging MPIM's I noticed that they are emulated as groups by default, as long as we don't pass ```mpim_aware=1``` to ```rtm.start```, so MPIM's are already handled properly because it's being treated as a group.

With this commit, the highlights on startup should be fully functioning as it was before.

tfheen:
Thanks for the patch, merged!

PR #399, opened 2017-07-19
Put users in groups according to presence

### LINKED ISSUE #398: Display user online status (presence)
It would be awesome if there was some way to indicate whether a user is online or not according to slack.

Slack has 3 modes: online, snooze and offline. There are no indicators of who is admin, so maybe we could use the voice (+) and another op flag to indicate presence.

tfheen:
Fixes #398


marceldev89:
Created #403 on top of this PR/branch since it's useless without this one.

PR #400, opened 2017-07-21
Split wee_slack into different files + makefile to concatenate them

ToxicFrog:
No changes to the actual code (yet), although tests have been updated to
use the right import names.

The makefile emits `slack.py` as a build artifact, and includes convenience targets (`make install` and `make enable`) to put it into your weechat plugins directory.

This also addresses PR/#343 and at least part of I/#198


ToxicFrog:
Note: after some discussion on IRC, this is due for a bit of reworking in the hopes of producing something that can be loaded directly with `/python load` without needing to `make` it first (so as not to introduce extra latency while developing) without removing the ability to produce a single-file version for compatibility with the weechat script repo.

trygveaa:
Another problem with this is that it breaks references in IDEs, though the solution for that might be the same as the one for loading it directly.

ToxicFrog:
@trygveaa I suspect so, yeah.

I haven't really had a chance to work on this lately (as you can tell) but I can't forgotten about it. It's not dead, just sleeping.

trygveaa:
I find it weird that `main.py` and `config.py` references lots of stuff that isn't defined or imported in those files, like `EventRouter` and `w`. That makes `pylint` and `flake8` give lots of errors and breaks autocompletion.

ToxicFrog:
I had a go at it with the chainsaw of natural selection, and it should be cleaner now. It's still not remotely lint-clean but I don't think it's *less* lint-clean now than it used to be, and all imports should now be explicit.

Tested: python2 -m pytest; loaded both the exploded and compacted versions in wee-chat and used them for a while.

ToxicFrog:
Unicode issues should be fixed now.

Getting `__future__` imports working in both exploded and imploded mode required an overhaul of the build system anyways, so the `## INCLUDE` lines are gone -- it now figures out where to include files based on `import` statements.

PR #401, opened 2017-07-22
Fix overzealous regex in unfurl_refs

ToxicFrog:
This fixes a bug where messages of the form `@person I think 3 > 2` end up trying to unfurl everything up to the *last* `>` in the message.


trygveaa:
Thanks for the fix. I added a test case for it.

PR #403, opened 2017-07-23
Update nicklist when team member presence changes

marceldev89:
Depends on #399

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1081,6 +1081,14 @@ class SlackTeam(object):
             dbg("Unexpected error: {}\nSent: {}".format(sys.exc_info()[0], data))
             self.set_connected()
 
+    def update_member_presence(self, user, presence):
+        user.presence = presence
+
+        for c in self.channels:
+            c = self.channels[c]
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -2151,7 +2151,15 @@ def process_manual_presence_change(message_json, eventrouter, **kwargs):
...
-    kwargs["user"].presence = message_json["presence"]
+    if "user" in kwargs:
+        user = kwargs["user"]
+        team = kwargs["team"]
+        user.presence = message_json["presence"]
+
+        for c in team.channels:
+            c = team.channels[c]
+            if user.id in c.members:
+                c.update_nicklist(user.id)
```
tfheen: This should probably be a function on the team, just "update presence for this user" and the team is responsible for iterating over any channels the user is in.

marceldev89: Will do. :)

marceldev89: Might as well just move everything to SlackTeam, right?

tfheen:
Thanks for the patch, tested and merged. :+1:

Issue #389, opened 2017-07-11
AttributeError: 'SlackTeam' object has no attribute 'identifier


gdestuynder:
On various commands, such as `/slack status`:

```
File "/home/kang/.weechat/python/autoload/wee_slack.py", line 2906, in slack_command_cb
EVENTROUTER.cmds[function_name]("", current_buffer, args)
File "/home/kang/.weechat/python/autoload/wee_slack.py", line 93, in wrapper
return f(data, current_buffer, *args, **kwargs)
File "/home/kang/.weechat/python/autoload/wee_slack.py", line 3051, in command_status
s = SlackRequest(team.token, "users.profile.set", {"profile": profile}, team_hash=team.team_hash,
channel_identifier=channel.identifier)
AttributeError: 'SlackTeam' object has no attribute 'identifier'
```

Chatting and joining channels and all otherwise works.

- wee-slack master branch (current HEAD 7bf0ea6a4eedc0c98152b3b13f1394676a39d4dd)
- python 2.7.13/3.6.1
- weechat 1.9

auscompgeek:
I suspect `channel_identifier` isn't very useful here.

PR #408, opened 2017-07-26
Fix AttributeError on /slack status in team buffer

auscompgeek:
Fixes: #389
Closes: #390


tfheen:
Looks good, thanks for the fix. :-)

PR #412, opened 2017-07-28
Set correct number of unread messages in hotlist

trygveaa:
While the commit 6718e2f fixed buffers not appearing in the hotlist,
after it the number of new messages for each buffer was gone. This makes
the number of messages for each buffer shown again.


tfheen:
Looks good, thanks for the patch!

PR #414, opened 2017-07-29
Improve rendering of changes in multiline messages

trygveaa:
This improves rendering when multiline messages are changed or reacted to, by updating each line of the existing message instead of cramming all of the new message into the last line. That means that if the number of lines in the message doesn't change, the message will still be rendered correctly after edits or reactions.

If the number of lines change, the rendering is still improved, however I don't know if it is possible to insert or delete lines in a weechat buffer without re-rendering the whole buffer, so it won't be perfect. If the number of lines of a message decreases after an edit, the extra lines will still be present, but they will be blanked. If the number of lines increases, the first lines will be set as normally, but the last line will contain all of the rest of the lines.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2716,6 +2716,8 @@ def modify_buffer_line(buffer, new_line, timestamp, time_id):
         # hold the structure of a line and of line data
         struct_hdata_line = w.hdata_get('line')
         struct_hdata_line_data = w.hdata_get('line_data')
+        # keep track of the number of lines with the matching time and id
+        number_of_matching_lines = 0
 
         while line_pointer:
             # get a pointer to the data in line_pointer via layout of struct_hdata_line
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -2694,6 +2694,8 @@ def modify_buffer_line(buffer, new_line, timestamp, time_id):
         # hold the structure of a line and of line data
         struct_hdata_line = w.hdata_get('line')
         struct_hdata_line_data = w.hdata_get('line_data')
+        # keep track of the number of lines with the matching time and id
+        number_of_matching_lines = 0
 
         while line_pointer:
             # get a pointer to the data in line_pointer via layout of struct_hdata_line
```
trygveaa: (This comment should be two lines below, but it doesn't seem to be possible to comment there)

The case when data is falsy isn't handled here now. The documentation of `hdata_pointer` doesn't state that it can return `NULL`/`None`, so I'm not sure why the `if` is there at all? I can add handling of it, but it will make the code a bit more complex.

trygveaa:
I'm also considering redrawing the whole buffer if the number of lines in a multiline message changes, to fix the remaining issue. I would think that doesn't happen that often so it wouldn't cause performance issues. Any thoughts?

tfheen:
I think redrawing should be fine if the number of lines change. I think the most common case there is slack expanding links, so we could special-case if it's the last message in the buffer if the performance cost is large(ish)?

trygveaa:
Ah, yes. As long as you can add new lines without a prefix, so they still "belong" to the same message.

For the record, before wee-ng I ran a patched version for a couple of months which redrew everything each time a multiline message was changed. I didn't notice any performance issues, however I have seen that some of you have much more users and activity than there are on the teams I participate in. @rawdigits said he saw major performance issues when doing that, but a large cause of that was because of reactions, and if we only do it when the number of lines changes, it won't happen on reactions.

wee_slack.py
tfheen: i don't know why there's an if there. Maybe we should have a lightweight assert-like functionality (but one which doesn't stop weechat completely) for invariants that must be true for us to continue?

tfheen:
I think coding it up and we can merge it and if people complain about performance we take it out again? Especially if it's in its own commit, it should be easy enough.

wee_slack.py
trygveaa: Good idea. I added some error handling so the modification will be aborted if it happens, instead of something unexpected happening. Please tell me if you want to log differently, or have a better error message. Possibly log to a more visible place than debug?

trygveaa:
> I think coding it up and we can merge it and if people complain about performance we take it out again?

Yeah, I agree. I'll see if I can do it, either an additional commit here, or a new PR if this is merged.

trygveaa:
While doing this I realized what's causing bare display mode to be broken in some cases. It seems to be because we are calling `hdata_update` with a string containing newlines. When doing this the newlines are printed in bare mode, but extra lines are not inserted, so the lines are overlapping with each other.

I fixed this by replacing the newlines with " | ". When I implement redrawing the whole buffer when the number of lines in a message changes, this won't be useful anymore, but I wanted it in the git history in case the redrawing causes performance issues and has to be changed.

tfheen:
While not perfect, I think this is an improvement, so let's merge this.

PR #417, opened 2017-08-01
Fix HTML entity-encoding of outgoing messages

trygveaa:
As pointed out by @auscompgeek on IRC, the HTML entity-encoding is broken. If you send e.g. `<test>`, it will be printed as `&amplttest&ampgt` in wee-slack and `&lttest&gt` in the web client.

We have to replace the & before we replace < and >, otherwise the & in &lt; and &gt; are going to be replaced. Additionally, we need to end the sequences with ; which was missing.


trygveaa:
The test setup seems a bit broken, but I added a test for this.

PR #423, opened 2017-08-16
Check that the bot_id exists before using it

### LINKED ISSUE #420: Using Reminder Slackbot stops further message from displaying
If we use reminder slackbot, no further messages show up in history after that.

trygveaa:
Instead of just checking that bot_id is not None, check that it exists
in the list of bots. This prevents a KeyError when trying to access a
bot that doesn't exist in the list.

This KeyError would happen if you use a reminder. The message from
slackbot then has bot_id B01 which is not in the list of bots from
rtm.start and gives bot_not_found if trying to look it up with the
bots.info API endpoint.

By ignoring the bot_id we fall back to the user/username, which in the
case of reminders is slackbot, so the name appears just as the other
messages from slackbot.

Fixes #420

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1911,7 +1911,7 @@ class SlackMessage(object):
     def get_sender(self):
         name = ""
         name_plain = ""
-        if 'bot_id' in self.message_json and self.message_json['bot_id'] is not None:
+        if self.message_json.get('bot_id') in self.team.bots:
             name = "{} :]".format(self.team.bots[self.message_json["bot_id"]].formatted_name())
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -1911,7 +1911,7 @@ def change_suffix(self, new_suffix):
     def get_sender(self):
         name = ""
         name_plain = ""
-        if 'bot_id' in self.message_json and self.message_json['bot_id'] is not None:
+        if 'bot_id' in self.message_json and self.message_json['bot_id'] in self.team.bots:
```
tfheen: Maybe even better written as `if self.message_json.get('bot_id') in self.team.bots` ?

trygveaa: Yeah, that's better. Fixed.

Issue #424, opened 2017-08-17
Attachment colors


Informatic:
It'd be nice to have some sort of implementation of attachment colors. (either full line colorization, or some sort of colorized prefix character)

I'll take a look into that when i have some time.

PR #426, opened 2017-08-17
Colorize attachment messages

Informatic:
This implements #424 feature request according to https://api.slack.com/docs/message-attachments#color

Reported attachment color doesn't seem to contain leading `#` character, but this has to be checked. So far I only tested it with `slackSend` Jenkins plugin. (while providing leading `#`) I need someone to verify if `good`/`warning`/`danger` codes are reported as hex color as well.

New configuration option `plugins.var.python.slack.colorize_attachments` allows user to disable attachment colorization and bring back previous behaviour.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2613,6 +2613,23 @@ def unhtmlescape(text):
                .replace("&amp;", "&")
 
 
+def rgb2term_prefix(color):
+    if color:
+        try:
+            # Strangely enough color_rgb2term requires hex color as string
+            # representation of 0xRRGGBB in base-10...
+            color_int = str(int(color.lower(), 16))
+            color_code = w.info_get('color_rgb2term', color_int)
+        except ValueError:
+            # In case color is not valid hex representation, just use it as is
+            # weechat should gracefuly handle most basic colors (yellow, blue,
+            # ...) or fall back to defaults
+            color_code = color.lower()
+
+        return w.color(color_code)
+    return ''
+
+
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -2613,6 +2613,23 @@ def unhtmlescape(text):
...
+            # representation of 0xRRGGBB in base-10...
+            color_int = str(int(color.lower(), 16))
+            color_code = w.info_get('color_rgb2term', color_int)
+        except ValueError:
+            # In case color is not valid hex representation, just use it as is
+            # weechat should gracefuly handle most basic colors (yellow, blue,
+            # ...) or fall back to defaults
+            color_code = color.lower()
+
+        return w.color(color_code)
```
bendem: Shouldn't this return `'<colour>|<reset> '`? I wouldn't want the whole attachment to be green, only the prefix (like it is in the actual slack client). Also, some attachment colours would be very unreadable (I'm thinking the jira dark blue will probably translate to dark blue, which is unreadable in dark terminals).

Informatic: TBH this whole part could use weechat format expressions, as I actually personally prefer it this way.
I was thinking about adding enum-like configuration for that (`prefix`/`message`/`disable`) as a temporary solution... (but got into doing other things...)

trygveaa:
Sorry for never responding to this. I added support for choosing to colorize either the prefix or the whole line and have committed it to master.

PR #430, opened 2017-09-04
Add support for sending multiline messages

### LINKED ISSUE #118: Send multiline messages

trygveaa:
While weechat plugins like multiline.pl and edit.py allow you to compose
multiline messages, wee-slack normally receives them line by line, so it
ends up as separate messages. There isn't any direct support for
receiving the complete message from weechat, but it can be done with
sort of a workaround.

There is a callback you can hook on to which allows you to intercept and
edit messages before they are sent to the buffer. This callback receives
the complete message with all the lines. By using this we can process
the whole message, and return an empty string from the callback which
means that the message will not be processed further.

Note that when this happens, weechat commands are not processed, which
is why we only use the callback for messages that contain a newline and
does not start with slash. Otherwise, we are returning the string sent
to the callback, which would be the same as the callback not being
hooked in.

I've also opened a PR in weechat for supporting multiline input
properly[0]. If that is merged, that can be used for newer weechat
versions instead of this.

[0]: https://github.com/weechat/weechat/pull/1063

Fixes #118

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -637,6 +637,19 @@ def buffer_input_callback(signal, buffer_ptr, data):
     return w.WEECHAT_RC_OK
 
 
+# Workaround for supporting multiline messages. It intercepts before the input
+# callback is called, as this is called with the whole message, while it is
+# normally split on newline before being sent to buffer_input_callback
+def input_text_for_buffer_cb(data, modifier, current_buffer, string):
+    if current_buffer not in EVENTROUTER.weechat_controller.buffers:
+        return string
+    message = decode_from_utf8(string)
+    if "\n" in message and not message.startswith("/"):
+        buffer_input_callback("EVENTROUTER", current_buffer, message)
+        return ""
+    return string
+
+
 def buffer_switch_callback(signal, sig_type, data):
     """
     incomplete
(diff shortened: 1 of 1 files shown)


trygveaa:
The advantage of this over #365 and #370 is that the user can type/send the multiline message in whatever way they wishes, i.e. with the `multiline.pl` or `edit.py` plugins. So the persons that want it the way #365 does it can get that with `edit.py`, and the persons that want they way #370 does it can get it with a modified `edit.py` plugin.

wee_slack.py
```diff
@@ -637,6 +637,19 @@ def buffer_input_callback(signal, buffer_ptr, data):
...
 
 
+# Workaround for supporting multiline messages. It intercepts before the input
+# callback is called, as this is called with the whole message, while it is
+# normally split on newline before being sent to buffer_input_callback
+def input_text_for_buffer_cb(data, modifier, current_buffer, string):
+    if current_buffer not in EVENTROUTER.weechat_controller.buffers:
+        return string
+    message = decode_from_utf8(string)
+    if "\n" in message and not message.startswith("/"):
```
auscompgeek: Just a really minor nitpick: the `message.startswith('/')` check should've probably gone first, since that's a constant-time check. Not that we _really_ care much about performance anyway.

trygveaa: Right, that's true. Though, this only runs on user input, so I don't really think it matters. But since the readability is the same either way, I changed it.

PR #434, opened 2017-09-15
Fix broken tests

trygveaa:
This fixes the tests that are broken in master, and a few other minor changes.

@tfheen / @rawdigits: Want to enable Travis for this repo? I think you only need to flick the switch at https://travis-ci.org/profile/wee-slack (only admins can do so).


tfheen:
I've flipped the travis switch now too.

PR #435, opened 2017-09-16
Print error message when initial connection fails

### LINKED ISSUE #432: No error shown when incorrect Slack API Token provided
If an incorrect, or no, Slack API Token is provided in the `plugins.var.python.slack.slack_api_token` option then nothing appears to happen following the `/python reload` cmd.

No error or activity messages are displayed making it hard for the user to troubleshoot the problem.

It would be helpful if an error message could be displayed indicating the missing or incorrect Slack API Token so that the user could resolve the issue.

trygveaa:
I chose to print this instead of using dbg since it's a rather critical
error message.

Fixes #432

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2033,78 +2033,78 @@ def handle_rtmstart(login_data, eventrouter):
     """
     metadata = pickle.loads(login_data["wee_slack_request_metadata"])
 
-    if login_data["ok"]:
-        # Let's reuse a team if we have it already.
-        th = SlackTeam.generate_team_hash(login_data['self']['name'], login_data['team']['domain'])
-        if not eventrouter.teams.get(th):
-
-            users = {}
-            for item in login_data["users"]:
-                users[item["id"]] = SlackUser(**item)
-                # users.append(SlackUser(**item))
-
-            bots = {}
-            for item in login_data["bots"]:
-                bots[item["id"]] = SlackBot(**item)
-
-            channels = {}
-            for item in login_data["channels"]:
-                channels[item["id"]] = SlackChannel(eventrouter, **item)
-
-            for item in login_data["ims"]:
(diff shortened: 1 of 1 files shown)


selckin:
+1, one of my slacks stopped connecting without any message, with this applied shows token_revoked

wee_slack.py
```diff
@@ -2103,6 +2102,10 @@ def handle_rtmstart(login_data, eventrouter):
 
         dbg("connected to {}".format(t.domain))
 
+    else:
+        w.prnt("", "ERROR: Failed connecting to Slack with token {}: {}"
+            .format(metadata.token, login_data["error"]))
+
```
tfheen: Hanging `else` clauses are a bit icky.  I'd much rather have it reverse the check for `login_data["ok"]` and just returning early.  This'll require a reindent, but it's much clearer to read.

trygveaa: Sure, that's better.

wee_slack.py
```diff
@@ -2031,77 +2031,80 @@ def handle_rtmstart(login_data, eventrouter):
...
+    # try:
+    #    ws = create_connection(web_socket_url, sslopt=sslopt_ca_certs)
+    #    w.hook_fd(ws.sock._sock.fileno(), 1, 0, 0, "receive_ws_callback", t.get_team_hash())
+    #    #ws_hook = w.hook_fd(ws.sock._sock.fileno(), 1, 0, 0, "receive_ws_callback", pickle.dumps(t))
+    #    ws.sock.setblocking(0)
+    #    t.attach_websocket(ws)
+    #    t.set_connected()
+    # except Exception as e:
+    #    dbg("websocket connection error: {}".format(e))
+    #    return False
```
tfheen: Maybe remove this too, now that you're already making changes here?  Commented-out blocks seems pointless.

trygveaa: Done.

PR #436, opened 2017-09-17
Update list of emoji completions

trygveaa:
This updates the list of emojis in `weemoji.json` to the emojis currently available in the Slack web client. It was done using this command:

```
curl -s 'https://cfr.slack-edge.com/bv1-1/emoji.84b5140d201ea94c81b4.min.js' | \
        sed -e 's/.*emoji.prototype.data = \({[^}]*}\).*/\1/' \
        -e 's/\\n/\n/g' -e 's/\\t/\t/g' -e 's/\\'/'/g' | \
        jq '{emoji: [to_entries | .[].value[3]] | flatten | sort}'
```

This does not use the API (as far as I can see, there doesn't exist any API call for getting this list, the `emoji.list` call only returns custom emojis) and is pretty fragile to changes, so we can't use it to automatically load the list from `wee_slack.py`.


trygveaa:
It's in the commit message at least. I could add it to the readme, but it's not guaranteed to work in the future, and it's not really useful for the users.

tfheen:
I wasn't entirely sure, I guess the commit message is fine (at least for now).

PR #437, opened 2017-09-17
Add completion of custom emojis

### LINKED ISSUE #431: IOError on initial installation
After following the installation instructions in the ReadMe (copy wee_slack.py to the python/autoload directory) nothing appears to happen when following the /python reload cmd.

Enabling debug results in seeing an error - `IOError(2, 'No such file or directory')`.

No indication of what file is missing is given making it difficult to trouble shoot.

The missing file was the weemoji.json file. After copying this into the .weechat directory the error no longer occurred.

It would be helpful if an extra line could be added to the install instructions detailing the requirement to copy the weemoji.json file as well as the wee_slack.py script.

It would also be useful if the error message included some extra details about which file is missing.

trygveaa:
This adds a few enhancements to the emoji completion:
 - It adds custom emojis to the emoji completion list for each team.
 - If `weemoji.json` is not found, it gives a better error message (Fixes #431)
 - It adds documentation for enabling tab completion of emojis.

### DIFF
--- a/README.md
+++ b/README.md
@@ -226,6 +226,14 @@ Example:
 /slack status :ghost: Boo!
 ```
 
+#### Emoji tab completions
+
+To enable tab completion of emojis, copy or symlink the `weemoji.json` file to your weechat config directory (e.g. `~/.weechat`). Then append `|%(emoji)` to the `weechat.completion.default_template` config option, e.g. like this:
+
+```
+/set weechat.completion.default_template "%(nicks)|%(irc_channels)|%(emoji)"
+```
+
 Optional settings
 -----------------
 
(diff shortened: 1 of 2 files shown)


tfheen:
Seems reasonable. Can we add hooks to pick up any new emojis added while wee-slack is running?

trygveaa:
Right, that would be a good idea. Seems like https://api.slack.com/events/emoji_changed can be used. I'll look into it.

Do you think we should parse the content of those events, or just call `load_emoji_completions` again? This note under the remove section makes me think that we should just call `load_emoji_completions` again so we don't have to think about aliases (we don't store if an emoji is an alias or not now):

> Note [...] that aliased emoji entries are always removed when the emoji name they alias to is removed.

tfheen:
Should be rare enough for emojis to be added or removed that we can just reload the bunch. If it becomes a burden we can redo it later.

trygveaa:
@tfheen: Done.

PR #440, opened 2017-09-23
Use the timestamp of the last message in mark_read

### LINKED ISSUE #307: Messages can be marked as read without having been seen
**Summary:**

The problem here is that channels get marked as being read asynchronously, and incorrectly use the current time rather than the actual time of the last read message in the buffer.

**To reproduce:**

* A slow internet connection will help a lot
* Be in a buffer.
* Have a friend type a message in that channel/im/group, but don't press Enter yet.
* Type Alt+1 (or similar) to switch to a different buffer.
* Immediately thereafter have your friend press Enter to send their message.
* The activity list will show the previous buffer come active, and then immediately be cleared when the buffer is marked as read.

**The net effect of this is that your friend's message will have been marked as read (including in the native Slack interface) even though you have never seen it.  If you didn't see the activity list flash (which is likely) then you may not ever be aware that it was sent.**

I have had a few occurrences of this, but they were vague and strange and at the time I couldn't rule out simpler explanations like just forgetting about having seen the message.

However since then I have had a very clear and explicit situation where this has happened, and where I saw the activity list flash as described above, and then immediately went back to the buffer to see the "missing" message (which had been marked as read, and which I would never have otherwise seen).

**Diagnosis:**

The problem is the use of `time.time()` on [wee_slack.py:662](https://github.com/wee-slack/wee-slack/blob/01beb42fec5c6dbfc0d8c53ce1de88b48b48f6db/wee_slack.py#L662).  This should instead be the timestamp of the last message in the buffer as weechat understands it.

Alternatively, the time should be grabbed (synchronously) when the user navigates away from the channel (or whatever other action causes the read marker to be updated), and this timestamp should then be passed in to `Channel.mark_read()` as a required argument.

trygveaa:
When sending a mark read message to slack, use the timestamp of the last message in the channel, instead of the current timestamp. This should prevent potentially marking messages as read, if they arrive after mark_read was called and the clock of the slack server and the machine wee-slack runs on is not exactly in sync.

According to issue #307 messages might also be marked as read if they arrive right after you switch buffer, and you have a slow internet connection. Though, since the time used in mark_read is set synchronously when the buffer is changed, I'm not sure if that can be the case. Either way, I think this patch should fix the issue.

If there are no messages in the channel, fall back to using the current time as timestamp.

Fixes #307

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1445,7 +1445,8 @@ class SlackChannel(object):
 
     def mark_read(self, ts=None, update_remote=True, force=False):
         if not ts:
-            ts = SlackTS()
+            msg_timestamps = self.sorted_message_keys()
+            ts = msg_timestamps[-1] if msg_timestamps else SlackTS()
         if self.new_messages or force:
             if self.channel_buffer:
                 w.buffer_set(self.channel_buffer, "unread", "")


wee_slack.py
```diff
@@ -1445,7 +1445,8 @@ def get_typing_list(self):
 
     def mark_read(self, ts=None, update_remote=True, force=False):
         if not ts:
-            ts = SlackTS()
+            msg_timestamps = self.sorted_message_keys()
```
tfheen: How expensive is this call? Should we rather just cache this on the channel object when appending messages?

trygveaa: Valid point. I have now changed `SlackChannel.messages` to be an OrderedDict and sort that when we store the messages, so we don't need to sort it on lookup.

wee_slack.py
```diff
@@ -1326,14 +1328,14 @@ def store_message(self, message, team, from_me=False):
...
-        if len(self.messages.keys()) > SCROLLBACK_SIZE:
-            mk = self.messages.keys()
-            mk.sort()
-            for k in mk[:SCROLLBACK_SIZE]:
-                msg_to_delete = self.messages[k]
-                if msg_to_delete.hash:
-                    del self.hashed_messages[msg_to_delete.hash]
-                del self.messages[k]
+
+        sorted_messages = sorted(self.messages.items())
```
tfheen: Isnt't this already sorted?

trygveaa: I guess it depends on if `store_message` can be called out of order or not. `OrderedDict` sorts by insertion order, so if this is called with a ts that is less than any of the existing messages, it won't be sorted. I think I put `sorted` here to be sure of that.

tfheen: ok, go ahead then. :-)

PR #443, opened 2017-09-25
fix: Make /topic behave the same way as on IRC

### LINKED ISSUE #405: /topic with no argument clears the topic
Being used to how it works in IRC I'd expect it to show the topic instead of clearing it.

trygveaa:
This refactors the topic command, and makes it behave the same way as it does on IRC. The fixes include:

- `/topic` prints the topic instead of clearing it
- `/topic #channel` prints the topic for that channel in that channels buffer instead of setting the topic to #channel
- `/topic #channel [...]` for a channel that doesn't exist gives a proper error message

The patch also removes some unused code, refactors the logic and removes the `/slack topic` command, since using `/topic` should suffice.

The reason it returs WEECHAT_RC_OK_EAT when the channel isn't found, is so weechat doesn't print an error message such as:

    irc: command "topic" must be executed on irc buffer (server or channel)

Fixes #405

### DIFF
--- a/_pytest/conftest.py
+++ b/_pytest/conftest.py
@@ -49,7 +49,9 @@ class FakeWeechat():
     this is the thing that acts as "w." everywhere..
     basically mock out all of the weechat calls here i guess
     """
-    WEECHAT_RC_OK = True
+    WEECHAT_RC_ERROR = 0
+    WEECHAT_RC_OK = 1
+    WEECHAT_RC_OK_EAT = 2
 
     def __init__(self):
         pass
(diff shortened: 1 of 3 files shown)


tfheen:
I think this looks good, but can we have some tests? There are a few nested cases here.

trygveaa:
Definitely. I've split out the parsing of the topic command to a separate function, since that can be tested without any mocking, and is easy to check the result of, and added tests for that. Do you think we should add tests for the `topic_command_cb` function as well?

tfheen:
If it's easy to do, sure. If not, it looks simple enough that I think we can go ahead with this.

trygveaa:
I've added tests for `topic_command_cb` as well now, so now everything should be covered. Some mocking were necessary for those tests though.

PR #444, opened 2017-09-26
feat: Remove Slacks auto prefixing of url protocols

trygveaa:
When you write multiple words separated by dots, or email adresses, Slack will turn those into links. Wee-slack will display this as `text (url)`, where url is just the text with http:// or mailto: prefixed. This makes the text hard to read, and might not even make sense (e.g. when writing server names or user@hostname).

When you receive such an url, the alt text of the url will be the original text. This means that we can check if the url is just the alt text with a protocol prefixed. If it is, we should just print the alt text and ignore the url.

I guess some people may want the urls, so they are able to click the links. And since it removes information coming from slack, I made it into an option instead of making it the default.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2587,10 +2587,12 @@ def unfurl_ref(ref, ignore_alt_text=False):
                 display_text = ref.split('|')[1]
             else:
                 url, desc = ref.split('|', 1)
-                if (config.unfurl_prevent_auto_linking and
-                        match_url = r"^\w+:(//)?{}$".format(re.escape(desc))
-                        re.match(match_url, url)):
+                match_url = r"^\w+:(//)?{}$".format(re.escape(desc))
+                url_matches_desc = re.match(match_url, url)
+                if url_matches_desc and config.unfurl_auto_link_display == "text":
                     display_text = desc
+                elif url_matches_desc and config.unfurl_auto_link_display == "url":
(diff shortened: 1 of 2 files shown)


wee_slack.py
```diff
@@ -2587,7 +2587,11 @@ def unfurl_ref(ref, ignore_alt_text=False):
                 display_text = ref.split('|')[1]
             else:
                 url, desc = ref.split('|', 1)
-                display_text = "{} ({})".format(url, desc)
+                if (config.unfurl_prevent_auto_linking and
+                        re.match(r"^\w+:(//)?{}$".format(desc), url)):
```
tfheen: Maybe use `re.escape(desc)` to ensure regex metacharacters aren't treated specially?

auscompgeek:
Could we also allow for an option to hide the other way (i.e. hide the user-typed URL if Slack auto-prefixed the link with the relevant protocol)?

wee_slack.py
trygveaa: Definitely, should have thought of that.

trygveaa:
> Could we also allow for an option to hide the other way (i.e. hide the user-typed URL if Slack auto-prefixed the link with the relevant protocol)?

Sure, I've added support for that now, and changed the option (and the help text) so you can choose between the three alternatives.

Issue #379, opened 2017-06-23
`/slack channels`, `/slack users` not working


jessebye:
Using commands `/slack channels` as well as `/slack users` results in the following error:  `Command not found: channels` or `Command not found: users`.

auscompgeek:
These existed before the wee-ng refactor. Copying them from before that will probably work.

hshin23:
`/slack join` doesn't work either. Is there a way to resolve this issue?

dzfweb:
@auscompgeek  where is the wee-ng code which i could copy?

dzfweb:
i fixed on my repository #https://github.com/dzfweb/wee-slack/commit/796894e1b9d45c5d1e830afd548e6b5f442ae52e 

already open a PR for it and other fixes #https://github.com/wee-slack/wee-slack/pull/386

tfheen:
Please don't intermingle fixes in PRs, it makes it very hard to review and selectively merge bugfixes and discuss features separately.

trygveaa:
This has now been fixed with #449.

PR #447, opened 2017-10-03
Update documentation on enabling debug mode

ericdwang:
Wanted to debug some changes and noticed that the `/slack debug` command doesn't exist anymore and instead debug mode is enabled via the `plugins.var.python.slack.debug_mode` setting. Guessing this is related to the refactoring mentioned in #379. If the command is being added back though then my bad, feel free to close this PR.


trygveaa:
Maybe mention `plugins.var.python.slack.debug_level` as well?

ericdwang:
Sure thing! Added.

trygveaa:
Thanks!

PR #449, opened 2017-10-05
Created "/slack users" and "/slack channels" command

dzfweb:


.vscode/settings.json
```diff
@@ -0,0 +1,3 @@
+{
+  "python.linting.pylintEnabled": false
+}
```
trygveaa: Why is this file added?

wee_slack.py
```diff
@@ -2925,6 +2925,26 @@ def msg_command_cb(data, current_buffer, args):
             channel.send_message(message)
     return w.WEECHAT_RC_OK_EAT
 
+@slack_buffer_required
+def command_channels(data, current_buffer, args):
```
trygveaa: Please decode data and args from utf8 byte strings to unicode strings like it is done in the other functions. This goes for `command_users` as well. I.e. like this:

```
     data = decode_from_utf8(data)
     args = decode_from_utf8(args)
```

wee_slack.py
```diff
@@ -2925,6 +2925,26 @@ def msg_command_cb(data, current_buffer, args):
...
+    e = EVENTROUTER
+    team = e.weechat_controller.buffers[current_buffer].team
+
+    for channel in team.get_channel_map():
+        line = "{:<25}".format(channel)
```
trygveaa: Why is this string padded when there are only one value printed?

wee_slack.py
```diff
@@ -2925,6 +2925,26 @@ def msg_command_cb(data, current_buffer, args):
...
+        team.buffer_prnt(line)
+    return w.WEECHAT_RC_OK_EAT
+
+@slack_buffer_required
+def command_users(data, current_buffer, args):
+    e = EVENTROUTER
+    team = e.weechat_controller.buffers[current_buffer].team
+
+    for user in team.get_username_map():
+        line = "{:<25}".format(user)
```
trygveaa: Might be nice to include the `presence` field from the user as well, to show if the user is active or away.

wee_slack.py
dzfweb: @trygveaa  i need to decode data even when i'm not using this parameter?

dzfweb:
Please, review the last changes i maded. I'll include the userss presence later because i need to change the way i'm getting the users list, once `get_username_map` return just the string names.

wee_slack.py
trygveaa: The reason I asked for this is that it's easy to forget if the functions were changed to use the arguments later. But it's inconvenient and ugly, so I created a PR with a decorator for it instead. So you don't need to do anything in regards to this.

wee_slack.py
```diff
@@ -2925,6 +2925,26 @@ def msg_command_cb(data, current_buffer, args):
...
+    for channel in team.get_channel_map():
+        team.buffer_prnt(channel)
+    return w.WEECHAT_RC_OK_EAT
+
+@slack_buffer_required
+def command_users(data, current_buffer, args):
+    e = EVENTROUTER
+    team = e.weechat_controller.buffers[current_buffer].team
+
+    team.buffer_prnt("Channels:")
```
trygveaa: This should print "Users", not "Channels".

wee_slack.py
```diff
@@ -2925,6 +2925,26 @@ def msg_command_cb(data, current_buffer, args):
...
     return w.WEECHAT_RC_OK_EAT
 
+@slack_buffer_required
+def command_channels(data, current_buffer, args):
+    e = EVENTROUTER
+    team = e.weechat_controller.buffers[current_buffer].team
+
+    team.buffer_prnt("Channels:")
+    for channel in team.get_channel_map():
+        team.buffer_prnt(channel)
```
trygveaa: I would still like these a bit indented, so you can see what they belong to and where they end. I.e. include four spaces before the channel names.

trygveaa:
I have reviewed the last changes, and added some new comments. You can iterate over the users with `team.users.values()` instead of `get_username_map` to get the presence as well.

dzfweb:
Changes maded

wee_slack.py
```diff
@@ -2925,6 +2925,26 @@ def msg_command_cb(data, current_buffer, args):
             channel.send_message(message)
     return w.WEECHAT_RC_OK_EAT
 
+@slack_buffer_required
+def command_channels(data, current_buffer, args):
+    e = EVENTROUTER
+    team = e.weechat_controller.buffers[current_buffer].team
+
+    team.buffer_prnt("    Channels:")
```
trygveaa: I meant add spaces before each of the channel/user names, so you can see where the list of channels/users end when more lines are printed. Like this:

```
Channels:
    channel1
    channel2
```

dzfweb: sorry, I had misunderstood @trygveaa, i think now it is ok.

trygveaa: Now you added padding after the names, not before. That's fine for the users as the presences are now aligned, but not much point for channels where only one value is printed.

But what I meant was adding spaces before the names, so they are indented.

trygveaa:
I've merged this to master now. I didn't want all the unrelated commits here in the git history, so I squashed them to one commit. That is why this is marked as closed and not merged.

Thanks!

Issue #377, opened 2017-06-19
Trouble navigating threads + channel list


jessebye:
Not sure if I'm missing something, but after I do a `/thread xyz` the thread doesn't open in the current buffer. I have to try to jump to it using `Alt` + `A` which doesn't always open it right away. And if I navigate away from the thread to another Slack channel, I don't know of a way to get back to the thread unless there is activity in the thread.

Another problem is that I can't find a way to scroll through the list of buffers (i.e. channels), so all I can see are the top ~20 channels. Thus using `Alt` + `J` + `01`...`99` I have to randomly guess which channel I'm going to open.

Really hoping I'm doing this wrong and there's a better way.

auscompgeek:
I agree with `/thread` not switching to the new buffer is annoying.

To answer the rest of your concerns, you should read the weechat FAQ and user guide.

trygveaa:
> Another problem is that I can't find a way to scroll through the list of buffers (i.e. channels), so all I can see are the top ~20 channels. Thus using Alt + J + 01...99 I have to randomly guess which channel I'm going to open.

I'm using [go.py](https://weechat.org/scripts/source/go.py.html/), which lets you go to a channel by (parital) name.

> Not sure if I'm missing something, but after I do a /thread xyz the thread doesn't open in the current buffer. I have to try to jump to it using Alt + A which doesn't always open it right away.

The number for the channel appears in the hotlist, so you can jump to that number with alt-j. Or if the number is above 99, you can use go.py and type the number. Though, I do agree that it should open automatically when running `/thread`.

> And if I navigate away from the thread to another Slack channel, I don't know of a way to get back to the thread unless there is activity in the thread.

For this, I also use go.py. Just type the channel name, and the channel as well as the threads for that channel will be listed.

jessebye:
@auscompgeek @trygveaa thanks for the tips, I discovered go.py shortly after and that has helped a lot! I would still love to be able to scroll the hotlist, seems like that would be built-in to weechat? I guess not.

Threads are still troublesome, because I don't always see when someone has replied to a thread that I am following. For example, I would not know if someone posted to a thread if:

1. I join a thread.
2. Later, I close weechat.
3. Someone replies to the thread.
4. I re-open weechat.
5. The thread is no longer open, so I have no way to know that anyone posted a reply to it. I would have to explicitly check for new messages in the thread by finding the original message with the thread ID and rejoining the thread.

[...]

PR #450, opened 2017-10-06
Switch to thread buffer on open if configured

auscompgeek:
This fixes `/thread` to obey the `switch_buffer_on_join` setting.

Ref: #377


trygveaa:
Thanks!

PR #451, opened 2017-10-09
Add mpim API type for multiparty DMs

ericdwang:
Previously MPDMs used the groups API, which worked fine except when trying to
close them. The closing events were never properly synced with Slack so the
buffers would always reappear after restarting WeeChat. Using the mpim API
(except for info since it doesn't exist: https://api.slack.com/methods#mpim)
seems to fix that.

Poking around the code to understand how things work, let me know if I'm doing anything incorrectly!


trygveaa:
Slack has a new [Conversations API](https://api.slack.com/docs/conversations-api) that they say we should use for anything channel-like (and which is required for shared channels), so I think we should rather switch to that.

If that requires large changes we can probably include this in the meantime though, but we should check it out.

ericdwang:
Yeah, I actually tried that API first, but after encountering errors I went with the mpim API since it just seemed to work. After looking into it more though the conversations join/open and leave/close methods seem fine, but the others have these issues:

- conversations.history doesn't have thread messages
- conversations.mark doesn't exist
- conversations.info doesn't have `unread_count_display`

Based on that I'll update "join" and "leave" in `SLACK_API_TRANSLATOR` to use the new methods.

trygveaa:
Okay, interesting.

I also saw that the conversations API is paginated by default, while the others aren't. We don't currently handle pagination in any way, so we'll probably have to add some support for that eventually.

trygveaa:
Hm, when I try to join a channel with this, the channel is not joined in the web client and if I reload wee-slack the channel disappears.

trygveaa:
The problem is that the join call sends in the channel name instead of the channel id, and `conversations.join` doesn't accept that.

ericdwang:
Great catch, thanks! Updated. Should've noticed that after comparing the methods, my bad.

trygveaa:
Now I've tested this a bit, and it seems to work fine. Thanks!

trygveaa:
I just discovered that this (probably inadvertently) fixes a bug introduced in #413. The bug is that if someone invites you to a private channel, wee-slack will create a public channel with the same name prefixed with `_`.

What happens is that in `SlackChannel.open` we issue a request to Slack if `update_remote` is True. In #413 the prefix of private channels was changed to `&`, so that means that the name we sent in to this join request now had `&` instead of `#`. Apparently, Slack accepts `#`, but translates `&` to `_`. Additionally `channels.join` creates the channel if it doesn't exist.

However, now we send in the identifier instead of the name, so this won't be a problem. Additionally, `conversations.join` won't create the channel if it doesn't exist (well, it can't since you don't give it a name).

If anything depended on wee-slack creating the channel in `SlackChannel.open`, this would be broken now. But I don't think that is the case, I'd guess that request is only there for joining the channel, not creating it.

ericdwang:
Oh interesting, I noticed that bug also while testing with private channels, but at the time it seemed so bizarre and unrelated that I just brushed it off to look at later. The fix is definitely unintentional, but that's good to know!

PR #452, opened 2017-10-10
Add ability to open multiparty DMs with /query and /slack talk

ericdwang:
`command_talk` can now accept multiple arguments, which allows commands
like `/query user1 user2` to open a MPDM with user1 and user2.

### DIFF
--- a/README.md
+++ b/README.md
@@ -123,8 +123,8 @@ Join a channel:
 
 Start a direct chat with someone or multiple users:
 ```
-/query [username] ([username2] [username3]...)
-/slack talk [username] ([username2] [username3]...)
+/query <username>[,<username2>[,<username3>...]]
+/slack talk <username>[,<username2>[,<username3>...]]
 ```
 
 List channels:
--- a/README.md
+++ b/README.md
@@ -121,10 +121,10 @@ Join a channel:
 /join [channel]
 ```
 
-Start a direct chat with someone:
+Start a direct chat with someone or multiple users:
 ```
-/query [username]
-/slack talk [username]
+/query [username] ([username2] [username3]...)
+/slack talk [username] ([username2] [username3]...)
 ```
 
 List channels:
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1048,16 +1048,6 @@ class SlackTeam(object):
     def get_username_map(self):
         return {v.slack_name: k for k, v in self.users.iteritems()}
 
-    def find_channel_by_name(self, name):
-        for channel in self.channels.itervalues():
-            if channel.slack_name == name:
-                return channel
-
-    def find_channel_by_members(self, channel_type, users):
-        for channel in self.channels.itervalues():
-            if channel.type == channel_type and channel.get_members() == users:
-                return channel
-
     def get_team_hash(self):
         return self.team_hash
 
@@ -2175,16 +2165,13 @@ def handle_groupsinfo(group_json, eventrouter, **kwargs):
 
 def handle_conversationsopen(conversation_json, eventrouter, object_name='channel', **kwargs):
     request_metadata = pickle.loads(conversation_json["wee_slack_request_metadata"])
-
     # Set unread count if the channel isn't new (channel_identifier exists)
-    try:
+    if hasattr(request_metadata, 'channel_identifier'):
         channel_id = request_metadata.channel_identifier
         team = eventrouter.teams[request_metadata.team_hash]
         conversation = team.channels[channel_id]
         unread_count_display = conversation_json[object_name]['unread_count_display']
         conversation.set_unread_count_display(unread_count_display)
-    except AttributeError:
-        pass
 
 
 def handle_mpimopen(mpim_json, eventrouter, object_name='group', **kwargs):
@@ -3022,9 +3009,10 @@ def command_talk(data, current_buffer, args):
     if channel_name.startswith('#'):
         channel_name = channel_name[1:]
 
-    chan = team.find_channel_by_name(channel_name)
+    # Try finding the channel by name
+    chan = team.channels.get(team.get_channel_map().get(channel_name))
 
-    # If the channel name doesn't exist, try finding a DM or MPDM instead
+    # If the channel doesn't exist, try finding a DM or MPDM instead
     if not chan:
         # Get the IDs of the users
         u = team.get_username_map()
@@ -3038,21 +3026,22 @@ def command_talk(data, current_buffer, args):
         if users:
             if len(users) > 1:
                 channel_type = 'mpim'
-                # Add the current user if not given since they'll be in MPDMs
-                if team.myidentifier not in users:
-                    users.add(team.myidentifier)
+                # Add the current user since MPDMs include them as a member
+                users.add(team.myidentifier)
             else:
                 channel_type = 'im'
 
-            chan = team.find_channel_by_members(channel_type, users)
+            # Try finding the channel by type and members
+            for channel in team.channels.itervalues():
+                if (channel.type == channel_type and
+                        channel.get_members() == users):
+                    chan = channel
+                    break
 
             # If the DM or MPDM doesn't exist, create it
             if not chan:
                 s = SlackRequest(team.token, SLACK_API_TRANSLATOR[channel_type]['join'], {'users': ','.join(users)}, team_hash=team.team_hash)
                 EVENTROUTER.receive(s)
-                dbg("found user")
-                # Find the channel after creating it
-                chan = team.find_channel_by_members(channel_type, users)
 
     if chan:
         chan.open()
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2981,20 +2981,51 @@ def command_users(data, current_buffer, args):
 @utf8_decode
 def command_talk(data, current_buffer, args):
     """
-    Open a chat with the specified user
-    /slack talk [user]
+    Open a chat with the specified user(s)
+    /slack talk [user] ([user2] [user3]...)
     """
 
     e = EVENTROUTER
     team = e.weechat_controller.buffers[current_buffer].team
-    channel_name = args.split(' ')[1]
+    users = args.split(' ')[1:]
     c = team.get_channel_map()
-    if channel_name not in c:
-        u = team.get_username_map()
+
+    if len(users) > 1:
+        # Add the current user if not specified since all MPDM names have it
+        if team.nick not in users:
+            users.append(team.nick)
+
+        # Use the user order given if no channels are found below
+        channel_name = 'mpdm-{}-1'.format('--'.join(users))
+
+        # Use the name of any existing channel with the same set of users
+        user_set = set(users)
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -121,10 +121,10 @@ Join a channel:
 /join [channel]
 ```
 
-Start a direct chat with someone:
+Start a direct chat with someone or multiple users:
 ```
-/query [username]
-/slack talk [username]
+/query [username] ([username2] [username3]...)
```
trygveaa: The help in weechat for `/query` says:
`[irc]  /query  [-noswitch] [-server <server>] <nick>[,<nick>...] [<text>]`

Usually, I think we should stick to the same syntax for our commands as irc uses, so separate the nicks with comma instead of space. The only thing that's making me unsure is that tab completion doesn't seem to be working when using comma.

README.md
```diff
@@ -121,10 +121,10 @@ Join a channel:
...
+/slack talk [username] ([username2] [username3]...)
```
trygveaa: And a minor thing. `[...]` denotes an optional value, while `<...>` denotes a variable. So this should be:
`/slack talk <username> [<username2> [<username3>...]]`

(I see that this wasn't correct for this command previously either.)

wee_slack.py
```diff
@@ -2938,20 +2938,51 @@ def msg_command_cb(data, current_buffer, args):
...
+
+        # Use the user order given if no channels are found below
+        channel_name = 'mpdm-{}-1'.format('--'.join(users))
+
+        # Use the name of any existing channel with the same set of users
+        user_set = set(users)
+        mpdm_regex = re.compile(r'-+')
+        for channel in c:
+            if channel.startswith('mpdm-') and user_set == set(
+                    mpdm_regex.split(channel)[1:-1]):
```
trygveaa: As far as I can see, the API doesn't give any guarantees about the name of the mpdm channels, and Slack has lately said that we can't use any usernames as ids, we have to use the actual ids. You should rather compare the `user_ids` tuple you create below with the `members` field of the channels. As well as using `channel.type` instead of startswith mpdm.

wee_slack.py
```diff
@@ -2938,20 +2938,51 @@ def msg_command_cb(data, current_buffer, args):
...
+                channel_name = channel
+                break
```
trygveaa: The logic in the above block seem a bit weird when you see it in combination with the if block below. Here you find the channel name based on the channels in c, and then you are checking if the name is in c below. I'd have to think a bit more about it to figure out if there are any cleaner way of writing it, it just seems a bit convoluted.

wee_slack.py
```diff
@@ -2938,20 +2938,51 @@ def msg_command_cb(data, current_buffer, args):
...
+        # Get the IDs of the users
         u = team.get_username_map()
-        if channel_name in u:
-            s = SlackRequest(team.token, "im.open", {"user": u[channel_name]}, team_hash=team.team_hash)
+        user_ids = tuple(u[user] for user in users if user in u)
+
+        # Open the DM or MPDM depending on the number of users
+        if user_ids:
+            if len(user_ids) > 1:
+                method = 'mpim.open'
```
trygveaa: We should probably use the `SLACK_API_TRANSLATOR` mapping here (though I see it wasn't used before either, not sure why).

trygveaa:
Ah, the reason it didn't work is that the `mpdm.open` api expects an argument named `users`, but here it is named `user_ids`. Did this work for you?

ericdwang:
Oh weird, that should be `users`, think that accidentally was changed when I was doing some refactoring. That code only gets run when starting a new group DM and I was testing with existing ones mostly so I didn't catch that (I see that it doesn't work either now), my bad.

README.md
ericdwang: Yeah I was trying to decide between spaces and commas also, and went with spaces because of tab completion haha. I didn't know that weechat's help specified commas though, so I think it makes sense to stick with their syntax instead.

README.md
ericdwang: Sounds good! Yeah, wasn't sure how to format this given the previous conventions.

wee_slack.py
ericdwang: Good call, especially since that PR is switching to use the Conversations API for open methods.

wee_slack.py
ericdwang: Yeah, that second check is definitely redundant for MPDMs. Maybe both the DM and MPDM sections can check `c` (already being done here for MPDMs) and set a variable (e.g. `channel_exists`) if the channel already exists, and then the check below becomes `if channel_exists` or something like that.

wee_slack.py
ericdwang: All great points, thanks! I guess I went with checking the channel name since that's what the DM check was doing and I didn't know about the `members` field.

Should DMs be checking the user ID also then instead of the username? I don't think `SlackDMChannel` is storing the ID currently in `members` but that'd be easy to add.

trygveaa: > Should DMs be checking the user ID also then instead of the username?

Yeah, that would be better.

> I don't think SlackDMChannel is storing the ID currently in members but that'd be easy to add.

No, SlackDMChannel doesn't have a `members` field, but it rather has the id in the `user` field. I don't think we should duplicate the information, even though the id isn't likely to change. We could add a `get_members` helper on SlackChannel which just returns `self.member` and override it in SlackDMChannel to return `[self.user]`, if that's clearer.

ericdwang: Adding `get_members` simplified the logic so I went with that, thanks for the suggestion!

wee_slack.py
ericdwang: So it turns out only mpim.open and conversations.open work for MPDMs, and those aren't in `SLACK_API_TRANSLATOR` currently. Might be better to hold off updating this until #451 is merged in that case, it'd also simplify `params` since conversations.open accepts `"users"` for both DMs and MPDMs.

wee_slack.py
```diff
@@ -1046,6 +1046,16 @@ def get_channel_map(self):
     def get_username_map(self):
         return {v.slack_name: k for k, v in self.users.iteritems()}
 
+    def find_channel_by_name(self, name):
+        for channel in self.channels.itervalues():
+            if channel.slack_name == name:
+                return channel
```
trygveaa: This method can be written like this, instead of looping over the channels:
```
channel_key = self.get_channel_map().get(name)
return self.channels.get(channel_key)
```

wee_slack.py
```diff
@@ -2957,27 +2987,45 @@ def msg_command_cb(data, current_buffer, args):
...
+    if not chan:
+        # Get the IDs of the users
+        u = team.get_username_map()
+        users = {u[user] for user in channel_name.split(',') if user in u}
+
+        if users:
+            if len(users) > 1:
+                channel_type = 'mpim'
+                # Add the current user if not given since they'll be in MPDMs
+                if team.myidentifier not in users:
```
trygveaa: Since users is a set, this if check is not really necessary.

wee_slack.py
```diff
@@ -2957,27 +2987,45 @@ def msg_command_cb(data, current_buffer, args):
...
+            else:
+                channel_type = 'im'
+
+            chan = team.find_channel_by_members(channel_type, users)
+
+            # If the DM or MPDM doesn't exist, create it
+            if not chan:
+                s = SlackRequest(team.token, SLACK_API_TRANSLATOR[channel_type]['join'], {'users': ','.join(users)}, team_hash=team.team_hash)
+                EVENTROUTER.receive(s)
+                dbg("found user")
```
trygveaa: I see that this message was here before as well, but I can't really see when it is useful.

wee_slack.py
```diff
@@ -2957,27 +2987,45 @@ def msg_command_cb(data, current_buffer, args):
...
+                # Find the channel after creating it
+                chan = team.find_channel_by_members(channel_type, users)
```
trygveaa: Since the request created above is asynchronous, so this runs before we get the response from Slack, I don't see when this would return anything other than None. But since this is just the same logic that was here previously, you don't have to change it in this PR.

wee_slack.py
```diff
@@ -2154,12 +2173,23 @@ def handle_groupsinfo(group_json, eventrouter, **kwargs):
...
-    conversation.set_unread_count_display(unread_count_display)
+
+    # Set unread count if the channel isn't new (channel_identifier exists)
+    try:
+        channel_id = request_metadata.channel_identifier
+        team = eventrouter.teams[request_metadata.team_hash]
+        conversation = team.channels[channel_id]
+        unread_count_display = conversation_json[object_name]['unread_count_display']
+        conversation.set_unread_count_display(unread_count_display)
+    except AttributeError:
```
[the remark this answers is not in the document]
ericdwang: It's actually unrelated to `unread_count_display`. The code isn't clear about where the AttributeError could happen, but the comment above this code block mentions this'll only run if `request_metadata.channel_identifier` exists (which new channels lack). It was just another small thing I noticed while testing. I'll change the try/except to `if hasattr(request_metadata, 'channel_identifier')` to make the code explicit.

wee_slack.py
ericdwang: Good point, will remove.

wee_slack.py
ericdwang: Me neither, will also remove.

wee_slack.py
ericdwang: I assumed the requests were async but wasn't sure, so yeah I just kept previous logic. I'll remove this line here though since I know it won't make a difference now.

wee_slack.py
ericdwang: Technically `get_channel_map` loops over the channels also, it just wouldn't be done explicitly here I guess. I just considered this a tiny optimization since it won't always loop over all channels and won't create a dict that'll just get used once and then thrown away. But if readability is a bigger concern I can definitely change it!

ericdwang:
Thanks for reviewing again! Will do, I'll apply the @ prefix check to multiple usernames also.

wee_slack.py
trygveaa: > Technically get_channel_map loops over the channels also, it just wouldn't be done explicitly here I guess.

Yeah, true, but I don't think the performance difference matters in this case. Especially since it's only called after user action.

trygveaa:
Thanks for implementing this, and responding to all my comments!

gianarb:
It looks to me like:

```
/query user1 user2 user3
```

does not work. It does not return anything and it does not create any new buffer. Do I need to do something to enable this feature?

trygveaa:
You have to do `/query user1,user2,user3` as described in `/help query`. It's a bit impractical since tab-completion doesn't work, but that's what's standard in weechat, and it would be hard to separate the message text (which you can write after the nicks) from the nicks if we separate them by space, so that's how it is in wee-slack as well.

gianarb:
Thanks a lot!

Il Ven 17 Gen 2020, 10:55 Trygve Aaberge <notifications@github.com> ha
scritto:

PR #453, opened 2017-10-10
Fix errors when joining DMs and multiparty DMs

ericdwang:
- The DM error was caused by not specifying `return_im`
(https://api.slack.com/methods/im.open#arguments) in the request which
led to `unread_count_display` not being included in the response and a
KeyError in `handle_imopen`
- The MPDM error was caused by using "name" instead of "channel" for the
channel ID (https://api.slack.com/methods/groups.info#arguments) in the
request which led to an error response and a KeyError in
`handle_groupsinfo`

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1286,7 +1286,7 @@ class SlackChannel(object):
 
         if self.type == "im":
             if "join" in SLACK_API_TRANSLATOR[self.type]:
-                s = SlackRequest(self.team.token, SLACK_API_TRANSLATOR[self.type]["join"], {"user": self.user, "return_im": "true"}, team_hash=self.team.team_hash, channel_identifier=self.identifier)
+                s = SlackRequest(self.team.token, SLACK_API_TRANSLATOR[self.type]["join"], {"user": self.user, "return_im": True}, team_hash=self.team.team_hash, channel_identifier=self.identifier)
                 self.eventrouter.receive(s)
 
     def destroy_buffer(self, update_remote):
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -1624,7 +1624,7 @@ def open(self, update_remote=True):
             self.eventrouter.receive(s)
         if update_remote:
             if "join" in SLACK_API_TRANSLATOR[self.type]:
-                s = SlackRequest(self.team.token, SLACK_API_TRANSLATOR[self.type]["join"], {"user": self.user}, team_hash=self.team.team_hash, channel_identifier=self.identifier)
+                s = SlackRequest(self.team.token, SLACK_API_TRANSLATOR[self.type]["join"], {"user": self.user, "return_im": "true"}, team_hash=self.team.team_hash, channel_identifier=self.identifier)
```
trygveaa: The type of "return_im" is specified to be boolean, not string, so this should be `True`.

ericdwang: Good call, I saw `"return_im": "true"` on line 1289 and just did same for consistency here (identical requests), but I'll update both to use `True`.

trygveaa: Ah, I hadn't noticed that line. It seems weird that this is done in both `create_buffer` and `open` though, especially since `create_buffer` is called from `open`, so it will send two requests when you open a DM that doesn't have an existing buffer.

But this is existing weirdness, and your PR is an improvement either way, so I'll merge it.

trygveaa:
Thanks!

PR #457, opened 2017-10-17
Hide activity from muted channels

### LINKED ISSUE #456: Muted channels are not muted in wee-slack
FWIW I can reproduce this fairly easily.  All I need to do is:

1. Join Channel $foo
2. Mute Channel $foo
3. Start wee-slack and ensure Channel $foo is not the currently visible one
4. Have friend talk in $foo as a test
5. See that Channel $foo is marked as having activity

trygveaa:
This sets the correct tags for messages in muted channels and prevents muted channel from being added to the hotlist on start and when a message arrives.

Threads in muted channels are also muted.

Fixes #456


trygveaa:
> Threads in muted channels are also muted.

Is this the correct behavior? I see now that the web client shows activity from threads in muted channels if you have participated in the thread.

I'm not sure how to determine if we should show activity from a thread though. I can't find any properties in the API responses which indicates if Slack would show activity from the thread.

trygveaa:
Now this hides highlights in muted channels as well. We might want to show those.

wee_slack.py
```diff
@@ -1177,11 +1181,12 @@ def rename(self):
     def set_unread_count_display(self, count):
         self.unread_count_display = count
         self.new_messages = bool(self.unread_count_display)
-        for c in range(self.unread_count_display):
-            if self.type == "im":
-                w.buffer_set(self.channel_buffer, "hotlist", "2")
-            else:
-                w.buffer_set(self.channel_buffer, "hotlist", "1")
+        if not self.muted:
```
tfheen: Can we do this as
```
if self.muted:
    return
```
and then leave the code as it is?

tfheen:
> Now this hides highlights in muted channels as well. We might want to show those.

Yeah, a bit undecided on whether this is what "mute" means. I kinda think it's fine, but it also feels like we need levels of muting.

wee_slack.py
trygveaa: Sure, changed it now.

trygveaa:
Hm, I see that the web client shows highlights in the channels list, but does not send notifications.

Personally, I think I would want highlights for my name and configured highlight words, but not for !channel and !here.

I also thought about showing highlights from muted channels as normal activity instead of a highlight, but not sure if that's possible without matching the highlight words ourselves (which I don't think we want to do).

frioux:
So I've been using this all day and it seems like sometimes the read notification (a magenta dashed line on my screen) is missing?

adamgibbins:
I'm not suggesting any alteration to this PR, but, personally this breaks my workflow. I mute stuff I'm not interested in seeing in the client or being pinged with on my phone - but still read it in weechat...eventually. If you're tying the two this destroys things for me, I'm sure I'll find a workaround and I understand why're doing this, but just FYI.

trygveaa:
@frioux: Hm, weird. I haven't had that problem. Is it gone when there are new messages, not just if you change to the channel when there aren't any new messages?

@adamgibbins: Since I already wondered how to handle highlights, and you have yet another workflow, maybe we should add a config option. E.g. `muted_channels_activity` with the options:

- all: Show all activity from muted channels, like other channels.
- highlights: Only show highlights from the channel.
- personal_highlights: Only show personal highlights from the channel, i.e. not @channel and @here.
- none: Don't show any activity from the channel. The messages will still be printed in the buffer.

frioux:
It's not consistent enough for me to have any real guidance yet.  Just
ignore that for now.  It's still better than it was before.

On Thu, Oct 19, 2017 at 06:28:07PM +0000, Trygve Aaberge wrote:
> @frioux: Hm, weird. I haven't had that problem. Is it gone when there are new messages, not just if you change to the channel when there aren't any new messages?
> 
> @adamgibbins: Since I already wondered how to handle highlights, and you have yet another workflow, maybe we should add a config option. E.g. `muted_channels_activity` with the options:
> 
> - all: Show all activity from muted channels, like other channels.
> - highlights: Only show highlights from the channel.
> - personal_highlights: Only show personal highlights from the channel, i.e. not @channel and @here.
> - none: Don't show any activity from the channel. The messages will still be printed in the buffer.
> 
> -- 
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub:
> https://github.com/wee-slack/wee-slack/pull/457#issuecomment-337996226

-- 
fREW Schmidt
https://blog.afoolishmanifesto.com

nwl:
Any update on if/whether this will be merged in?

tfheen:
> Threads in muted channels are also muted.

On further thought, I think that threads are a bit like short-lived channels, so we should treat them independently of their containing channel for muting purposes (at least if we can mute threads too?)

sudoforge:
Is there an update on the acceptance criteria for this? I need to stay in various channels, but _really_ hate that every message highlights the buffer and adds it to the hotlist (_looking at you, `#general`...).

If it's just the merge conflict, I'd be happy to resolve this if the original author is no longer interested.

trygveaa:
@nwl, @sudoforge: Sorry for the long delay! I've now implemented the config options described in my last comment, and resolved the conflicts. I'll test it a bit more, and then probably merge it soon.

> On further thought, I think that threads are a bit like short-lived channels, so we should treat them independently of their containing channel for muting purposes (at least if we can mute threads too?)

You can't mute threads, but Slack has a concept of following threads. However, they don't have any documented APIs for that, and they say the events we get regarding threads are not meant for developer use.

I decided to change it though, and not mute threads at all. You might want to participate and get notifications for threads in muted channels, so it's better if we show activity. And the official client shows threads from muted channels in "New threads", so not muting them matches that better. Maybe we can add an option or something if there is demand for it, but I'm not going to do that in this PR at least.

sudoforge:
@trygveaa RE: Threads, it may be important to note that while unread messages _do_ show up in the "New Threads" feed, they do _not_, to my knowledge, trigger push notifications if the channel is muted. I can do some testing around that later if that would be helpful, as I'm simply attempting to recall from memory here and may well be entirely wrong.

trygveaa:
@sudoforge: Yes, that's true, but we don't have any way to decide if highlights will give you a notification or not.

trygveaa:
> my reference to notifications was simply to draw a relationship between a muted channel and a thread from within that channel. In other words, I think we should also prevent threads from being highlighted or added to the hotlist if the parent channel is muted.

But the hotlist in weechat is closer to the sidebar in Slack than to notifications, and if we don't show them in the hotlist, you will miss the messages completely.

> As you mentioned, weechat notifications and whether or not the buffer is highlighted and added to the hotlist are separate actions. I personally ignore IRC channels, but still receive notifications for specific words/phrases. This is standard behavior, and what I would expect from wee-slack as well -- any chance you're reconsidering the above quoted statement?

What is standard behavior? Are you talking about disabling notifications from IRC buffers in weechat? You should be able to do that the same way for wee-slack buffers too.

trygveaa:
@sudoforge: I've now added a tag, `slack_muted_channel`, to all messages printed in muted channels and threads in muted channels. If you set your push notification plugin to ignore messages with this tag, you will get the same behavior as the official client.

sudoforge:
@trygveaa I think that's a great solution. Thanks for the work on this.

PR #464, opened 2017-10-31
Allow sending special messages by escaping them with / or \

trygveaa:
This allows you to send messages starting with / or messages like s/a/b/
and +:test: as normal text messages instead of performing the action
they normally would. This is done by prefixing the message with / or \,
and then the prefix character is removed from the resulting message.

Allowing / as a prefix, so you can write e.g. //slack is the normal way
to send messages starting with / in weechat. Using \ instead is added
for the other special messages (e.g. /+:test: is interpreted as a
command, so you have to write \+:test: instead). If you want to write a
message starting with \, just start it with \\.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -644,7 +644,7 @@ def buffer_input_callback(signal, buffer_ptr, data):
     if not channel:
         return w.WEECHAT_RC_ERROR
 
-    reaction = re.match("^\s*(\d*)(\+|-):(.*):\s*$", data)
+    reaction = re.match("^(\d*)(\+|-):(.*):\s*$", data)
     substitute = re.match("^(\d*)s/", data)
     if reaction:
         if reaction.group(2) == "+":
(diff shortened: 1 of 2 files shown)


tfheen:
Slack lets you do the same thing by prefixing with space (which is then eaten). Maybe we should too? Otherwise LGTM.

trygveaa:
Ah, yeah, then we should use that. I'll remove `\` unless you object, don't see any point in having three different ways to do it.

tfheen:
No strong feelings on `\`, so sure dropping that sounds fine.

trygveaa:
Fixed.

wee_slack.py
```diff
@@ -664,6 +664,8 @@ def buffer_input_callback(signal, buffer_ptr, data):
             old = old.replace(r'\/', '/')
             channel.edit_nth_previous_message(msgno, old, new, flags)
     else:
+        if data.startswith('//') or data.startswith(' '):
```
auscompgeek: `data.startswith(('//', ' '))` (works in Python 2.5+)

trygveaa: Ah, nice. Fixed.

Issue #467, opened 2017-11-03
colorize_lines.pl does not work with wee-slack


lbolla:
I use `colorize_lines.pl`(https://weechat.org/scripts/source/colorize_lines.pl.html/) to colorize the whole line with the same color as the sender nick.
The plugin does not work with wee-slack, probably because wee-slack is not triggering the event that colorize_lines is listening to? (which is `weechat::hook_modifier("500|weechat_print","colorize_cb", "");`).

trygveaa:
This is caused by the `irc_privmsg` tag missing from the messages. PR #468 fixes that for new messages, but not for backlog messages.

ToxicFrog:
It would be easy enough to modify the PR to apply to backlog messages as well. I don't know how that'll interact with weechat's native hilighting, though.

trygveaa:
The backlog tags include `no_highlight` and `notify_none`, so I would think that prevents any highlighting?

lbolla:
Just tried PR #468 and it seems to work fine for my use case.

PR #468, opened 2017-11-03
Add irc_privmsg to message tags; remove unused "highlightme" tagset

### LINKED ISSUE #283: Unable to use filters reliably
I want to filter bots from certain channels. On IRC I'd do something similar to `/filter add somechannel_github irc.freenode.#somechannel nick_GitHub *` to filter the GitHub nick from #somechannel. However in Slack channels this does not work.

If I toggle `/debug tags` weechat shows the tags for every message, it looks similar to this for an IRC message: `irc_privmsg,notify_message,prefix_nick_85,nick_GitHub,host_~GitHub@somehost,log1`

However on Slack it looks like this:
`notify_message,log1,irc_privmsg` - i.e. all the nick related filters are missing, which I think is why I'm unable to filter so I've no way of distinguishing bots without using some nasty string based regex.

ToxicFrog:
In plain weechat, irc_privmsg is used for all messages received via the
IRC PRIVMSG command, which includes messages to channels, not just PMs.

Not having this tag means some weechat settings related to hilighting
don't behave as expected, and some other scripts like anotify don't
play nice with slack buffers.

Fixes #283.


trygveaa:
Should we add this tag to backlog messages as well? I see that e.g. `colorize_lines.pl` mentioned in #467 need that for it to work properly.

However, messages printed to the team buffer seem to be broken if they contain this tag, so if we do that, we would need to change those from using the backlog tags. I'm not sure why they use the backlog tags in the first place though, I would think it's a bug that they don't trigger the hotlist.

wee_slack.py
```diff
@@ -2838,20 +2838,20 @@ def tag(tagset, user=None):
     else:
         default_tag = 'nick_unknown'
     tagsets = {
+        # messages in the team/server buffer, e.g. "new channel created"
+        "team": "irc_notice,notify_private",
```
trygveaa: I think this should include `log3`.

wee_slack.py
```diff
@@ -2838,20 +2838,20 @@ def tag(tagset, user=None):
...
-        "highlightme": "notify_highlight,log1",
+        "muted": "irc_privmsg,no_highlight,notify_none",
         # when receiving a direct message
-        "dm": "notify_private,notify_message,log1,irc_privmsg",
-        "dmfromme": "notify_none,log1,irc_privmsg",
+        "dm": "irc_privmsg,notify_private,notify_message,log1",
+        "dmfromme": "irc_privmsg,notify_none,log1",
         # when this is a join/leave, attach for smart filter ala:
         # if user in [x.strip() for x in w.prefix("join"), w.prefix("quit")]
         "joinleave": "irc_smart_filter,no_highlight",
```
trygveaa: While we're changing things here, this should include `log4`.

wee_slack.py
```diff
@@ -2838,20 +2838,20 @@ def tag(tagset, user=None):
...
         # when posting messages to a muted channel
-        "muted": "no_highlight,notify_none,logger_backlog_end",
-        # when my nick is in the message
-        "highlightme": "notify_highlight,log1",
+        "muted": "irc_privmsg,no_highlight,notify_none",
         # when receiving a direct message
-        "dm": "notify_private,notify_message,log1,irc_privmsg",
-        "dmfromme": "notify_none,log1,irc_privmsg",
+        "dm": "irc_privmsg,notify_private,notify_message,log1",
+        "dmfromme": "irc_privmsg,notify_none,log1",
```
trygveaa: Hm, why doesn't this include `no_highlight`?

ToxicFrog:
Added the suggested tags.

I'm not 100% sure that backlog should have `no_highlight,notify_none` -- do we want to permit backlog messages to trigger hilights?

auscompgeek:
> do we want to permit backlog messages to trigger hilights?

I think this should happen iff the messages haven't been marked as read yet.

ToxicFrog:
Looking at `buffer_prnt`, it looks like that's what already happens -- messages you've already read get the `backlog` tagset and messages you haven't get `dm`, `dmfromme`, or `default` as appropriate. So I think this is good to go as is.

PR #470, opened 2017-11-08
Display nicklist in MPDM messages

com4:
- Fix a regression from 95fa084893660b655388604084f082a080f1ba32 that disabled the nicklist in multi-person direct
  messages


tfheen:
Thanks!

PR #471, opened 2017-11-08
Add option to disable sending typing notification to Slack

com4:
- Many IM clients that support alerting the receiving user that you're
  typing include an option to disable this behavior. This commit adds
  the `send_typing_notice` boolean option. When it is false
  `typing_notification_cb` is not registered for the
  `input_text_changed` Weechat signal. By default this option is
  true.

https://xkcd.com/1886/


trygveaa:
Thanks!

PR #475, opened 2017-11-13
Hilight team identifier in slack

immae:
This adds a hilight on team identifier in Slack.

The reason is that the api sometimes gives those identifiers without "@" prefix, especially in case of hooks (the bug also appears on slack non-web clients).

Since the identifier is quite "unique" there is almost no risk of unwanted hilights


trygveaa:
Huh, weird that it does. But as you say, there's ~no risk of unwanted highlights, so I don't see any problem adding this.

immae:
Thanks!

PR #476, opened 2017-11-13
Allow query with @'es

immae:
This allows /query to accept @username arguments as if they were username.

Since the completion command prefixes @, even in case of /query, it can help to allow those rather than being obliged to remove it manually


tfheen:
Looks good, thanks!

PR #478, opened 2017-11-20
Hide closed group dms

alyssais:
Group DMs shouldn't be displayed just because `is_member` is `True`, because it always will be. Instead, if a channel/DM/whatever has an `is_open` property, use that, and only fall back to `is_member` when `is_open` isn't present (such as for channels).

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1251,7 +1251,10 @@ class SlackChannel(object):
             self.create_buffer()
             return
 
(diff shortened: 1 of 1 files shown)


trygveaa:
> [...] is_member is True, because it always will be.

For the reference, this only seems to be the case when using `xoxs` tokens. When using `xoxp` tokens, is_member isn't present for mpdm's.

trygveaa:
Looks good, thanks!

scotte:
Thanks for fixing this!

Issue #135, opened 2015-11-26
Use login api call instead of api tokens


rawdigits:
We currently use user-generated tokens for auth. Many teams don't allow their users to generate tokens, so we should support login via either OAUTH or the login endpoint.

rawdigits:
use auth.signin

scotte:
It would be nice if this was an option so both can be supported - for cases (like mine) where I can generate an API token, but not use OAUTH/login (which goes through a custom OAUTH implementation requiring a browser).

rawdigits:
support username/password in the tokens var like so:

'xoxp-asdf9asd89f78asdf89asd7f,user:pass,xoxp-a9sdf78asd9fa7sdf89'

oblitum:
Just had my token revoked in a public team (cpplang) I used to join. I'm not sure whether it's a general trend, now I had to request the token and wait for admin response. No idea whether it will ever come.

oblitum:
It seems like oauth was merged once but then vanished from codebase?

- https://github.com/wee-slack/wee-slack/pull/245

oblitum:
Hmm, I'm reading strong backslash against legacy tokens lately, may be the reason I've got my revoked and I think it's not even wise to ask for more:

- https://twitter.com/myetherwallet/status/900327327971897344
- https://altcointalk.io/topic/128/slack-user-info-accesibility

kasperlewau:
I had a bunch of tokens revoked myself, and we've gone the mile of revoking all legacy tokens in our org for the very reasons specified by @oblitum 

The OAuth code appears to have been lost here https://github.com/wee-slack/wee-slack/commit/95fa084893660b655388604084f082a080f1ba32#diff-67e177ac444a14d6bcd7be6306ab4a20L1133 as part of the `wee-ng-merge` #334 

I haven't yet figured out which particular commit removed the code.

fwiw: I absolutely _love_ wee-slack and my heart sank when I had to go back to the CPU drain that is `ssb-interop`, courtesy of the native Slack client on my Arch box.

selckin:
You can extract a token from the websocket traffic in your browser that will work for a while, until a better solution can be made

matthijskooijman:
What's the status of this? When I run `/slack register`, I get offered an url like `https://slack.com/oauth/authorize?client_id=123.456&scope=client` which suggests that oauth is now supported.

[...]

PR #487, opened 2018-01-15
Restore OAuth support

ajdavis:
Partly addresses #135. This restores some OAuth logic that was lost in a large merge, and it updates the lost logic to be compatible with the latest wee-slack code.

The old logic had simply printed the Slack token to the buffer. I added a new feature: the new logic now automatically appends the token to the "plugins.var.python.slack.slack_api_token" setting and tells the user to restart weechat.

There are two more missing features after this which I propose to add:

1. Fix the OAuth redirect failure. Open a local webserver listening on a random available port, then include the `redirect_uri` parameter to redirect the auth flow back to localhost (if possible, this'll require research and maybe help from you to configure the wee-slack app on slack.com)
2. Update the `config_changed` callback to reload teams if the slack_api_token setting has changed

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,9 +95,9 @@ Log in to Slack:
 /slack register
 ```
 
-This command opens your browser and prompts you to authorize WeeChat with Slack.
-Once you've accomplished this, copy the "code" portion of the URL in the browser
-and pass it to this command:
+This command prints a link you should open in your browser to authorize WeeChat
+with Slack. Once you've accomplished this, copy the "code" portion of the URL in
+the browser and pass it to this command:
 
 ```
 /slack register [YOUR_SLACK_TOKEN]
--- a/README.md
+++ b/README.md
@@ -88,10 +88,35 @@ weechat
 **NOTE:** If weechat is already running, the script can be loaded using ``/python load python/autoload/wee_slack.py``
 
 #### 4. Add your Slack API key(s)
+
+Log in to Slack:
+
+```
+/slack register
+```
+
+This command opens your browser and prompts you to authorize WeeChat with Slack.
+Once you've accomplished this, copy the "code" portion of the URL in the browser
+and pass it to this command:
+
+```
+/slack register [YOUR_SLACK_TOKEN]
+```
+
+Your Slack team is now added, and you can complete setup by restarting the
+wee-slack plugin.
+
+```
+/python reload slack
+```
+
+Alternatively, you can click the "Request token" button at the
+[Slack legacy token page](https://api.slack.com/custom-integrations/legacy-tokens),
+and paste it directly into your settings:
+
 ```
 /set plugins.var.python.slack.slack_api_token [YOUR_SLACK_TOKEN]
 ```
-^^ (find this at https://api.slack.com/custom-integrations/legacy-tokens using the "Request token" button)
 
 If you don't want to store your API token in plaintext you can use the secure features of weechat:
 
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -6,6 +6,7 @@ from collections import OrderedDict
 from functools import wraps
 from itertools import islice
 
+import textwrap
 import time
 import json
 import pickle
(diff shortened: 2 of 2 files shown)


wee_slack.py
```diff
@@ -2964,6 +2964,50 @@ def me_command_cb(data, current_buffer, args):
...
+                    w.config_set_plugin('slack_api_token', d['access_token'])
+                else:
+                    # Add new token to existing set, joined by comma.
+                    tok = config.get_string('slack_api_token')
+                    w.config_set_plugin('slack_api_token',
+                                        ','.join([tok, d['access_token']]))
+                w.prnt(current_buffer,
+                       "Success! Added team \"%s\"" % (d['team_name'],))
+                w.prnt(current_buffer,
+                       "Please restart weechat")
```
tfheen: This should just ask them to reload wee-slack.

ajdavis: Done, thanks.

ajdavis:
New patch, thanks for reviewing.

wee_slack.py
```diff
@@ -2964,6 +2964,50 @@ def me_command_cb(data, current_buffer, args):
...
+1) Paste this into a browser: https://slack.com/oauth/authorize?client_id=2468770254.51917335286&scope=client
+2) Select the team you wish to access from wee-slack in your browser.
+3) Click "Authorize" in the browser **IMPORTANT: the redirect will fail, this is expected**
+4) Copy the "code" portion of the URL to your clipboard
+5) Return to weechat and run `/slack register [code]`
+6) Add the returned token per the normal wee-slack setup instructions
+
+
+"""
+        w.prnt(current_buffer, message)
```
trygveaa: I don't think we should add anything other than the actual messages to channel buffers, so I would want these messages to be printed to the team buffer instead. That goes for all the messages in this command.

wee_slack.py
```diff
@@ -2964,6 +2964,50 @@ def me_command_cb(data, current_buffer, args):
...
+    else:
+        aargs = args.split(None, 2)
+        if len(aargs) != 2:
+            w.prnt(current_buffer, "ERROR: invalid args to register")
+        else:
+            uri = (
+                "https://slack.com/api/oauth.access?"
+                "client_id={}&client_secret={}&code={}"
+            ).format(CLIENT_ID, CLIENT_SECRET, aargs[1])
+            ret = urllib.urlopen(uri).read()
```
trygveaa: Hm, we go through weechat for all other requests (except for the websocket), but maybe it's fine to use urllib for this. It's certainly less hassle to use urllib.

wee_slack.py
```diff
@@ -2964,6 +2964,50 @@ def me_command_cb(data, current_buffer, args):
...
+                    # Add new token to existing set, joined by comma.
+                    tok = config.get_string('slack_api_token')
+                    w.config_set_plugin('slack_api_token',
+                                        ','.join([tok, d['access_token']]))
+                w.prnt(current_buffer,
+                       "Success! Added team \"%s\"" % (d['team_name'],))
+                w.prnt(current_buffer,
+                       "Please reload wee-slack")
+            else:
+                w.prnt(current_buffer, "Failed! Error is: " + d['error'])
```
trygveaa: Even though this is probably printed right after you issue the command, it's always better to give some more context in the error messages. So something like:

```
"ERROR: Couldn't get OAuth token: {}".format(d['error'])
```

wee_slack.py
```diff
@@ -3615,6 +3659,9 @@ def get_string(self, key):
     def get_int(self, key):
         return int(w.config_get_plugin(key))
 
+    def is_default(self, key):
+        return w.config_get_plugin(key) == self.default_settings.get(key)
```
trygveaa: `w.config_get_plugin` returns a string while `default_settings.get` returns a `Settings` object, so this comparison doesn't work.

wee_slack.py
```diff
@@ -2964,6 +2964,50 @@ def me_command_cb(data, current_buffer, args):
...
+                    w.config_set_plugin('slack_api_token', d['access_token'])
+                else:
+                    # Add new token to existing set, joined by comma.
+                    tok = config.get_string('slack_api_token')
+                    w.config_set_plugin('slack_api_token',
+                                        ','.join([tok, d['access_token']]))
+                w.prnt(current_buffer,
+                       "Success! Added team \"%s\"" % (d['team_name'],))
+                w.prnt(current_buffer,
+                       "Please reload wee-slack")
```
trygveaa: I guess most people would know how to reload plugins, but we could always be helpful and tell them what to run:

```
"Please reload wee-slack with: /script reload slack"
```

wee_slack.py
```diff
@@ -2964,6 +2964,50 @@ def me_command_cb(data, current_buffer, args):
     return w.WEECHAT_RC_OK_EAT
 
 
+def command_register(data, current_buffer, args):
+    CLIENT_ID="2468770254.51917335286"
+    CLIENT_SECRET="dcb7fe380a000cba0cca3169a5fe8d70" #this is not really a secret
```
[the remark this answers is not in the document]
ajdavis: Done. (In my defense, I know this isn't PEP8, I'm cherry-picking an old commit by @rawdigits.)

wee_slack.py
```diff
@@ -2964,6 +2964,50 @@ def me_command_cb(data, current_buffer, args):
...
+    if args == 'register':
+        message = """
+#### Retrieving a Slack token via OAUTH ####
```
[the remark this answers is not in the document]
ajdavis: Done.

wee_slack.py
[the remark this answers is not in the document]
ajdavis: Done.

ajdavis: Done.

wee_slack.py
ajdavis: Done.

ajdavis:
I've fixed the ``is_default`` logic, nice catch. I also updated README. The instructions there are temporary, I'm going to make OAuth much simpler in my next patch.

ajdavis:
Hi! I think I've addressed all your comments. If you merge this I can proceed to the next steps, which will make the whole OAuth flow much smoother.

wee_slack.py
```diff
@@ -2964,6 +2965,52 @@ def me_command_cb(data, current_buffer, args):
...
+    if args == 'register':
+        message = textwrap.dedent("""
+            #### Retrieving a Slack token via OAUTH ####
+
+            1) Paste this into a browser: https://slack.com/oauth/authorize?client_id=2468770254.51917335286&scope=client
+            2) Select the team you wish to access from wee-slack in your browser.
+            3) Click "Authorize" in the browser **IMPORTANT: the redirect will fail, this is expected**
+            4) Copy the "code" portion of the URL to your clipboard
+            5) Return to weechat and run `/slack register [code]`
+""")
```
trygveaa: This line can also be indented.

wee_slack.py
```diff
@@ -2964,6 +2965,52 @@ def me_command_cb(data, current_buffer, args):
...
+        team.buffer_prnt(message)
+        return
+
+    aargs = args.split(None, 2)
+    if len(aargs) != 2:
```
trygveaa: Did you see my request for giving more useful names to the parts of args? I'd rather do it like this:

```
_, oauth_code, *extra_args = args.split()
if extra_args:
```

wee_slack.py
```diff
@@ -2964,6 +2965,52 @@ def me_command_cb(data, current_buffer, args):
...
+        team.buffer_prnt("ERROR: invalid args to register")
```
trygveaa: I think this message would be clearer if it says: `ERROR: too many arguments given for register command`

wee_slack.py
```diff
@@ -2964,6 +2965,52 @@ def me_command_cb(data, current_buffer, args):
     return w.WEECHAT_RC_OK_EAT
 
 
+def command_register(data, current_buffer, args):
+    e = EVENTROUTER
+    team = e.weechat_controller.buffers[current_buffer].team
```
trygveaa: My suggestion of printing to the team wasn't that good after all. If you aren't connected to a team already, this will fail.

This command doesn't really have any relevance to the possibly existing teams or where you run it though, so I think we should just always print to the core buffer for this command. That can be done with `w.prnt("", message)`.

wee_slack.py
```diff
@@ -2964,6 +2965,52 @@ def me_command_cb(data, current_buffer, args):
...
+    if config.is_default('slack_api_token'):
+        w.config_set_plugin('slack_api_token', d['access_token'])
+    else:
+        # Add new token to existing set, joined by comma.
+        tok = config.get_string('slack_api_token')
+        w.config_set_plugin('slack_api_token',
+                            ','.join([tok, d['access_token']]))
+
+    team.buffer_prnt("Success! Added team \"%s\"" % (d['team_name'],))
+    team.buffer_prnt("Please reload wee-slack")
```
trygveaa: Can you change this message to: `Please reload wee-slack with: /script reload slack`?

README.md
```diff
@@ -88,10 +88,35 @@ weechat
...
 
 #### 4. Add your Slack API key(s)
+
+Log in to Slack:
+
+```
+/slack register
+```
+
+This command opens your browser and prompts you to authorize WeeChat with Slack.
```
trygveaa: Well, it doesn't open the browser. It just prints a link, so maybe: `This command prints a link you should open in your browser [...]`

ajdavis: Done.

wee_slack.py
ajdavis: Done.

wee_slack.py
ajdavis: Done.

wee_slack.py
ajdavis: Sorry, I think I missed that request. I like your idea but it's Python 3-only so I've modified it.

wee_slack.py
ajdavis: Done.

wee_slack.py
ajdavis: Done.

ajdavis:
New patch, thanks for your review.

oblitum:
@ajdavis thanks for working on this! It's my stumbling block to use Slack on WeeChat. Do you know how long the token stays valid, is the authorization permanent?

wee_slack.py
trygveaa: No problem. And my bad, I thought it worked with python 2.

trygveaa:
@oblitum: The token is valid permanently, unless you revoke it. From https://api.slack.com/docs/oauth
> The token will continue functioning until the installing user either revokes the token and/or uninstalls your application.
> Please note that these access tokens do not expire.

oblitum:
@trygveaa OK, thanks.

trygveaa:
> Fix the OAuth redirect failure. Open a local webserver listening on a random available port

@ajdavis: If you plan on doing this, note that not everyone runs weechat on their local machine (it's common to open it through ssh), so localhost for wee-slack may not be where the user runs their browser.

trygveaa:
Thanks!

oblitum:
Great! but also sad that I didn't realize apps would need to be authorized peer team.... sigh.

Issue #482, opened 2017-12-04
newest wee-slack broken after new user added to org


ryansyah:
Hello,

I've been a daily user of wee-slack now for a little over a year. My configuration broke this last weekend on the most recent RC code in what appears to me as user mixup/identification issue.

Fun side note, my name is Ryan Stark and another Ryan Stark was added to our org over the last couple of weeks.

Here is a debug output sample from a fresh weechat-1.9.1 on FreeBSD-11.1, using weechat-1.0-RC1 (says 1.99 internally):

```
python scripts loaded: slack
Plugins loaded: alias, aspell, buflist, charset, exec, fifo, irc, logger, lua, perl, python, relay, ruby, script, tcl, trigger, xfer
python: stdout/stderr: Traceback (most recent call last):
python: stdout/stderr:   File "/usr/home/syah/.weechat/python/autoload/wee_slack.py", line 502, in handle_next
python: stdout/stderr:     EVENTROUTER.handle_next()
python: stdout/stderr:   File "/usr/home/syah/.weechat/python/autoload/wee_slack.py", line 490, in handle_next
python: stdout/stderr:     self.handlers[function_name](j, self, **kwargs)
python: stdout/stderr:   File "/usr/home/syah/.weechat/python/autoload/wee_slack.py", line 2104, in handle_rtmstart
python: stdout/stderr:     channels[item["id"]] = SlackDMChannel(eventrouter, users, **item)
python: stdout/stderr:   File "/usr/home/syah/.weechat/python/autoload/wee_slack.py", line 1596, in __init__
python: stdout/stderr:     kwargs["name"] = users[dmuser].name
python: stdout/stderr: KeyError: u'U03XXXXXX'
```

My Slack Community member ID is: U0CYYYYYY
The 'new' Ryan Stark Slack Community member ID is: U03XXXXXX

I should also note that when I revert to 0.99.11 code,  the plugin does function although it has it's own errors reported:
```
slack: DEBUG: Loading messages from cache.
slack: DEBUG: Completed loading messages from cache.
python scripts loaded: slack_extension
Plugins loaded: alias, aspell, buflist, charset, exec, fifo, irc, logger, lua, perl, python, relay, ruby, script, tcl, trigger, xfer
python: stdout/stderr: Traceback (most recent call last):
python: stdout/stderr:   File "/usr/home/foo/.weechat/python/autoload/wee_slack.py", line 2316, in url_processor_cb
[...]

PR #462, opened 2017-10-27
Add support for shared channels

ericdwang:
Since shared channels only work with the new conversations API, a new
class of SlackChannel and API type were added to support them. Also
`rtm.start` doesn't include external users so their information needs to
be fetched individually when getting the channel history.

To distinguish external users from team members their nicks get appended
with an asterisk, which can be changed with a new setting. They also
appear in a new "External" group in the nicklist since their presences
can't be fetched with `users.getPresence`.

Unfortunately `conversations.history` doesn't include thread messages so
threads won't show up in shared channels.

Sidenote: the asterisk suffix is also completely arbitrary so I'm open to any suggestions if it doesn't make sense!

PR #492, opened 2018-01-18
Fixes for some trivial errors 

rarbab:
These are fixes for a couple of things that broke after my organization was migrated to Enterprise Grid.


rarbab:
It looks like "Avoid error on DM buffer with unknown user" is the same issue as PR #482.

trygveaa:
With d5af7d8362580eb9ae4564ef52694087d427bbab and the merge of #462, this should not be necessary any more. Please comment if there still is any issues.

rarbab:
Things do seem to be working now. Thanks!

PR #494, opened 2018-01-20
SlackUser: use profile.display_name if available

ToxicFrog:
Since recent changes in Slack, usernames can no longer be set directly, and
are instead inferred from the email address. This is not particularly useful,
so the recommendation is to instead use display_name. display_name may have
spaces in it, however, which confuses tab completion.

With this change, wee-slack uses display name for both display and completion,
after first stripping all whitespace from it. Ref linking (via linkify_text)
ensures that @mentions continue to work properly even when this whitespace
removal means that the name being @mentioned locally doesn't match the "real"
username.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -826,7 +826,7 @@ def nick_completion_cb(data, completion_item, current_buffer, completion):
     for m in current_channel.members:
         u = current_channel.team.users.get(m, None)
         if u:
-            w.hook_completion_list_add(completion, "@" + u.slack_name, 1, w.WEECHAT_LIST_POS_SORT)
+            w.hook_completion_list_add(completion, "@" + u.name, 1, w.WEECHAT_LIST_POS_SORT)
     return w.WEECHAT_RC_OK
 
 
@@ -885,7 +885,7 @@ def complete_next_cb(data, current_buffer, command):
 
     for m in current_channel.members:
         u = current_channel.team.users.get(m, None)
-        if u and u.slack_name == word:
+        if u and u.name == word:
             # Here, we cheat.  Insert a @ in front and rely in the @
             # nicks being in the completion list
             w.buffer_set(current_buffer, "input", line_input[:word_start] + "@" + line_input[word_start:])
@@ -1059,7 +1059,7 @@ class SlackTeam(object):
         return {v.slack_name: k for k, v in self.channels.iteritems()}
 
     def get_username_map(self):
-        return {v.slack_name: k for k, v in self.users.iteritems()}
+        return {v.name: k for k, v in self.users.iteritems()}
 
     def get_team_hash(self):
         return self.team_hash
@@ -1550,7 +1550,7 @@ class SlackChannel(object):
 
         if user and len(self.members) < 1000:
             user = self.team.users[user]
-            nick = w.nicklist_search_nick(self.channel_buffer, "", user.slack_name)
+            nick = w.nicklist_search_nick(self.channel_buffer, "", user.name)
             # since this is a change just remove it regardless of where it is
             w.nicklist_remove_nick(self.channel_buffer, nick)
             # now add it back in to whichever..
@@ -1898,13 +1898,20 @@ class SlackUser(object):
     """
 
     def __init__(self, **kwargs):
-        # We require these two things for a vaid object,
+        # We require these two things for a valid object,
         # the rest we can just learn from slack
         self.identifier = kwargs["id"]
-        self.slack_name = kwargs["name"]
-        self.name = kwargs["name"]
+        self.profile = {}  # in case it's not in kwargs
         for key, value in kwargs.items():
             setattr(self, key, value)
+
+        if self.profile.get("display_name"):
+            self.slack_name = self.profile["display_name"]
+            self.name = self.profile["display_name"].replace(' ', '')
+        else:
+            # No display name set. Fall back to the deprecated username field.
+            self.slack_name = kwargs["name"]
+            self.name = self.slack_name
         self.update_color()
 
     def __repr__(self):


ToxicFrog:
FWIW, I've been running this locally for about two months with no real issues, I just didn't get around to pushing it until now.

The most obvious problem is that, since display names are non-unique, @mentions become potentially ambiguous (and `linkify_text` isn't particularly smart about resolving the ambiguity) -- but since usernames *also* aren't required to be unique anymore, this doesn't actually introduce a new problem, it just changes which names (if any) it manifests for.

wee_slack.py
```diff
@@ -1895,9 +1895,14 @@ def __init__(self, **kwargs):
         # the rest we can just learn from slack
         self.identifier = kwargs["id"]
         self.slack_name = kwargs["name"]
-        self.name = kwargs["name"]
+        self.profile = {}  # in case it's not in kwargs
         for key, value in kwargs.items():
             setattr(self, key, value)
+        if self.profile.get("display_name", ""):
```
trygveaa: You don't need the second argument here.

trygveaa:
Oh, another thing. MPDM names still use the username, they should probably use the display name instead.

trygveaa:
It's easy for the user to change the display name, so we should also implement the [`user_change` event](https://api.slack.com/events/user_change) to change it in wee-slack.

auscompgeek:
> since usernames also aren't required to be unique anymore

I'm pretty sure usernames returned from the Slack API are still guaranteed to be unique. Where did you read that they aren't?

trygveaa:
I think they are unique in one team, but since people from other teams can join in shared channels, you might get two persons with the same username? (unless they change username for non-team members somehow)

ToxicFrog:
> I'm pretty sure usernames returned from the Slack API are still guaranteed to be unique. Where did you read that they aren't?

So, I just reread the Slack API docs on this trying to make sense of it all; here's my understanding. Please correct me if I got any of this wrong:

- **User ID**: the `id` field. This is unique *to a given slack* but not necessarily unique across slacks.  It is the canonical way to refer to a user (e.g. `@mention`s need to use `<@$USER_ID>`) and is not human-readable.
- **Username**: the `name` field. This is *deprecated*. It should not be displayed, is not supported in `@mention`s and should be ignored if the server sends it to you. It used to be set by the user; it's now automatically inferred from the user's email address (by dropping everything after the `@`).
- **Display Name**: the `display_name` and `display_name_normalized` fields. It's set by the user, does not have to be unique even with a slack, and can contain arbitrary UTF8 text, including whitespace. This is what clients are meant to prefer displaying. The `_normalized` version has non-Latin characters stripped; it's unclear what happens if a user's name consists *entirely* of non-Latin chararacters.
- **Real Name**: the `real_name` and `real_name_normalized` fields. Roughly analagous to IRC's "real name" WHOIS field. Otherwise similar to Display Name.

So, as far as collision potential goes:
- Display Name and Real Name have no uniqueness constraints at all
- Username can collide in shared channels; it used to be unique within a slack, but it's unclear if that still holds now that it's set automatically
- ID can also collide in shared channels (you're meant to use other fields like `is_stranger` to differentiate in that case) but cannot collide within a slack.

And as for display and protocol:
- incoming user refs *may* have usernames but *will* have IDs; we should rely solely on the ID when unfurling.
- outgoing user refs *must* have IDs.
- the UI should display (and use for tab completion purposes) display names; it's up to the UI to handle disambiguation when multiple users have the same display name.

ToxicFrog:
And looking over the code again, I'm wondering if the `name` vs `slack_name` distinction needs more work in SlackUser -- right now `name` is display name and `slack_name` is username, but username is deprecated and will presumably go away entirely at some point, which implies all of our code that manipulates users should refer either to `name` (for display to IRC) or `identifier` (for communication with Slack) -- but in that case, is it even worth keeping `slack_name` around? Perhaps it should be removed, or repurposed as "display name with whitespace intact".

wee_slack.py
```diff
@@ -1895,9 +1895,14 @@ def __init__(self, **kwargs):
...
+            # slack_name is derived from the user's email address and is often
+            # not useful. display_name is preferred if set, but may contain
+            # whitespace, which confuses tab completion.
+            self.name = self.profile["display_name"].replace(' ', '')
```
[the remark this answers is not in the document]
ToxicFrog: I hate it, but I would be ok with adding a configuration option for this and don't have strong feelings about what the default is. :)

wee_slack.py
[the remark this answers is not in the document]
ToxicFrog: Done.

ToxicFrog:
Updated with review comments.

> This seems to break prefixing a nick-completion with @ when tabbing twice, but replacing slack_name with name in complete_next_cb didn't help (not that I particulary like that functionality, but it's there).

This is working again now.

wee_slack.py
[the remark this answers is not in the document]
trygveaa: Okay. I don't have a strong opinion, so I think we can wait with a config option until someone requests it.

trygveaa:
@ToxicFrog: After that update, many users (including myself) has a blank nick/empty string.

ToxicFrog:
> After that update, many users (including myself) has a blank nick/empty string.

Whoops. I will investigate.

ToxicFrog:
Ok, this is now worksforme at least.

trygveaa:
Yeah, now it works again.

wee_slack.py
```diff
@@ -1898,13 +1898,23 @@ class SlackUser(object):
...
         for key, value in kwargs.items():
             setattr(self, key, value)
+
+        if self.profile.get("display_name"):
+            self.slack_name = self.profile["display_name"]
+            # TODO: configuration options here for 'what character to replace with'
+            # and 'whether to use normalized display name or real one'.
+            self.name = self.profile["display_name"].replace(' ', '')
+        else:
+            # No display name set. This should never happen, but if it does,
```
trygveaa: That this should never happens isn't correct. [The docs](https://api.slack.com/methods/users.info) say:
> The profile hash contains as much information as the user has supplied in the default profile fields: first_name, last_name, real_name, display_name, skype, and the image_* fields. Only the image_* fields are guaranteed to be included. Data that has not been supplied may not be present at all, may be null or may contain the empty string ("").

On my testing team, multiple users has `display_name` set to the empty string. (I see that this is the same cause as my previous comment, which is now fixed, but you should change this comment as well).

wee_slack.py
```diff
@@ -1898,13 +1898,23 @@ class SlackUser(object):
...
-        self.slack_name = kwargs["name"]
-        self.name = kwargs["name"]
+        self.profile = {}  # in case it's not in kwargs
         for key, value in kwargs.items():
             setattr(self, key, value)
+
+        if self.profile.get("display_name"):
+            self.slack_name = self.profile["display_name"]
+            # TODO: configuration options here for 'what character to replace with'
+            # and 'whether to use normalized display name or real one'.
```
trygveaa: Is there any reason for including an option to use the normalized display name? I mean, it's often good to include configuration options, but too many options will just be impractical, both for the user and the code.

wee_slack.py
ToxicFrog: Yeah, forgot to update the comment along with the code. Done now.

wee_slack.py
ToxicFrog: I have a few regulars on one of my slacks who have names that start with decorative non-latin characters, which makes tab-completing them difficult; I at least would use such an option. I'll take out the TODO, though -- if other people want it someone will file an issue.

trygveaa: Okay, didn't think about tab completion. Fair enough.

PR #495, opened 2018-01-20
Suppress SSLWantReadError errors

ToxicFrog:
This exception is expected to some degree and does not affect the functioning
of wee-slack. With this change, it will no longer appear in the log.


karlp:
how often is "to some degree" though?  I get it once a minute, every minute, of every day.  And yes, it doesn't seem to have any real affect, other than log spew, but is that how often you expect it?

PR #496, opened 2018-01-27
Fix presence subscription now mandatory to get state changes

immae:
https://api.slack.com/docs/presence-and-status#batching

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -195,7 +195,6 @@ def get_nick_color_name(nick):
 ##### BEGIN NEW
 
 IGNORED_EVENTS = [
-    "hello",
     # "pref_change",
     # "reconnect_url",
 ]
@@ -1108,7 +1107,7 @@ class SlackTeam(object):
                 # The fast reconnect failed, so start over-ish
                 for chan in self.channels:
                     self.channels[chan].got_history = False
-                s = SlackRequest(self.token, 'rtm.start', {}, retries=999)
+                s = initiate_connection(self.token, retries=999)
                 self.eventrouter.receive(s)
                 self.connecting = False
                 # del self.eventrouter.teams[self.get_team_hash()]
@@ -1151,6 +1150,11 @@ class SlackTeam(object):
             if user.id in c.members:
                 c.update_nicklist(user.id)
 
+    def subscribe_users_presence(self):
+        self.send_to_websocket({
+            "type": "presence_sub",
+            "ids": self.users.keys(),
+            }, expect_reply=False)
 
 class SlackChannel(object):
     """
@@ -2234,6 +2238,8 @@ def handle_history(message_json, eventrouter, **kwargs):
 
 ###### New/converted process_ and subprocess_ methods
 
+def process_hello(message_json, eventrouter, **kwargs):
+    kwargs['team'].subscribe_users_presence()
 
 def process_reconnect_url(message_json, eventrouter, **kwargs):
     kwargs['team'].set_reconnect_url(message_json['url'])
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -1108,7 +1108,7 @@ def connect(self):
                 # The fast reconnect failed, so start over-ish
                 for chan in self.channels:
                     self.channels[chan].got_history = False
-                s = SlackRequest(self.token, 'rtm.start', {}, retries=999)
+                s = SlackRequest(self.token, 'rtm.start', {"batch_presence_aware": True}, retries=999)
```
trygveaa: `urllib.urlencode` will turn this into `batch_presence_aware=True`, not `batch_presence_aware=true`, so you won't get batched presences.

Though, I tried sending `"true"` instead and still doesn't get batched presences (i.e. `presence_change` has a `user` field instead of a `users` field). When sending `1` instead, I get them, so that seems like a bug in the docs or API. We should send `1` so we actually get the batched events.

I'll report a bug to slack on this.

wee_slack.py
```diff
@@ -2248,6 +2255,11 @@ def process_presence_change(message_json, eventrouter, **kwargs):
         user = kwargs["user"]
```
trygveaa: The docs state that we can remove processing of the `user` field:
> Sometimes you'll get a single event for a single user but if you use batch_presence_aware=true, that single user event will be single item in the users array.

wee_slack.py
```diff
@@ -2248,6 +2255,11 @@ def process_presence_change(message_json, eventrouter, **kwargs):
...
         team = kwargs["team"]
         team.update_member_presence(user, message_json["presence"])
+    if "users" in message_json:
+        team = kwargs["team"]
+        for user_id in message_json["users"]:
+            user = self.teams[team].users[user_id]
```
trygveaa: `self` isn't defined here. You should just use `team`.

wee_slack.py
```diff
@@ -3751,7 +3763,7 @@ def trace_calls(frame, event, arg):
 
             tokens = config.slack_api_token.split(',')
             for t in tokens:
-                s = SlackRequest(t, 'rtm.start', {})
+                s = SlackRequest(t, 'rtm.start', {"batch_presence_aware": True })
```
trygveaa: I'd like a method for connecting with `rtm.start` so we don't have to duplicate the options.

wee_slack.py
```diff
@@ -1100,6 +1099,7 @@ def connect(self):
                     # self.attach_websocket(ws)
                     self.set_connected()
                     self.connecting = False
+                    self.subscribe_users_presence()
```
trygveaa: This can be removed.

wee_slack.py
```diff
@@ -1151,6 +1151,11 @@ def update_member_presence(self, user, presence):
             if user.id in c.members:
                 c.update_nicklist(user.id)
 
+    def subscribe_users_presence(self):
+        self.send_to_websocket({
+            "type": "presence_sub",
+            "ids": self.users.keys(),
+            }, expect_reply=True)
```
trygveaa: Do we expect a reply?

wee_slack.py

wee_slack.py

wee_slack.py
immae: Hmm no, I probably put it to true to make some tests and forgot to revert

wee_slack.py
immae: Should we keep it until it is more stable, or do you want it removed now?

wee_slack.py

wee_slack.py
trygveaa: Hm, maybe it's best to wait, though I suspect we won't remember to remove it. But we can wait until I get a reply on the bug I submitted at least.

immae: Do you have a reference / url for your bug report? I’ll add a TODO comment in the code

trygveaa: No, I didn' get any. I don't think it's public, they'll just (hopefully) email me.

immae: Ok then just a TODO comment :p

ToxicFrog:
This appears to have broken my larger (~4k users) slack -- I get disconnected shortly after connecting, and after bouncing back and forth a few times it starts failing to connect at all with the message "ratelimited". 

At a guess, it's hitting this:

> The message server will disconnect any client that sends a message longer than 16 kilobytes. This includes all parts of the message, including JSON syntax, not just the message text.

4k users in one subscription message is going to blow past that easily.

PR #499, opened 2018-01-29
Temporary fix user presence subscription crashing big teams

immae:
Big teams will make big json and go over the limitation of 4k per websocket message.


wee_slack.py
```diff
@@ -1152,9 +1152,12 @@ def update_member_presence(self, user, presence):
                 c.update_nicklist(user.id)
 
     def subscribe_users_presence(self):
+        # FIXME: We should try to be smarter to fetch the users whom we
```
trygveaa: Can you write that this is done because of a size limitation in the API?

PR #501, opened 2018-01-30
Don't tag team messages as normal messages

trygveaa:
When we tag messages to the team buffer as normal messages, they get
highlights like normal messages. The messages we send there are
informational messages that we don't want to trigger highlights.


trygveaa:
I added the `no_highlight` tag to team messages as well, and a custom slack tag for each type. See the commit messages for details.

PR #502, opened 2018-02-02
Don't add deleted users to the nicklist

trygveaa:
This check was done when generating complete nicklist, but not when
updating a single user. Deleted users are still part of the member list
of a channel, if they don't leave the channel first. This caused a
deleted user to be inserted into the nicklist when processing a joined
channel message or when receiving a presence change for that user.

When investigating this, I saw that we still receive presence change
events for deleted users if we subscribe to them. We might want to
filter out the deleted users when generating the list of users to
subscribe to.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1557,6 +1557,8 @@ class SlackChannel(object):
 
         if user and len(self.members) < 1000:
             user = self.team.users[user]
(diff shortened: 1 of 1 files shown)


khronosschoty:
Thanks for the fix!

PR #503, opened 2018-02-11
Complete OAuth flow automatically

ajdavis:
Followup to #487

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,7 +89,15 @@ weechat
 
 #### 4. Add your Slack API key(s)
 
-Log in to Slack:
+If weechat is running on your local machine, you can log in to Slack easily:
+
+```
+/slack register local
+```
+
+This command opens your browser, and completes the authentication process
+automatically when you click "Authorize" in the browser. If you are using
+weechat over ssh, use:
 
 ```
 /slack register
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -8,18 +8,22 @@ from itertools import islice
 
 import textwrap
 import time
+import threading
 import json
 import pickle
 import sha
 import os
 import re
 import urllib
+import urlparse
 import sys
 import traceback
 import collections
 import ssl
 import random
 import string
+import webbrowser
+from BaseHTTPServer import BaseHTTPRequestHandler, HTTPServer
 
 from websocket import create_connection, WebSocketConnectionClosedException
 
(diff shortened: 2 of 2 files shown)


ajdavis:
Instead of redirecting to a nonexistent domain, I propose redirecting to httpbin, where it's easier to see the OAuth code. Additionally, if weechat is running on localhost, we can complete the OAuth flow entirely automatically.

This requires that both "http://httpbin.org" and "http://localhost:54782" (a port number I chose arbitrarily) are added to the OAuth redirect URIs in the Slack configuration for wee-slack:

screen shot 2018-02-11 at 12 47 27 pm

auscompgeek:
I think sending OAuth tokens over the internet, in plaintext, is a bad idea.

ajdavis:
Good point. Now using HTTPS for displaying the OAuth code on httpbin.

jgoerzen:
I'm not sure sending these tokens to an arbitrary third-party site is really a good plan either.  I'd rather it either go to localhost or keep the current behavior.

ajdavis:
I sort of agree. But with the current behavior, which redirects to "notadomain" or something, that can have the same effect. When I complete OAuth flow on my apartment's WiFi, my ISP, which is Verizon, directs me to their own search page, something like:

https://searchassist.verizon.com/main?FailedURI=https%3A%2F%2Fnotadomain%2Fcode=...

Some people will have a similar experience, where they're redirected to an arbitrary site, and this site is controlled by their local gateway's DNS configuration. Consider what might happen at a hotel or Internet café. Furthermore, it's likely to be harder to read the code out of that URL than it is from httpbin.

However, if you feel strongly that we should keep this behavior instead of switching to httpbin, I'm happy to undo that part of my patch.

auscompgeek:
I think redirecting to localhost would be sensible to avoid DNS hijacking.

tigertoes:
@auscompgeek using localhost will not work when the instance of weechat differs (i.e. when my desktop browser is signed in to Slack, but my weechat instance is running on a server somewhere else).

auscompgeek:
@tigertoes I think you missed the entire point that was being expressed in the last few comments. Leaking user tokens to third parties, _especially to DNS hijacking servers and/or over plain text_, is objectively bad. Redirecting to localhost instead of a (currently) nonexistent domain is a strict improvement to the current scenario, regardless of where weechat is running.

tigertoes:
@auscompgeek a service on localhost **may** however be responding to that endpoint, leading to undesirable behaviour and end user foot-shooting. The concerns about `not.a.realhost` being registered at some time in the future could be mitigated by using a `.invalid` address instead, which under [RFC6761§6.4](https://tools.ietf.org/html/rfc6761#section-6.4) and [RFC2606§2](https://tools.ietf.org/html/rfc2606#section-2) cannot be reserved for future (mis)use lest the RFCs are made obselete.

tfheen:
Since there's a requirement for the port to be specified up front, we can't just find a free port (else we could bind to port 0 and let the OS allocate one for us). However, HTTPServer will be unhappy if there's something else already listening on the port, so we can catch that and give the user a sensible error message.

In addition to that, I agree with not sending tokens to httpbin makes sense, but we should have the nonexistent name end with .invalid to prevent it from being registered. wee-slack-registration.invalid or something along those lines?

trygveaa:
This makes all of weechat hang while the server is running, which is especially a problem both because it can make connections time out and because if the user never visits the callback page (which will happen e.g. if the slack team only allows approved apps and wee-slack is not approved) it will hang forever. I've also seen weechat crash when using threading in scripts (though, not experienced it in this PR), so I'm not sure that's safe.

Additionally, I don't really want lots of code in the script for setting up an http server etc. Especially since it will only work for users running weechat locally, and the users will have to figure out if they should use `/slack register` or `/slack register local`.

Instead I have added a page with GitHub Pages and set the redirect_uri to this page. This means that you won't have automatic completion like in this PR, but you will get to a page which has an input field with the command you need to run in weechat, so you just have to copy that input field. That should be simple and easy enough to understand.

I included `#` at the end of the redirect_uri. This means that the OAuth code will only be read by the browser, and not sent to the server. So even though the page is hosted at GitHub Pages, they will not receive the OAuth codes.

trygveaa:
> I included # at the end of the redirect_uri. This means that the OAuth code will only be read by the browser, and not sent to the server. So even though the page is hosted at GitHub Pages, they will not receive the OAuth codes.

So, Slack doesn't allow `redirect_uri` to contain `#` anymore. It's technically not allowed by the OAuth specification, so I can't blame them, but it worked fine before now. This means that we have to find another solution.

Are you guys really that worried about the code being leaked to a third party? The code is only valid until it is used, and for a maximum of 10 minutes, so I don't think leaking it to GitHub Pages is a big deal. Thoughts?

aereaux:
One option would be to use a helper script like weechat-matrix does. See https://github.com/poljar/weechat-matrix/blob/69ad2a9c03d516c212d3d0700dbb2bfe654f6365/matrix/server.py#L821 and https://github.com/poljar/weechat-matrix/blob/950d912c8289bc79304bf3faac8520f007d1707d/main.py#L377

trygveaa:
That would only work if weechat is run on the local machine though. I think a lot of people (myself included) use it on a remote server over ssh.

PR #386, opened 2017-07-08
fixed private multiline +editor to compose a multiline message 

dzfweb:
This PR include:
PR #https://github.com/wee-slack/wee-slack/pull/365 
Issue #https://github.com/wee-slack/wee-slack/issues/385

PR #506, opened 2018-02-17
colorise block `code` in slack

yeled:
This is the code from #386

@dzfweb please shout out if you want to re-make this PR yourself. I think it's a good thing and don't want it to get lost. I couldn't find it in your [fork](https://github.com/dzfweb/wee-slack
)
It looks like this: 
image


dzfweb:
I've lost my fork with this code wrongly. It is a great feature which i use everyday and I think other people would like to use too.

auscompgeek:
@yeled Maybe you want to set the author of the commit to dzfweb.

yeled:
I've the time:motivation ratio to do this properly. The code as it stands will strip useful code depending on $reasons I can't remember.

@dzfweb I am going to close this - please submit a new PR if you think you have a newer version free from bugs or with a more generic regex enging

Issue #247, opened 2016-08-10
Ability to see real names


nertwork:
I'm wondering if there's a way to add in something like /whois username?

I haven't seen anything setup to do that yet and it's a feature that I need as we don't have a standard for usernames.

tfheen:
There currently isn't, no. This is a duplicate of #100, so I'll close this. It should be fairly easy to implement, so if you want to put together a PR, I would be happy to review it.

Issue #100, opened 2015-08-05
Implement nickname > real name translation


moritzheiber:
wee-slack right now uses the regular nicknames assigned to each entity within Slack. However, most of my colleagues are using obscure handles I don't necessarily associate them with. It would be amazing to have wee-slack "translate" these nicknames into whatever first and last name people have entered on the website.

Thanks!

tfheen:
I started on this, and it's fairly easy to do, apart from you running into unicode comparison weirdness in a bunch of places, and it feels super odd, so it certainly has to be behind a setting.

rawdigits:
+1 on making it an option and not the default. I have tried this in the normal client and don't recognize anyone by their full name. :)

LovesTha:
Would be nice for this to be a per slack server setting. As the slacks I'm in that have a professional bend are very much about real names, the more social/hobby oriented ones are alias based.

LovesTha:
Any chance your attempt to make this change was in a branch that could be shared. This is a language I'm not particularly familiar with so attempting my own patch could do with a kickstart.

trygveaa:
@LovesTha: I added a global option now, but I agree that having this as a per server setting would be nice. I opened issue #710 for that.

PR #507, opened 2018-02-28
Add whois command

Shaac:
The display name is not always very explicit. This PR adds a whois command getting the real name from the display name.

This implements #247 and could be a response to #100 

I made weechat print the result on the current buffer, which seems more convenient, but to be more coherent with weechat IRC behaviour, it could be printed on the team buffer.


auscompgeek:
Can we also show other parts of the profile as well? Email (if available), status, …

tfheen:
I reimplemented this based on your work (and added a reference to your work).  Thanks for the contribution! :-)

Issue #516, opened 2018-03-09
rtm.start times out due to >24mb of response data


cleverca22:
```
        elif return_code != -1:
            w.prnt("", "fail with {}".format(return_code))
            self.reply_buffer.pop(request_metadata.response_id, None)
            self.delete_context(data)
        else:
            if request_metadata.response_id not in self.reply_buffer:
                self.reply_buffer[request_metadata.response_id] = ""
            self.reply_buffer[request_metadata.response_id] += out
            w.prnt("", "first 128 chars {}".format(out[:128]));
            w.prnt("", "downloaded {}mb bytes...".format(len(self.reply_buffer[request_metadata.response_id])/1024/1024))
```

```
first 128 chars "deleted":false,"color":"e475df","real_name":"KOH TIONG SIM","tz":"Asia\/Chongqing","tz_label":"China Standard Time","tz_offset"
downloaded 24mb bytes...
first 128 chars Dongliang Jin","tz":"Asia\/Chongqing","tz_label":"China Standard Time","tz_offset":28800,"profile":{"title":"","phone":"","skype
downloaded 24mb bytes...
first 128 chars t":"","status_emoji":"","status_expiration":0,"avatar_hash":"gf6a1c8ce016","first_name":"tochi","last_name":"eke-okoro","image_2
downloaded 24mb bytes...
first 128 chars hash":"gb05e7790bc2","first_name":"Tom","last_name":"Bot","image_24":"https:\/\/secure.gravatar.com\/avatar\/5a92944743d38eadd87
downloaded 24mb bytes...
fail with -2
```

i am unable to connect to a slack because of the sheer number (16059) of users within it
i suspect weechat is having a timeout on the curl call to handle rtm.start, and weechat is not rendering any error at all, it just silently fails to connect without the above patches

aha, and `/set plugins.var.python.slack.slack_timeout  "200000"` is enough to make it connect

the `rtm.start` reply is 27mb in total, and the download gets slower as it progresses due to the string append getting more costly as it grows
storing the parts of the http response in an array, and then joining them at the end would save cpu time overall and also make it download faster

tfheen:
Can you try https://err.no/p/2018-03-09-ADu1m3CRblk/stdin and see if that gives you acceptable performance with the default slack timeout?

cleverca22:
```
15:45:04           weechat     | python scripts loaded: slack
15:45:12 XXXX.slack.com     | Connected to Slack
```
[...]

PR #520, opened 2018-03-09
Use StringIO instead of string concatenation

tfheen:
This makes a big difference for larger teams where rtm.start can time
out just due to Python's strings being immutable.

Fixes: #516


rawdigits:
This seems totally fine to me. 👍

Issue #428, opened 2017-08-23
weechat.look.nick_suffix/prefix no longer works


synic:
From this ticket, the feature was added: https://github.com/wee-slack/wee-slack/issues/140

However, it seems to have gone away.  Is there a replacement?  Was it removed on purpose?

auscompgeek:
This probably wasn't removed deliberately, but rather likely disappeared with the wee-ng refactor. PRs welcome, probably.

PR #522, opened 2018-03-09
Re-add support for weechat.look.nick_{prefix,suffix}

tfheen:
Fixes: #428


rawdigits:
I don't use this, but the code looks like it will return what is wanted here. 👍

PR #525, opened 2018-03-13
Add explicit_changes setting causing modifications to trigger explicit updates

jgoerzen:
Background:

I am using weechat through a relay (weechat.el using the weechat
protocol).  I also keep logs using the standard weechat mechanism.

Neither of these methods capture a message being edited, a thread
being created, etc.  They are just completely silent on it.

It would be nice to have an option that would disable features that
edit history, and instead add new lines to the buffer ("Thread a8f
created for message 3", "Message 4 edited to: xxxx", etc).  I wouldn't
mind seeing reactions this way either, though that could be annoying
to some.

This accomplishes that.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1418,6 +1418,7 @@ class SlackChannel(object):
 
     def change_message(self, ts, text=None, suffix=None):
         ts = SlackTS(ts)
+        m = None
         if ts in self.messages:
             m = self.messages[ts]
             if text:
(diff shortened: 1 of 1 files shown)


trygveaa:
Sorry for not answering to this PR. This would probably be pretty noisy, but I see the issue with not seeing changes in the relay client.

As far as I know you will get the updated history with changes if you reconnect the relay client, though that might be impractical.

Is this still something people are interested in?

aereaux:
Is there some chance that this will be fixed in the weechat relay API and in clients?

trygveaa:
@aereaux: I don't think anyone here can answer that. I tried searching in the weechat issues, but didn't find any feature request for this. I would guess the chance of this being fixed is small unless someone who wants the feature looks into what it would take to fix it.

PR #528, opened 2018-03-14
Fix self-nick color in messages

### LINKED ISSUE #527: use chat_nick_self color for sender's own messages
In #318 the sender's own messages are coloured based on the auto-chosen colour by weechat. Unfortunately, this doesn't respect the `weechat.color.chat_nick_self` setting that should override the auto-chosen colour.

This is a simple fix, and I hope to have a PR submitted shortly.

wohali:
The work in #318 correctly gets most users' nick coloring, but this information was previously discarded for the user's own messages.

This fix correctly uses the sender's own `weechat.color.chat_nick_self` preference for self-sent messages.

Fixes #527

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2009,12 +2009,11 @@ class SlackMessage(object):
         name_plain = ""
         if 'user' in self.message_json:
             if self.message_json['user'] == self.team.myidentifier:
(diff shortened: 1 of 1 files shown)


tfheen:
Thanks, looks good to me. :+1:

PR #529, opened 2018-03-14
Add support to disable teammate link previews

### LINKED ISSUE #526: Fully disable URL previews / unfurling
Despite comment #381 wee-slack still displays the full URL previews, even after disabling URL preview in the web client (no effect).

Example:
```
19:10 <wohali> I've added the wee-slack application for our Slack space. This
               lets me connect to Slack from weechat, a text-mode client that I
               prefer. Open source, if you have any concerns, look here:
               https://github.com/wee-slack/wee-slack
19:10 wee-slack/wee-slack
19:10 wee-slack - A WeeChat plugin for http://Slack.com (Slack.com).
      Synchronizes read markers, provides typing notification, search, etc..

```

Ideally, disabling URL previews could be blacklisted/whitelisted, or possibly limited in the number of characters returned. (Example: an in-lined Twitter status is less disruptive than an inlined generic webpage description.

I have a code modification for this I'm working on, and hope to submit a PR soon.

wohali:
This PR introduces a new config option, `plugins.var.python.slack.link_previews`, set to `true` by default.
* With the default `true` setting, wee-slack works as it does today.
* Disabling this setting (`false`, `off`, etc.) will disable link previews of URLs pasted by teammates. It does not affect previews or disable displaying additional attachment content from bots, services, server notices, etc. The setting will affect new incoming messages, or will be retroactively applied if you `/script reload slack`, `/python reload`, or restart weechat.

Fixes #526.

This isn't the whitelist/blacklist functionality that the Slack web UI has, but it's a start! The values of the setting could be changed to a comma-delimited list of domains that would be whitelisted, or blacklisted if prefixed by a `!`, perhaps.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2745,6 +2745,8 @@ def unhtmlescape(text):
 
 
 def unwrap_attachments(message_json, text_before):
+    if 'user' in message_json and not config.link_previews:
+        return ''
     text_before_unescaped = unhtmlescape(text_before)
     attachment_texts = []
     a = message_json.get("attachments", None)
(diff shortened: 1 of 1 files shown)


wohali:
FYI I have just discovered that this isn't a complete solution to link embedding from teammates. I'm going to close this PR and will open a new one once I've got a more complete solution. Consider this a WIP :)

PR #544, opened 2018-03-27
Instructions Incorrect on Fresh installation

andrewbolster:
the `python/autoload` directory structure does not exist in the default fresh install of weechat


auscompgeek:
Doesn't it? I have a ~/.weechat/guile/autoload, and I sure as hell have not installed any guile scripts ever.

mikeofmany:
Answering my own question.  Recent release broke the scripts into individual packages.  So a straight weechat install isn't going to work.. on homebrew the command for all the script packages is: "brew install weechat --with-aspell --with-curl --with-python --with-perl --with-ruby --with-lua --with-guile"

tfheen:
This should be resolved now that we tell people to install weechat with Python, so I'm going to close this.

PR #549, opened 2018-04-03
Fix channels not being marked as read when latest message is in a thread

ericdwang:
`SlackChannel.mark_read` was only looking at non-thread messages when
determining the latest timestamp to use, so thread messages past that
weren't included. They were then still counted in `unread_count_display`
from the `.info` API methods, and channels with them would always appear
unread on startup.


auscompgeek:
Nice. Looks like this would also give a performance boost as well.

ericdwang:
Working on a better way to handle this and resolve some other thread-related issues.

Issue #552, opened 2018-04-09
Presence status - Active not working 


paulozanco:
Should set the presence.set status to _auto_ instead of _active_.
Based on the slack documentation, there is no way to set the status to active directly.

https://api.slack.com/docs/presence-and-status

**Manual Away**
An application can call users.setPresence to manually mark a user as _away_ or _auto_. A manual status set using this method will persist between connections.

A manual away status set using this method overrides the automatic presence determined by the message server.

Setting presence back to auto indicates that the automatic status should be used instead. There's no way to force a user status to _active_.

PR #554, opened 2018-04-09
resolved issue #552

paulozanco:
Should set the presence.set status to _auto_ instead of _active_.
Based on the slack documentation, there is no way to set the status to active directly.

https://api.slack.com/docs/presence-and-status

**Manual Away**
An application can call users.setPresence to manually mark a user as _away_ or _auto_. A manual status set using this method will persist between connections.

A manual away status set using this method overrides the automatic presence determined by the message server.

Setting presence back to auto indicates that the automatic status should be used instead. There's no way to force a user status to _active_.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -3434,7 +3434,7 @@ def command_back(data, current_buffer, args):
     /slack back
     """
     team = EVENTROUTER.weechat_controller.buffers[current_buffer].team
-    s = SlackRequest(team.token, "presence.set", {"presence": "active"}, team_hash=team.team_hash)
+    s = SlackRequest(team.token, "presence.set", {"presence": "auto"}, team_hash=team.team_hash)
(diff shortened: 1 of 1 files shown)


trygveaa:
Hm, so we currently use `presence.set`, and that documentation is for `users.setPresence`. Apparently, they have both existed for a long time (https://twitter.com/slackapi/status/547186709222268928), but is now acting differently. The current code is working for me, while this PR is not. Are you saying that's different for you?

Either way, we should change to `users.setPresence`, since that's the only one documented, and with that, using auto as you're doing is the correct way. So could you change that as well?

Also, could you write a more descriptive commit message? The current one doesn't tell you anything without looking at the issue tracker, which is impractical when viewing the log.

paulozanco:
Hi, I created another Pull Request based on your recommendations and I closed this one.
Thanks for sharing your moment.

PR #555, opened 2018-04-10
Modernize Python 2 code to get ready for Python 3

cclauss:
Some minimal, "safe" changes to make the code syntax compatible with both Python 2 and Python 3.
* Run futurize --stage1 http://python-future.org/futurize_cheatsheet.html
* Define __basestring__ and __unicode__ in Python 3.
* Modify Travis CI to start running Python 3 tests in __allow_failures__ mode.

Ready for review. Python 3 syntax compatibility does not mean that the port is complete.

### DIFF
--- a/_pytest/test_utf8_helpers.py
+++ b/_pytest/test_utf8_helpers.py
@@ -23,19 +23,23 @@ def test_decode_preserves_iterable_type():
     assert type(value_tuple) == type(decode_from_utf8(value_tuple))
 
 def test_decodes_utf8_string_to_unicode():
-    assert u'æøå' == decode_from_utf8(b'æøå')
+    assert u'æøå' == decode_from_utf8(b'æøå')  # noqa
 
 def test_decodes_utf8_dict_to_unicode():
-    assert {u'æ': u'å', u'ø': u'å'} == decode_from_utf8({b'æ': b'å', b'ø': b'å'})
+    assert {u'æ': u'å', u'ø': u'å'} == decode_from_utf8({b'æ': b'å', b'ø': b'å'})  # noqa
 
 def test_decodes_utf8_list_to_unicode():
-    assert [u'æ', u'ø', u'å'] == decode_from_utf8([b'æ', b'ø', b'å'])
+    assert [u'æ', u'ø', u'å'] == decode_from_utf8([b'æ', b'ø', b'å'])  # noqa
 
 def test_encode_preserves_string_without_utf8():
     assert b'test' == encode_to_utf8(u'test')
 
 def test_encode_preserves_byte_strings():
-    assert b'æøå' == encode_to_utf8(b'æøå')
+    try:
+        unicode        # Python 2
+        assert b'æøå' == encode_to_utf8(b'æøå')  # noqa
+    except NameError:  # Python 3
+        pass
 
 def test_encode_preserves_mapping_type():
     value_dict = {'a': 'x', 'b': 'y', 'c': 'z'}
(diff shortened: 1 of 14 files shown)


_pytest/test_utf8_helpers.py
```diff
@@ -35,7 +35,11 @@ def test_encode_preserves_string_without_utf8():
     assert b'test' == encode_to_utf8(u'test')
 
 def test_encode_preserves_byte_strings():
-    assert b'æøå' == encode_to_utf8(b'æøå')
+    try:
+        unicode        # Python 2
+        assert b'æøå' == encode_to_utf8(b'æøå')
```
trygveaa: This causes a syntax error in Python 3 ("bytes can only contain ASCII literal characters"), so we need a different fix than this.

wee_slack.py
```diff
@@ -30,9 +31,16 @@
...
     import weechat
-except:
+except ImportError:
     pass
 
+try:
+    basestring     # Python 2
+    unicode
+except NameError:  # Python 3
+    basestring = (str, )
```
trygveaa: Why a tuple of `str` instead of just `str`?

wee_slack.py
```diff
@@ -30,9 +31,15 @@
...
     import weechat
-except:
+except ImportError:
     pass
 
+try:
+    basestring     # Python 2
+    unicode
+except NameError:  # Python 3
+    basestring = unicode = str
```
trygveaa: I didn't say using a tuple was necessarily wrong, but I wondered the reason for using it. :) Or wasn't there any specific reason?

_pytest/test_utf8_helpers.py
```diff
@@ -23,19 +23,23 @@ def test_decode_preserves_iterable_type():
     assert type(value_tuple) == type(decode_from_utf8(value_tuple))
 
 def test_decodes_utf8_string_to_unicode():
-    assert u'æøå' == decode_from_utf8(b'æøå')
+    assert u'æøå' == decode_from_utf8(b'æøå')  # noqa
```
trygveaa: `b'æøå'` is a syntax error, so `# noqa` won't help as that is only a indication to linters. I'm not sure myself at the moment how to deal with this, will have to think about it.

cclauss: We are covered.  The try / except block keeps that line from ever executing on Python 3 and the noqa keeps the linters from complaining.

wee_slack.py
cclauss: I originally thought there was a reason for making it a tuple of one but further test proved that I was wrong.

trygveaa: Ah, okay :)

_pytest/test_utf8_helpers.py
cclauss: What does [__six.b('æøå')__](http://six.readthedocs.io/#binary-and-text-data) do?

trygveaa: Are you asking me? I've never used six :)

But it documents that it encodes in latin-1 on python 3, which we don't want to use. It made me realize that `b'...'` is equivalent to `'...'` in python 2 though, so we can just remove the b's to get rid of the syntax errors. However, that may not give the correct behavior in python 3, will have to check this.

wee_slack.py
```diff
@@ -1,6 +1,7 @@
 # -*- coding: utf-8 -*-
 
 from __future__ import unicode_literals
+from __future__ import print_function
```
auscompgeek: This should really be a single future statement.

_pytest/test_utf8_helpers.py
auscompgeek: I guess these tests never made any sense in the first place. Go figure.

Something like `decode_from_utf8('æøå'.encode())` might make sense though. But remind me why we have a `decode_from_utf8()` in the first place.

trygveaa: @auscompgeek: I do think the tests make sense, even though the `b` in front of the strings was redundant. But they just make sense in python 2, which they were written for, not in python 3 which handles unicode differently.

We have `decode_from_utf8` to recursively convert structures from bytestrings to unicode. All the strings we receive from weechat are bytestrings, but unicode strings are much more convenient and safer to work with. So we need to convert everything going from/to weechat.

trygveaa:
Sorry for not following up on this sooner. I've now created a PR which should have complete support for Python 3, #686.

PR #556, opened 2018-04-10
Use flake8 to find undefined names

cclauss:
__flake8__ tests currently flag __server__ as an undefined name twice in __wee_slack.py__.  Undefined names can raise a __NameError__ at runtime.  This PR adds flake8 testing to our Travis test runs.  ~~This PR will continue to fail until the two undefined names below are resolved.~~

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./wee_slack.py:3316:13: F821 undefined name 'server'
            server.buffer_prnt("Usage: /slack slash /someslashcommand [arguments...].")
            ^
./wee_slack.py:3417:13: F821 undefined name 'server'
            server.buffer_prnt("Usage: /slack status [status emoji] [status text].")
            ^
2     F821 undefined name 'server'
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,10 +1,19 @@
 language: python
 python:
   - "2.7"
-install:
-  - 'pip install websocket-client pytest'
+cache: pip
 
-script: python -m pytest
+install:
+  - pip install flake8 pytest websocket-client
+
+before_script:
+  # stop the build if there are Python syntax errors or undefined names
+  - flake8 .
+  # exit-zero treats all errors as warnings.
+  - flake8 . --exit-zero --select=C,E,F,W
+
+script:
+  - python -m pytest
 
 notifications:
   irc: "chat.freenode.net#wee-slack-dev"
--- a/.flake8
+++ b/.flake8
@@ -1,0 +1,7 @@
+[flake8]
+select = E901,E999,F821,F822,F823
+count = True
+max-complexity = 10
+max-line-length = 120
+show-source = True
+statistics = True


trygveaa:
Thanks! I've fixed those two errors now.

Can you add a configuration file for flake8 as well, so we can run it from the command line without specifying any arguments (at least for the error/not exit-zero case)?

.travis.yml
```diff
@@ -1,10 +1,19 @@
...
+
 install:
-  - 'pip install websocket-client pytest'
+  - pip install flake8 pytest websocket-client
+
+before_script:
+  # stop the build if there are Python syntax errors or undefined names
+  - flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics
+  # exit-zero treats all errors as warnings.  The GitHub editor is 127 chars wide
+  - flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
```
trygveaa: The github editor size seems like a pretty arbitrary limit. Maybe go for 120 instead, which is a bit more common? (I'd go for 100 or 80, but that gives a bit too many warnings at the moment, so let's decrease it further at a later time.)

.travis.yml
```diff
@@ -1,10 +1,19 @@
...
+
 install:
-  - 'pip install websocket-client pytest'
+  - pip install flake8 pytest websocket-client
+
+before_script:
+  # stop the build if there are Python syntax errors or undefined names
+  - flake8 .
+  # exit-zero treats all errors as warnings.
+  - flake8 . --count --exit-zero --max-complexity=10 --max-line-length=120 select=E,F,W --statistics
```
trygveaa: You need to select `C90` as well to include the complexity check, and you're missing `--`, so: `--select=C90,E,F,W`.

You can specify `max-complexity` and `max-line-length` in the config file.

You can drop `--count` and `--statistics` since those are specified in the config.

trygveaa:
Thanks!

PR #558, opened 2018-04-11
Add show_user_presence config option

xdg:
This commit adds a `show_user_presence` config option that defaults to
true.  When true, the `+` appears before present users in the buffer
list, as they did before this commit.  But when false, the `+` mark is
omitted and presence is no longer shown.

This helps users who want their terminal programs to flag activity but
don't consider presence changes to be activity worth flagging.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1658,10 +1658,10 @@ class SlackDMChannel(SlackChannel):
             print_color = self.color
         else:
             print_color = ""
-        if not present:
-            prepend = " "
-        else:
+        if present and config.show_user_presence:
             prepend = "+"
+        else:
+            prepend = " "
         select = {
             "default": self.slack_name,
             "sidebar": prepend + self.slack_name,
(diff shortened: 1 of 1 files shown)


auscompgeek:
I think it would make more sense to not have the space there at all if this is turned off.

xdg:
I'm wasn't sure if the space is necessary for formatting reasons so I didn't want to change unrelated formatting behavior.  This PR just allows one to make all users look "not present".

auscompgeek:
IIUC, the space is only there to avoid the short name (used in buflist) jumping when other users with open DMs go online/offline. The short name shouldn't be used anywhere other than the short buffer name.

auscompgeek:
Anyway, I think this looks fairly good either way. The setting name is misleading though - presence is also shown in other places (namely the nicklist).

xdg:
@auscompgeek Any suggestions for a better name?

Issue #566, opened 2018-04-24
Backticks don't disable formatting inside a message


eras:
Not sure if it's possible to fix this with wee-slack or rather with weechat itself ;), but it seems backtick quoting doesn't appropriately disable ie. bold formatting inside the region. There is bug #354 which apparently can be fixed with wee-slack so I dared to file this as well :).

Ie. `\`hello \*bar\*\`` will result in `bar` being highlighted when the expected result is that the asterisks are visible and no formatting is performed.

ToxicFrog:
Yeah, like #354 this one is my fault and fixing it will require a client side formatter that's smarter than the existing search-and-replace implementation.

Helpfully, the Slack API docs don't seem to include detailed semantics or a reference implementation for their markdown processor, although they do talk about ways to turn it off, so exactly replicating the behaviour of slack is probably a nonstarter, but it can definitely be less bad than it is.

auscompgeek:
I've filed #567 as some quick fixes until we get a real fix.

PR #567, opened 2018-04-24
Minor formatting fixes

auscompgeek:
First and foremost, this prevents formatting going across newlines. Consider this example message:
```markdown
* an item
* another item
```

Clearly, `an item` should not be made bold. Slack doesn't.

This also preserves the original typed message, with `*_formatting_*` characters galore, for a few reasons:
* as a quick fix for #566 until we fix that properly, and
* Slack preserves formatting characters when copy-pasting.

This also prevents formatting going across backticks, such as in the following example:
```markdown
* Run this command: `find . -name '*.exe'`
```

### DIFF
--- a/_pytest/test_formatting.py
+++ b/_pytest/test_formatting.py
@@ -12,3 +12,16 @@ import wee_slack
 ])
 def test_does_not_format(text):
     assert wee_slack.render_formatting(text) == text
+
+
+@pytest.mark.parametrize("text", [
+    "`hello *bar*`",
+    "`*`",
+    "`* *`",
+    "`* * *`",
+    "`* * * *`",
+    "`* * * * *`",
+    "`* * * * * *`",
+])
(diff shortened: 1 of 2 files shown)


tfheen:
Would it be possible to add some tests for this so we can prevent breakage in the future?

auscompgeek:
A test for newlines would probably make sense, and probably also the formatting across backticks. However, I could imagine that people would prefer the character-preserving behaviour to be configurable once #566 and #354 are fixed properly.

trygveaa:
> Avoid formatting matching into code spans/blocks

This change actually breaks some formatting, since bold text can include \`. E.g. ```*bold ` text*``` is shown in bold in the web client, and is currently shown in bold in wee-slack, but is not bold after this change.

But it's probably better to prevent bolding of your example than to preserve this bolding, so I'll merge this.

Thanks!

PR #568, opened 2018-04-26
Fix nick sanitization

crawford:
This now properly replaces spaces with underscores.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2938,8 +2938,7 @@ def format_nick(nick):
 
 def tag(tagset, user=None):
     if user:
-        user.replace(" ", "_")
-        default_tag = "nick_" + user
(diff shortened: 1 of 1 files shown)


tfheen:
Thanks, nice catch. :-)

wee_slack.py
```diff
@@ -2938,8 +2938,7 @@ def format_nick(nick):
 
 def tag(tagset, user=None):
     if user:
-        user.replace(" ", "_")
-        default_tag = "nick_" + user
+        default_tag = "nick_" + user.replace(" ", "_")
```
trygveaa: Hm, I can't see any place the user would contain spaces that needs to be replaced here. Did you experience any problems before this fix?

crawford: Yes. The RSS bot posts messages with spaces in the nick.

PR #569, opened 2018-04-29
insert newline after triple backslashes if not present

dbarrosop:
It's quite often that users post code in slack like this:

    ```if blah:
        return True
    ```

This simple patch should make this more readable by inserting a newline after the triple backslashes if not present already:

    ```
    if blah:
        return True
    ```


wee_slack.py
```diff
@@ -1407,6 +1407,7 @@ def destroy_buffer(self, update_remote):
             self.eventrouter.receive(s)
 
     def buffer_prnt(self, nick, text, timestamp=str(time.time()), tagset=None, tag_nick=None, **kwargs):
+        text = re.sub(r"```(?!\n)", "```\n", text)
```
trygveaa: When the message changes, the message will be rendered again, which will overwrite this. To make this persist through message edits, you need to place this in the function called `render` instead.

This adds a newline at the end of the message if the message ends with ```. This should not happen.

dbarrosop:
Will look at it in a few days, thanks!

wee_slack.py
dbarrosop: I think both comments are now addressed

trygveaa:
Now, it adds a blank line before the opening \`\`\`. It also removes the first character after \`\`\` if there is no newline between them.

PR #578, opened 2018-05-14
fix: corrected registration instructions

grindhold:
in the registration dialog, wee_slack asks the user to reload the
slack script via /script. in my weechat 2.1 on debian sid, this is
not the proper way to reload scripts anymore. instead one has to
use the /python command to achieve the same effect. this commit
corrects the misleading instruction message.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -3142,7 +3142,7 @@ def command_register(data, current_buffer, args):
                             ','.join([tok, d['access_token']]))
 
     w.prnt("", "Success! Added team \"%s\"" % (d['team_name'],))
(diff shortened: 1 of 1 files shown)


trygveaa:
Hm, strange that the weechat docs still says you can use `/script`. https://weechat.org/files/doc/stable/weechat_scripting.en.html#load_script

grindhold:
maybe the debian people did something downstream that alters the behaviour, although i doubt it. i have no time to investigate this, sorry.

trygveaa:
No worries, I was just curious :) I guess it might have to do with the package splitting they did for debian.

We use `/python` in the readme, so it's good to be consistent either way.

PR #581, opened 2018-05-15
readme: add hint about split-packaging in distros

grindhold:
some distributions do not distribute weechat with support for all
plugins by default, but use separate packages to provide support
for different scripting languages. users should be made aware of that


auscompgeek:
I don't understand why you opened another PR for this.

grindhold:
Because i wanted to make sure you only have to add one commit. I don't exactly know wheter github makes it possible to add PRs as a single squash commit. And i did not want to make a forcepush. I hate forcepushes.

auscompgeek:
GitHub has a squash merge button.

Issue #573, opened 2018-05-09
can't register when behind a proxy


Coolgeek789:
I can't register when behind a proxy. error : 
```
 python : stdout/stderr (slack) : Traceback (most recent call last):
 python : stdout/stderr (slack) :   File "/home/frendu/.weechat/python/autoload/wee_slack.py", line 119, in wrapper
 python : stdout/stderr (slack) :     return f(*decode_from_utf8(args), **decode_from_utf8(kwargs))
 python : stdout/stderr (slack) :   File "/home/frendu/.weechat/python/autoload/wee_slack.py", line 3282, in slack_command_cb
 python : stdout/stderr (slack) :     EVENTROUTER.cmds[function_name]("", current_buffer, args)
 python : stdout/stderr (slack) :   File "/home/frendu/.weechat/python/autoload/wee_slack.py", line 3097, in command_register
 python : stdout/stderr (slack) :     d = json.loads(ret)
 python : stdout/stderr (slack) :   File "/usr/lib/python2.7/json/__init__.py", line 339, in loads
 python : stdout/stderr (slack) :     return _default_decoder.decode(s)
 python : stdout/stderr (slack) :   File "/usr/lib/python2.7/json/decoder.py", line 364, in decode
 python : stdout/stderr (slack) :     obj, end = self.raw_decode(s, idx=_w(s, 0).end())
 python : stdout/stderr (slack) :   File "/usr/lib/python2.7/json/decoder.py", line 382, in raw_decode
 python : stdout/stderr (slack) :     raise ValueError("No JSON object could be decoded")
 python : stdout/stderr (slack) : ValueError: No JSON object could be decoded
```

After tracking down, I changed the line "ret = urllib.urlopen(uri).read()" by "ret = urllib2.urlopen(uri).read()" in function command_register.

It works now. It use http_proxy/https_proxy env var to lookup the proxy address. Maybe it should use the proxy system given by weechat.

yardenac:
Thanks. Maybe this should be a pull request

Coolgeek789:
I'm no python dev, but I can try to make something. I'm not sure how much time it will take though.
I'm sure you can make something much faster than me. And much cleaner too

trygveaa:
We should probably use `w.hook_process_hashtable` to request the url instead of urllib, like we do in `local_process_async_slack_api_request`. This will make weechat perform the request, so it should use the proxy settings automatically.

PR #582, opened 2018-05-16
add http proxy support

Coolgeek789:
See issue #573
The proxy can be configured within weechat using /proxy command and weechat.network.proxy_curl settings.
Upload has not be tested because I didn't have anythin to upload. But it should work

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -225,7 +225,7 @@ class ProxyWrapper(object):
             if self.proxy_port:
                 port = ":{}".format(self.proxy_port)
             else:
-                port= ""
+                port = ""
                 
             return "--proxy {}{}{}".format(user, self.proxy_address, port)
 
@@ -1143,7 +1143,8 @@ class SlackTeam(object):
             self.connecting = True
             if self.ws_url:
                 try:
-                    # only http proxy is currenlty supported
+                    # only http proxy is currently supported
+                    proxy = ProxyWrapper()
                     if proxy.has_proxy == True:
                         ws = create_connection(self.ws_url, sslopt=sslopt_ca_certs, http_proxy_host=proxy.proxy_address, http_proxy_port=proxy.proxy_port, http_proxy_auth=(proxy.proxy_user, proxy.proxy_password))
                     else:
@@ -3475,8 +3476,9 @@ def command_upload(data, current_buffer, args):
         file_path = file_path.replace(' ', '\ ')
 
     # only http proxy is currenlty supported
+    proxy = ProxyWrapper()
     proxy_string = proxy.curl()
-    command = 'curl -F file=@{} -F channels={} -F token={} {} {}'.format(file_path, channel.identifier, team.token, proxy, url)
+    command = 'curl -F file=@{} -F channels={} -F token={} {} {}'.format(file_path, channel.identifier, team.token, proxy_string, url)
     w.hook_process(command, config.slack_timeout, '', '')
 
 @utf8_decode
@@ -3938,9 +3940,6 @@ if __name__ == "__main__":
 
     if w.register(SCRIPT_NAME, SCRIPT_AUTHOR, SCRIPT_VERSION, SCRIPT_LICENSE,
                   SCRIPT_DESC, "script_unloaded", ""):
-
-        global proxy
-        proxy = ProxyWrapper()
 
         weechat_version = w.info_get("version_number", "") or 0
         if int(weechat_version) < 0x1030000:
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -3164,12 +3164,12 @@ def command_register_callback(data, command, return_code, out, err):
                 w.config_set_plugin('slack_api_token',
                                     ','.join([tok, d['access_token']]))
 
-                w.prnt("", "Success! Added team \"%s\"" % (d['team_name'],))
-                w.prnt("", "Please reload wee-slack with: /script reload slack")
-                return w.WEECHAT_RC_OK_EAT
-           w.prnt("", "ERROR: problem when trying to get Slack OAuth token. Got 0 length answer. Err: ".format(err))
-           w.prnt("", "Check the network or proxy settings")
-           return w.WEECHAT_RC_OK_EAT
+            w.prnt("", "Success! Added team \"%s\"" % (d['team_name'],))
+            w.prnt("", "Please reload wee-slack with: /script reload slack")
+            return w.WEECHAT_RC_OK_EAT
+        w.prnt("", "ERROR: problem when trying to get Slack OAuth token. Got 0 length answer. Err: ".format(err))
+        w.prnt("", "Check the network or proxy settings")
+        return w.WEECHAT_RC_OK_EAT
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -1103,7 +1103,21 @@ def connect(self):
...
+                        if proxy_type == "http":
+                            proxy_address = w.config_string(weechat.config_get("{}.address".format(proxy_string)))
+                            proxy_port = w.config_integer(weechat.config_get("{}.port".format(proxy_string)))
+                            proxy_user = w.config_string(weechat.config_get("{}.username".format(proxy_string)))
+                            proxy_password = w.config_string(weechat.config_get("{}.password".format(proxy_string)))
+
+                    if not proxy_name:
+                        ws = create_connection(self.ws_url, sslopt=sslopt_ca_certs)
+                    else:
+                        ws = create_connection(self.ws_url, sslopt=sslopt_ca_certs, http_proxy_host=proxy_address, http_proxy_port=proxy_port, http_proxy_auth=(proxy_user, proxy_password))
```
tfheen: Can you merge those two if statements?

wee_slack.py
```diff
@@ -3126,24 +3140,38 @@ def command_register(data, current_buffer, args):
...
-    ret = urllib.urlopen(uri).read()
-    d = json.loads(ret)
-    if not d["ok"]:
-        w.prnt("",
-            "ERROR: Couldn't get Slack OAuth token: {}".format(d['error']))
-        return
+    params = {'useragent': 'wee_slack {}'.format(SCRIPT_VERSION)}
+    w.hook_process_hashtable('url:', params, config.slack_timeout, "", "")
+    w.hook_process_hashtable("url:{}".format(uri), params, config.slack_timeout, "command_register_callback", "")
+    return
```
tfheen: Unneeded return.

wee_slack.py
```diff
@@ -3126,24 +3140,38 @@ def command_register(data, current_buffer, args):
...
-    else:
-        # Add new token to existing set, joined by comma.
-        tok = config.get_string('slack_api_token')
-        w.config_set_plugin('slack_api_token',
-                            ','.join([tok, d['access_token']]))
+@utf8_decode
+def command_register_callback(data, command, return_code, out, err):
+    if return_code == 0:
+        if len(out) > 0:
+            d = json.loads(out)
```
tfheen: To make this more readable and prevent deep nesting, it's better to write this as

`if return code != 0: ` and then do the error handling + return. It also means matching up the condition and the code that runs is easier.

wee_slack.py
```diff
@@ -3404,10 +3432,28 @@ def command_upload(data, current_buffer, args):
...
+    proxy_name = proxy = w.config_string(weechat.config_get('weechat.network.proxy_curl'))
+    if not proxy_name:
+        proxy_string = "weechat.proxy.{}".format(proxy_name)
+        proxy_type = w.config_integer(weechat.config_get("{}.type".format(proxy_string)))
+        if proxy_type == "http":
+            proxy_address = w.config_string(weechat.config_get("{}.address".format(proxy_string)))
+            proxy_port = w.config_string(weechat.config_get("{}.port".format(proxy_string)))
+            proxy_user = w.config_string(weechat.config_get("{}.username".format(proxy_string)))
+            proxy_password = w.config_string(weechat.config_get("{}.password".format(proxy_string)))
+            proxy = "--proxy "
```
tfheen: These functions are called a few times, can you factor it into a function instead that just returns the settings?

wee_slack.py
```diff
@@ -3404,10 +3432,28 @@ def command_upload(data, current_buffer, args):
...
+
+            if proxy_user and proxy_password:
+                proxy = "{}{}:{}@".format(proxy,proxy_user,proxy_password)
+                proxy = "{}{}".format(proxy,proxy_address)
+                if proxy_port:
+                    proxy = "{}:{}".format(proxy,proxy_port)
+
+    command = 'curl -F file=@{} -F channels={} -F token={} {} {}'.format(file_path, channel.identifier, team.token, proxy, url)
```
tfheen: I think this should be rewritten to use the built-in curl support in weechat. Are you up for that? (If not, that's ok and we can do it in a later iteration.)

wee_slack.py
```diff
@@ -3404,10 +3432,28 @@ def command_upload(data, current_buffer, args):
...
+            proxy_port = w.config_string(weechat.config_get("{}.port".format(proxy_string)))
+            proxy_user = w.config_string(weechat.config_get("{}.username".format(proxy_string)))
+            proxy_password = w.config_string(weechat.config_get("{}.password".format(proxy_string)))
+            proxy = "--proxy "
+
+            if proxy_user and proxy_password:
+                proxy = "{}{}:{}@".format(proxy,proxy_user,proxy_password)
+                proxy = "{}{}".format(proxy,proxy_address)
+                if proxy_port:
+                    proxy = "{}:{}".format(proxy,proxy_port)
```
tfheen: This could use a little bit of style cleanup: missing space after comma, and reusing the string like you're doing does not for the most readable code.

wee_slack.py
```diff
@@ -3404,10 +3432,28 @@ def command_upload(data, current_buffer, args):
...
+
+    command = 'curl -F file=@{} -F channels={} -F token={} {} {}'.format(file_path, channel.identifier, team.token, proxy, url)
     w.hook_process(command, config.slack_timeout, '', '')
 
-
```
tfheen: This blank line should be here, according to PEP8. :-)

Coolgeek789:
will do my best but need more guiding here. I'm not a python dev. it's my first time doing this kind of things :)

"Can you merge those two if statements?" => can you tell me how to merge those two ?

wee_slack.py
Coolgeek789: It seems that weechat doesn't support uploading files. maybe in a later version of it.

wee_slack.py
```diff
@@ -3877,6 +3939,9 @@ def initiate_connection(token, retries=3):
     if w.register(SCRIPT_NAME, SCRIPT_AUTHOR, SCRIPT_VERSION, SCRIPT_LICENSE,
                   SCRIPT_DESC, "script_unloaded", ""):
 
+        global proxy
+        proxy = ProxyWrapper()
```
tfheen: It's probably better to just call this the places you use it so you pick up any changes in weechat's settings.

wee_slack.py
```diff
@@ -190,6 +190,46 @@ def prnt_date_tags(self, buffer, date, tags, message):
...
+            
+            if self.proxy_user and self.proxy_password:
+                user = "{}:{}@".format(self.proxy_user, self.proxy_password)
+            else:
+                user = ""
+                    
+            if self.proxy_port:
+                port = ":{}".format(self.proxy_port)
+            else:
+                port= ""
```
tfheen: Super minor change: Add a missing space before `=`

wee_slack.py
```diff
@@ -1103,7 +1143,12 @@ def connect(self):
             self.connecting = True
             if self.ws_url:
                 try:
-                    ws = create_connection(self.ws_url, sslopt=sslopt_ca_certs)
+                    # only http proxy is currenlty supported
```
tfheen: Minor typo in currently.

PR #595, opened 2018-06-22
Remove insecure instructions of storing Slack API tokens in weechat

mbbx6spp:
Changes to README.md that remove instructions promoting plaintext storage of Slack API tokens in user's weechat configurations.

### DIFF
--- a/README.md
+++ b/README.md
@@ -113,18 +113,13 @@ wee-slack plugin.
 ```
 
 Alternatively, you can click the "Request token" button at the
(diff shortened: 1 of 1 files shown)


trygveaa:
The `/slack register` command stores the token as plaintext. We might want that to use the secure features before telling our users to do? I guess users might assume it stores the tokens securely if that's the way we tell users to do it.

trygveaa:
I'm closing this because of the reason mentioned in my last comment.

If someone wants to make a PR adding support for storing the token securely in `/slack register`, that would be welcome.

PR #596, opened 2018-07-13
Adds a decorator `slack_buffer_force`

gdestuynder:
It forces the function to operate on a Slack buffer. This is used for
`command_{away,back}` in particular which can be called by Weechat's
`/away [-]all` command and expects to be honored by ALL buffers.

This means it also fixes `auto_away.py` and `screen_away.py` scripts for
example

Note: i don't know if anyone had a different fix in mind, but this annoyed me quite a bit (not being set away correctly) - couldn't find an issue about it somehow (i'm probably blind)

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -91,6 +91,18 @@ SLACK_API_TRANSLATOR = {
 
 ###### Decorators have to be up here
 
+def slack_buffer_force(f):
+    """
+    Force to run in a Slack buffer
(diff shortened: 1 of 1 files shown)


gdestuynder:
sorry never got around to adding this :) (can keep it closed)

trygveaa:
@gdestuynder: I implemented support for `-all` in 00093fa. Haven't tested if it works with auto_away.py and screen_away.py though.

PR #608, opened 2018-08-19
Colorize edits and reactions

immae:
This colorizes a bit "metatext" like "edited", "deleted" and reactions, to distinguish them visually from the real text.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2592,7 +2592,9 @@ def subprocess_message_changed(message_json, eventrouter, channel, team):
     channel.change_message(new_message["ts"], message_json=new_message)
 
 def subprocess_message_deleted(message_json, eventrouter, channel, team):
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -2990,9 +2999,14 @@ def create_user_status_string(profile):
 def create_reaction_string(reactions):
     count = 0
     if not isinstance(reactions, list):
-        reaction_string = " [{}]".format(reactions)
+        reaction_string = " {}[{}{}]{}".format(
+                w.color("245"),
+                reactions,
+                w.color("245"),
```
trygveaa: Why do you repeat the color? Can the reaction string include colors?

immae: I do change the emoji colors, but you’re right it’s not relevant in this commit, I removed the useless mentions

trygveaa:
Thanks!

PR #616, opened 2018-08-23
Add option to show thread messages in parent channel

### LINKED ISSUE #546: Option to De-thread Threads
Would it be possible to have an option setting to disable threads entirely? I'm not really feeling opening up a new channel everytime I want to jump into a thread, I'd prefer they just log as regular chat messages, maybe with an indicator that they belong to a thread, instead of being in their own channel.

janza:
Closes #546

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -208,7 +208,7 @@ class ProxyWrapper(object):
         self.proxy_user = ""
         self.proxy_password = ""
         self.has_proxy = False
-        
+
         if self.proxy_name:
             self.proxy_string = "weechat.proxy.{}".format(self.proxy_name)
             self.proxy_type = w.config_string(weechat.config_get("{}.type".format(self.proxy_string)))
@@ -220,21 +220,21 @@ class ProxyWrapper(object):
                 self.has_proxy = True
             else:
                 w.prnt("", "\nWarning: weechat.network.proxy_curl is set to {} type (name : {}, conf string : {}). Only HTTP proxy is supported.\n\n".format(self.proxy_type, self.proxy_name, self.proxy_string))
-        
+
     def curl(self):
         if not self.has_proxy:
             return ""
-        
+
         if self.proxy_user and self.proxy_password:
             user = "{}:{}@".format(self.proxy_user, self.proxy_password)
         else:
             user = ""
-                    
+
         if self.proxy_port:
             port = ":{}".format(self.proxy_port)
         else:
             port = ""
-                
+
         return "--proxy {}{}{}".format(user, self.proxy_address, port)
 
 
@@ -1511,7 +1511,7 @@ class SlackChannel(object):
             m.message_json.update(message_json)
         if text:
             m.change_text(text)
-        new_text = m.render(force=True)
+        new_text = self.render(m, force=True)
         modify_buffer_line(self.channel_buffer, new_text, ts.major, ts.minor)
 
     def edit_nth_previous_message(self, n, old, new, flags):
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -2536,6 +2536,10 @@ def subprocess_thread_message(message_json, eventrouter, channel, team):
             if parent_message.thread_channel:
                 parent_message.thread_channel.buffer_prnt(message.sender, text, message.ts, tag_nick=message.sender_plain)
 
+            if config.thread_messages_in_channel:
+                channel.buffer_prnt(
+                    message.sender, '[{}] {}'.format(parent_message.hash, text), message.ts, tag_nick=message.sender_plain)
```
trygveaa: I think we should colorize the prefix, in the same color as the suffix, i.e. `w.color(config.thread_suffix_color)`.

wee_slack.py
```diff
@@ -3853,6 +3857,9 @@ class PluginConfig(object):
             desc='When activity occurs on a buffer, unhide it even if it was'
             ' previously hidden (whether by the user or by the'
             ' distracting_channels setting).'),
+        'thread_messages_in_channel': Setting(
```
trygveaa: Can you place this right above `thread_suffix_color`, i.e. alphabetically?

mkaito:
Just from a commit discipline point of view, those whitespace changes should be in a separate commit. And trailing whitespace should never have been accepted in the first place, but :man_shrugging:

mkaito:
On a different note, this is amazing for those of us who rely on Slack for work. People *love* them some threads, and threads are currently pretty hard to consume. This PR is amazing news.

janza:
I've updated the code and added color to those hash markers.

trygveaa:
I noticed now that if the message is edited, the prefix is lost. To fix this we have to add the prefix in the render method of the message instead of directly printing it directly. The problem with that is that the prefix shouldn't be printed in actual threads, so I think we need some flag to signal if the message should be rendered for a channel or a thread.

janza:
Ah, good catch, I'll try to fix that.

tomatosoupcan:
My hero, thanks for this

smaslennikov:
This is gonna be gold

mkaito:
> I noticed now that if the message is edited, the prefix is lost.

Oh, so that's what's happening. I noticed that thread messages shown in the parent channel would frequently not show the thread-id prefix. But I think this happens not just on edits. It would probably be solved in the same way; by adding the prefix to the render method.

trygveaa:
> But I think this happens not just on edits. It would probably be solved in the same way; by adding the prefix to the render method.

Yeah, it happens for anything that changes the message, so also if someone reacts to the message, or if it contains a link which is unfurled. And yes, they will all be fixed by moving it to the render method.

I thought of another thing. We should tag these messages with a separate tag, e.g. add `thread` to the tag function and use that. This would allow you to choose which channels you want thread messages to appear in, or hide all by default and show them by showing filtered messages.

janza:
I've updated the PR with changes in how messages are rendered. Did this by adding another render method to `SlackChannel` and `SlackThreadChannel` which takes care of rendering the prefix in different types of channels. 

I've noticed that messages in thread channels don't get updated (when editing, adding emoji, etc) but that is the case in master as well so I didn't bother with that for now.

I can make another PR for the message tagging part.

mkaito:
I'm not sure what's going on, but with this latest commit applied, it seems to... not show any recent messages. It loads channel history but only shows a part of it... maybe? This is weird.

mkaito:
Been using for a few days, works quite well!

Only thing I'm missing is a `reply <thread id> <message>` command at this point.

janza:
Opened #630 for the command to reply to existing threads.

trygveaa:
Thanks!

PR #617, opened 2018-08-25
Subscribe threads

immae:
This notifies (or automatically opens) threads concerning the user:
- When someone mentions him in any subthread
- When a response is given to a thread in which the user participated, but the thread was closed inbetween
- When a thread is open in response to a message the user wrote.

This implementation works, but is still not complete for the last two cases:
The user may receive a "notifiable" thread message associated to a channel which he didn’t open yet (since last reboot for instance), and won’t be notified of that until he goes to open the channel and "fetches history" (better than nothing, but getting the notification earlier would require fetching the history of the channel as soon as we get notified, which is probably not a good idea)

I think this makes the feature quite usable already and that the above bug should be addressed in a later PR, as it would require some non-trivial action to fix.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2124,6 +2124,12 @@ class SlackMessage(object):
     def __hash__(self):
         return hash(self.ts)
 
+    def open_thread(self, switch=False):
+        self.thread_channel = SlackThreadChannel(EVENTROUTER, self)
+        self.thread_channel.open()
+        if switch:
+            w.buffer_set(self.thread_channel.channel_buffer, "display", "1")
+
     def render(self, force=False):
         text = render(self.message_json, self.team, self.channel, force)
         if (self.message_json.get('subtype') == 'me_message' and
@@ -3398,12 +3404,7 @@ def thread_command_callback(data, current_buffer, args):
                     pm = channel.messages[SlackTS(args[1])]
                 except:
                     pm = channel.hashed_messages[args[1]]
-                tc = SlackThreadChannel(EVENTROUTER, pm)
-                pm.thread_channel = tc
-                tc.open()
-                # tc.create_buffer()
-                if config.switch_buffer_on_join:
-                    w.buffer_set(tc.channel_buffer, "display", "1")
+                pm.open_thread(switch=config.switch_buffer_on_join)
                 return w.WEECHAT_RC_OK_EAT
         elif args[0] == '/reply':
             count = int(args[1])
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1110,8 +1110,9 @@ class SlackTeam(object):
     def formatted_name(self, **kwargs):
         return self.domain
 
-    def buffer_prnt(self, data):
-        w.prnt_date_tags(self.channel_buffer, SlackTS().major, tag("team"), data)
+    def buffer_prnt(self, data, message=False):
+        tag_name = "team_message" if message else "team_info"
+        w.prnt_date_tags(self.channel_buffer, SlackTS().major, tag(tag_name), data)
 
     def send_message(self, message, subtype=None, request_dict_ext={}):
         w.prnt("", "ERROR: Sending a message in the team buffer is not supported")
@@ -2199,6 +2200,24 @@ class SlackMessage(object):
         else:
             pass
 
+    def has_mention(self):
+        return w.string_has_highlight(self.message_json.get('text'), ",".join(self.channel.mentions()))
+
+    def notify_thread(self, action=None, sender_id=None):
+        if config.auto_open_threads:
+            self.open_thread()
+        elif sender_id != self.team.myidentifier:
+            if action == "mention":
+                template = "You were mentioned in thread {hash}, channel {channel}"
+            elif action == "participant":
+                template = "New message in thread {hash}, channel {channel} in which you participated"
+            elif action == "response":
+                template = "New message in thread {hash} in response to own message in {channel}"
+            else:
+                template = "Notification for message in thread {hash}, channel {channel}"
+            message = template.format(hash=self.hash, channel=self.channel.formatted_name())
+
+            self.team.buffer_prnt(message, message=True)
 
 class SlackThreadMessage(SlackMessage):
 
@@ -2548,8 +2567,10 @@ def subprocess_thread_message(message_json, eventrouter, channel, team):
 
             text = message.render()
             # channel.buffer_prnt(message.sender, text, message.ts, **kwargs)
-            if parent_message.thread_channel:
+            if parent_message.thread_channel and parent_message.thread_channel.active:
                 parent_message.thread_channel.buffer_prnt(message.sender, text, message.ts, tag_nick=message.sender_plain)
+            elif message.ts > channel.last_read and message.has_mention():
+                parent_message.notify_thread(action="mention", sender_id=message_json["user"])
 
 #    channel = channels.find(message_json["channel"])
 #    server = channel.server
@@ -2587,8 +2608,17 @@ def subprocess_channel_leave(message_json, eventrouter, channel, team):
 
 
 def subprocess_message_replied(message_json, eventrouter, channel, team):
-    pass
-
+    parent_ts = message_json["message"].get("thread_ts")
+    parent_message = channel.messages.get(SlackTS(parent_ts))
+    # Thread exists but is not open yet
+    if parent_message is not None \
+            and not (parent_message.thread_channel and parent_message.thread_channel.active):
+        channel.hash_message(parent_ts)
+        last_message = max(message_json["message"]["replies"], key=lambda x: x["ts"])
+        if message_json["message"].get("user") == team.myidentifier:
+            parent_message.notify_thread(action="response", sender_id=last_message["user"])
+        elif any(team.myidentifier == r["user"] for r in message_json["message"]["replies"]):
+            parent_message.notify_thread(action="participant", sender_id=last_message["user"])
 
 def subprocess_message_changed(message_json, eventrouter, channel, team):
     new_message = message_json.get("message", None)
@@ -3103,7 +3133,8 @@ def tag(tagset, user=None):
         default_tag = 'nick_unknown'
     tagsets = {
         # messages in the team/server buffer, e.g. "new channel created"
-        "team": "no_highlight,log3",
+        "team_info": "no_highlight,log3",
+        "team_message": "irc_privmsg,notify_message,log1",
         # when replaying something old
         "backlog": "irc_privmsg,no_highlight,notify_none,logger_backlog",
         # when posting messages to a muted channel
@@ -3767,6 +3798,10 @@ class PluginConfig(object):
     # Following this procedure, the keys remain the same, but the values are
     # the real (python) values of the settings.
     default_settings = {
+        'auto_open_threads': Setting(
+            default='false',
+            desc='Automatically open threads when mentioned or in'
+            'response to own messages.'),
         'background_load_all_history': Setting(
             default='false',
             desc='Load history for each channel in the background as soon as it'
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1368,11 +1368,16 @@ class SlackChannel(object):
     def set_related_server(self, team):
         self.team = team
 
+    def mentions(self):
+        return {'@' + self.team.nick, self.team.myidentifier}
+
+    def highlights(self):
+        return self.team.highlight_words.union(self.mentions()).union({"!here", "!channel", "!everyone"})
+
     def set_highlights(self):
         # highlight my own name and any set highlights
         if self.channel_buffer:
-            highlights = self.team.highlight_words.union({'@' + self.team.nick, self.team.myidentifier, "!here", "!channel", "!everyone"})
-            h_str = ",".join(highlights)
+            h_str = ",".join(self.highlights())
             w.buffer_set(self.channel_buffer, "highlight_words", h_str)
 
     def create_buffer(self):


wee_slack.py
```diff
@@ -3867,6 +3908,10 @@ class PluginConfig(object):
             desc='When activity occurs on a buffer, unhide it even if it was'
             ' previously hidden (whether by the user or by the'
             ' distracting_channels setting).'),
+        'auto_open_threads': Setting(
```
trygveaa: Can you place this alphabetically in the list?

immae: Done

wee_slack.py
```diff
@@ -3104,6 +3144,7 @@ def tag(tagset, user=None):
     tagsets = {
         # messages in the team/server buffer, e.g. "new channel created"
         "team": "no_highlight,log3",
+        "team_highlight": "irc_privmsg,notify_message,log1",
```
trygveaa: This is not a necessarily a highlight, just a normal notify that allows highlights. You won't get highlighted by any of the messages that prints with this (well, unless you add one of those words to the highlight list), so not sure we need this separate taglist?

wee_slack.py
```diff
@@ -2548,8 +2554,17 @@ def subprocess_thread_message(message_json, eventrouter, channel, team):
 
             text = message.render()
             # channel.buffer_prnt(message.sender, text, message.ts, **kwargs)
-            if parent_message.thread_channel:
+            if parent_message.thread_channel and parent_message.thread_channel.active:
                 parent_message.thread_channel.buffer_prnt(message.sender, text, message.ts, tag_nick=message.sender_plain)
+            elif message.has_mention():
+                if config.auto_open_threads:
```
trygveaa: Can you create a function for this block, so you don't have to repeat it?

wee_slack.py
```diff
@@ -2587,8 +2602,33 @@ def subprocess_channel_leave(message_json, eventrouter, channel, team):
 
 
 def subprocess_message_replied(message_json, eventrouter, channel, team):
-    pass
-
+    if "message" not in message_json:
```
trygveaa: Can this happen?

wee_slack.py
```diff
@@ -2587,8 +2602,33 @@ def subprocess_channel_leave(message_json, eventrouter, channel, team):
...
+        return
+    parent_ts = message_json["message"].get("thread_ts")
+    if parent_ts:
+        parent_message = channel.messages.get(SlackTS(parent_ts), None)
```
trygveaa: `None` is unnecessary here, it defaults to that.

wee_slack.py
```diff
@@ -2587,8 +2602,33 @@ def subprocess_channel_leave(message_json, eventrouter, channel, team):
 
 
 def subprocess_message_replied(message_json, eventrouter, channel, team):
-    pass
-
+    if "message" not in message_json:
+        return
+    parent_ts = message_json["message"].get("thread_ts")
+    if parent_ts:
```
trygveaa: I think this `if` is unnecessary too?

wee_slack.py
```diff
@@ -2199,6 +2203,8 @@ def remove_reaction(self, reaction, user):
         else:
             pass
 
+    def has_mention(self):
+        return any(word in self.message_json.get('text') for word in self.channel.mentions())
```
trygveaa: You can use `w.string_has_highlight` to check if the string matches.

https://weechat.org/files/doc/stable/weechat_plugin_api.en.html#_string_has_highlight

wee_slack.py
immae: I guess not but I couldn’t find a real specification for it in the slack api documentation

wee_slack.py
trygveaa: This part:
```
                if config.auto_open_threads:
                     parent_message.open_thread()
                else:
                    team.buffer_prnt(
                            "You were mentionned in thread {}, channel {}".format(
                                parent_message.hash,
                                channel.formatted_name()
                            ), highlight=True)
```

Except for a different message, this is repeated three times.

wee_slack.py
immae: team messages (the "team" version) are not even marked in my weechat (I don’t know if I did something about it or if it came that way), so I only see them if I think of going to the team buffer.

These ones are "more important" than the usual debug messages, so I guess they should be marked as important? Otherwise there is no point in notifying...

Given that, I think there should be a difference between the "debug" messages tags and the "important" messages (maybe highlight isn’t the right term, I have no opinion about that), and as far as I understand tags are the way to do it in weechat (?)

wee_slack.py
trygveaa: [The example in the docs](https://api.slack.com/events/message/message_replied) includes it, and the [docs describing threads](https://api.slack.com/docs/message-threading) describes what the message will contain, so I think it's safe to assume it will always be there.

wee_slack.py
immae: Understood. Something like `parent_message.notify_thread(action=("mention"|"participant"|"response"))` ?

immae: Or `parent_message.notify_thread("You were mentionned in bla bla bla")` ?

wee_slack.py
trygveaa: Hm, the core buffer appears in my hotlist when we print to it. I guess it depends on some weechat settings. But sure, I agree that the messages are different. Can you rename `team_highlight` to `team_message` and `team` to `team_info` though?

immae: :+1:

wee_slack.py
trygveaa: Yeah, the first of those is fine.

wee_slack.py
```diff
@@ -1110,8 +1110,12 @@ def set_highlight_words(self, highlight_str):
     def formatted_name(self, **kwargs):
         return self.domain
 
-    def buffer_prnt(self, data):
-        w.prnt_date_tags(self.channel_buffer, SlackTS().major, tag("team"), data)
+    def buffer_prnt(self, data, message=False):
+        if message:
+            tag_name = "team_message"
+        else:
+            tag_name = "team_info"
```
trygveaa: This can go on one line:
```
tag_name = "team_message" if message else "team_info"
```

wee_slack.py
```diff
@@ -2199,6 +2203,24 @@ def remove_reaction(self, reaction, user):
...
+                template = "You were mentionned in thread {hash}, channel {channel}"
+            elif action == "participant":
+                template = "New message in thread {hash}, channel {channel} in which you participated"
+            elif action == "response":
+                template = "New message in thread {hash} in response to own message in {channel}"
+            else:
+                template = "Notification for message in thread {hash}, channel {channel}"
+            message = template.format(hash=self.hash, channel= self.channel.formatted_name())
+
+            team.buffer_prnt(message, message=True)
```
trygveaa: You're missing `self.` here.

wee_slack.py
```diff
@@ -2199,6 +2203,24 @@ def remove_reaction(self, reaction, user):
...
+        else:
+            if action == "mention":
+                template = "You were mentionned in thread {hash}, channel {channel}"
+            elif action == "participant":
+                template = "New message in thread {hash}, channel {channel} in which you participated"
+            elif action == "response":
+                template = "New message in thread {hash} in response to own message in {channel}"
+            else:
+                template = "Notification for message in thread {hash}, channel {channel}"
+            message = template.format(hash=self.hash, channel= self.channel.formatted_name())
```
trygveaa: Remove the space after `channel=`.

wee_slack.py

wee_slack.py

wee_slack.py

wee_slack.py

wee_slack.py

wee_slack.py

wee_slack.py

wee_slack.py

wee_slack.py

immae:
All done so far I think

trygveaa:
If the channel history is not loaded, you don't get any notification. I'm not sure what the best way to support this is, since we don't have the hash yet, but we should try to fix it.

When the channel history is loaded, you might get old 'You were mentioned in' messages. These should not be printed for backlog messages.

immae:
> When the channel history is loaded, you might get old 'You were mentioned in' messages. These should not be printed for backlog messages.

But then that means that if you log out of slack, you will never be warned about important threads unless you’re going to the web application?

wee_slack.py
```diff
@@ -2188,6 +2200,24 @@ def remove_reaction(self, reaction, user):
...
 
+    def has_mention(self):
+        return w.string_has_highlight(self.message_json.get('text'), ",".join(self.channel.mentions()))
+
+    def notify_thread(self, action=None):
+        if config.auto_open_threads:
+            self.open_thread()
+        else:
+            if action == "mention":
+                template = "You were mentionned in thread {hash}, channel {channel}"
```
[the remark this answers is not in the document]
immae: :white_check_mark:

wee_slack.py
```diff
@@ -2576,8 +2608,16 @@ def subprocess_channel_leave(message_json, eventrouter, channel, team):
...
+    parent_ts = message_json["message"].get("thread_ts")
+    parent_message = channel.messages.get(SlackTS(parent_ts))
+    # Thread exists but is not open yet
+    if parent_message is not None \
+            and not (parent_message.thread_channel and parent_message.thread_channel.active):
+        channel.hash_message(parent_ts)
+        if message_json["message"].get("user") == team.myidentifier:
+            parent_message.notify_thread(action="response")
+        elif any(team.myidentifier == r["user"] for r in message_json["message"]["replies"]):
+            parent_message.notify_thread(action="participant")
```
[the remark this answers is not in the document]
immae: I changed, but I’m not happy with the solution which feels hacky. If you have a better one...

mkaito:
> When the channel history is loaded, you might get old 'You were mentioned in' messages. These should not be printed for backlog messages.

You mean like the 200 notifications I get every time I switch to a DM buffer?

trygveaa:
@immae:
> But then that means that if you log out of slack, you will never be warned about important threads unless you’re going to the web application?

No, it should notify for new thread messages. What I meant is that messages you have already read should be ignored. With the current behavior, a notification is printed for every thread message in the history (as far back as we load history) when you switch to a channel after reload (or for all channels without switching to them if background_load_all_history is true).

I think we should implement this by ignoring thread messages with a `ts` earlier than `channel.last_read`. Note that this isn't optimal though, since it just keeps track of the last read message in the channel and isn't affected by threads. There is two issues with that:

* `last_read` isn't set when reading thread messages, only when reading the channel. So you will still get notification for thread messages newer than any channel messages when you load wee-slack. This isn't worse than the current behavior though. I don't think we should set `last_read` to something newer than an unread message in the channel when reading a thread, but if it's already set to the latest message in the channel, we could probably set it to the thread message ts.

* If there is an older message in a thread that you haven't read, and a newer message arrives in the channel which you read, last_read will be set to newer than the thread message. This means that if you reload wee-slack, you won't get the notification again. We will of course notify for the thread message when it first arrives if you have wee-slack running. And if you don't have wee-slack running when the message arrives, and don't read in other clients, you will get the notification when you load wee-slack. So I don't think this is a huge problem.

I'm open for other suggestions on how to fix this, but since I can't see any documentation for handling thread notifications or unread state of thread messages, I don't think we have a lot of options.

@mkaito:
> You mean like the 200 notifications I get every time I switch to a DM buffer?

I'm not sure where you're going with this. If this is an existing problem you experience, you should create a separate issue for that. And regardless of existing bugs, we should try to make new features behave like we want, so we don't introduce more undesired behavior.

trygveaa:
> But then that means that if you log out of slack, you will never be warned about important threads unless you’re going to the web application?

This is already the case in this branch btw. Two of the notification types depend on a reply event, and that is only sent when you're connected. So you won't get these for things that hawhen you load the plugin.

immae:
```
index 99e1b09..f2fcfbf 100644
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2569,7 +2569,7 @@ def subprocess_thread_message(message_json, eventrouter, channel, team):
             # channel.buffer_prnt(message.sender, text, message.ts, **kwargs)
             if parent_message.thread_channel and parent_message.thread_channel.active:
                 parent_message.thread_channel.buffer_prnt(message.sender, text, message.ts, tag_nick=message.sender_plain)                                                                                                                   
-            elif message.has_mention():
+            elif message.ts > channel.last_read and message.has_mention():
                 parent_message.notify_thread(action="mention")

 #    channel = channels.find(message_json["channel"])
```

wee_slack.py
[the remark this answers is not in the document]
trygveaa: Ugh, I took it for granted that the `message_replied` event had a field for the user that wrote the message, but I guess not...

I don't see any other way this can be checked from the `message_replied` event. I would think the replies array is sorted, but maybe it's best like you did to be safe. You have to check it for both cases though, so place it with the first if (before hash_message).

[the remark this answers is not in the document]
trygveaa: You will have to do a similar check in `subprocess_thread_message`. There you can just check the user of the message though.

yeled:
this is working beautifully for me as is.

wee_slack.py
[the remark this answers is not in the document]
immae: Done in both cases. I’m wondering wether the check should be moved to the notify_thread method, but that would mean passing the thread message as a parameter to the method, which seems wrong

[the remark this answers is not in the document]
trygveaa: What about passing just the user to notify_thread?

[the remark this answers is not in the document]
immae: ok

trygveaa:
Thanks so much for implementing this!

paulRbr:
Huge thanks @immae :green_heart:

PR #621, opened 2018-08-29
Use emoji_aliases when available

immae:
This makes use of the emoji_aliases script when available, to convert emojis strings to their real value.

I couldn’t find a proper way to do it reliably (listening to an event or so), so I make an attempt into importing the script, which turns out to be a python script, in wee-slack. This should work correctly if the script is not installed.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -25,6 +25,11 @@ try:
 except:
     from StringIO import StringIO
 
+try:
+    from emoji_aliases import convert_aliases_to_emoji
+except:
+    convert_aliases_to_emoji = None
+
 from websocket import create_connection, WebSocketConnectionClosedException
 
 # hack to make tests possible.. better way?
@@ -3161,6 +3166,12 @@ def modify_buffer_line(buffer_pointer, ts, new_text):
 
     for pointer, line in zip(pointers, lines):
         data = w.hdata_pointer(hdata.line, pointer, 'data')
+        if convert_aliases_to_emoji is not None:
+            try:
+                line = convert_aliases_to_emoji(None, None, None, line.encode('utf-8'))
+            except:
+                dbg("emoji issue: {}\n".format(traceback.format_exc()))
+                pass
         w.hdata_update(hdata.line_data, data, {"message": line})
 
     return w.WEECHAT_RC_OK


FabianFrank:
I've tried to use this PR locally and failed, basically even when the code runs it seems to never substitute any emojis (I made sure that emoji_aliases.py is loaded correctly and actually being called). I'd like to help getting this ready for merge, do you have some pointers about your setup so I can try to get it to work?

trygveaa:
This completely breaks wee-slack for me if `emoji_aliases.py` is installed, but not loaded. I also get some errors if it isn't installed or is loaded. This happens because the code in emoji_aliases is run when it's imported, and this includes some weechat commands.

Also, emoji_aliases doesn't change the lines wee-slack print for me, so this only makes emojies used when a line is edited.

I have previously said that converting to emoji should be done with a separate script instead of wee-slack, but you could be right that it's not possible with the line editing we do so maybe we should do something like this.

We can't use another weechat script for it though, but I think we could use the [`emoji` package from PyPI](https://pypi.org/project/emoji/).

immae:
I’m surprised it doesn’t work for you, as on my side it works without any problem.

What are the errors you get?

trygveaa:
If it's loaded I get these errors when loading wee-slack:
```
01:40 =!= python: unable to register script "emoji_aliases" (another script already exists with this name)
01:40 =!= python: unable to call function "hook_modifier", script is not initialized (script: -)
01:40 =!= python: unable to call function "hook_modifier", script is not initialized (script: -)
01:40 =!= python: unable to call function "hook_modifier", script is not initialized (script: -)
01:40 =!= python: unable to call function "hook_modifier", script is not initialized (script: -)
01:40 =!= python: unable to call function "hook_modifier", script is not initialized (script: -)
01:40 =!= python: unable to call function "hook_modifier", script is not initialized (script: -)
01:40 =!= python: unable to call function "hook_modifier", script is not initialized (script: -)
01:40 =!= python: unable to call function "hook_modifier", script is not initialized (script: -)
01:40 =!= python: unable to call function "hook_modifier", script is not initialized (script: -)
01:40 =!= python: unable to call function "hook_modifier", script is not initialized (script: -)
01:40 =!= python: unable to call function "hook_modifier", script is not initialized (script: -)
```

If I try to load wee-slack with emoji_aliases installed, but not loaded, I just get this output and nothing more happens:
```
01:55 python: loading script "/home/trygve/.weechat/python/autoload/wee_slack.py"
01:55 python: registered script "emoji_aliases", version 1.0.2 (Convert emoji aliases to unicode emoji.)
01:55 =!= python: script "emoji_aliases" already registered (register ignored)
```

I was wrong above about getting errors if emoji_aliases isn't installed. I tested it after running `/script remove emoji_aliases.py `, but it turns out that leaves the .pyc file, so it could still be imported.

immae:
I finally found by chance the probable origin of the problem (thanks to @paulRbr ). There are two things:
- In my setup, I have emoji_aliases.py and wee_slack.py in weechat/python/ directly, autoload only contains symbolic links to them (that’s important for import paths apparently)
- Also, I slightly modified emoji_aliases.py, the important path being that I wrap the w.register with an `if __name__ == "__main__":` so that it doesn’t load when wee-slack imports it.

I guess that last point is a no-go to have it used, since it requires modifying another module. However, this module can be made simple enough to be included in wee_slack directly if you think it can be useful.

trygveaa:
> In my setup, I have emoji_aliases.py and wee_slack.py in weechat/python/ directly, autoload only contains symbolic links to them (that’s important for import paths apparently)

Hm, I don't think this matters? Afaik, python tries to import from the directory the file loaded from, no matter if it is a symlink or not.

> Also, I slightly modified emoji_aliases.py, the important path being that I wrap the w.register with an `if __name__ == "__main__":` so that it doesn’t load when wee-slack imports it.

Yes, that would solve the issue, so my statement that we can't use another weechat script wasn't really correct.

> I guess that last point is a no-go to have it used, since it requires modifying another module.

We would have to send a PR for the change, and get the script in the weechat repo updated. However, it would still be a problem, because people might have an old version of it which would cause problems.

I actually started on implementing emoji rendering directly a while ago, but I had some issues with skin tones, and multibyte emojis. I've pushed it to https://github.com/trygveaa/wee-slack/tree/emoji-rendring if you want to take a look.

immae:
For the import, that’s what we thought at first but it was actually important to have the file present in .weechat/python directly, so I guess the path doesn’t contain "." in python when loaded via weechat.

I looked at your branch, you went much further than what I did. In addition, I concluded (maybe wrongly) some time ago that "splitted" emojis (skin tone, flags) are simply not handled in terminals currently, so maybe it’s not worth trying to solve them.
The test that lead to that conclusion was that copying a flag from browser to my terminal showed two "emojis", and conversly the two emojis were combined into a flag when I copied from terminal to browser, so the issue was only at display and not on weechat side

trygveaa:
> For the import, that’s what we thought at first but it was actually important to have the file present in .weechat/python directly, so I guess the path doesn’t contain "." in python when loaded via weechat.

It works for me if I have it in the directory I start weechat from, if that's what you meant by `.`. Just having it in `autoload` doesn't work though, so if people only have it there and not in `.weechat/python`, that would be an issue. Anyways, not really important.

> I concluded (maybe wrongly) some time ago that "splitted" emojis (skin tone, flags) are simply not handled in terminals currently, so maybe it’s not worth trying to solve them.
The test that lead to that conclusion was that copying a flag from browser to my terminal showed two "emojis", and conversly the two emojis were combined into a flag when I copied from terminal to browser, so the issue was only at display and not on weechat side

They work in the terminal I use (kitty), but I guess not many others support them.

immae:
This PR actually only works with a hacked version of emoji_aliases, which I didn’t realize/remember at the time when I wrote it.
Closing it as it is now unrelevant

PR #622, opened 2018-08-29
Allow reactions and edition of messages in threads

immae:
This PR allows edition/deletion and reactions to messages in threads, by using the same mechanisms as the ones used for other channels.

I made both SlackChannel and SlackThreadChannel inherit from a common class SlackChannelCommon that will posess common methods (including the ones necessary for reactions and edition/deletion)


trygveaa:
Thanks!

PR #623, opened 2018-08-30
reactions: Add a config parameter to change color of suffixed reacs

paulRbr:
As of for now the `darkgray` color is hard coded inside the plugin for reactions suffixed on messages.

My current terminal background was `darkgray` so it caused a problem :see_no_evil:.

I reckon it would be nice to have a setting to overide this color with a plugin config.

WDYT?

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -3064,9 +3064,10 @@ def create_reaction_string(reactions):
     count = 0
     if not isinstance(reactions, list):
         reaction_string = " {}[{}]{}".format(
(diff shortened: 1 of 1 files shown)


trygveaa:
Thanks!

trygveaa:
FYI, I renamed the option to `color_reaction_suffix` (and the existing `thread_suffix_color` to `color_thread_suffix`), so the color options are grouped together.

paulRbr:
Thanks! Looks good :)

PR #626, opened 2018-08-31
Select message cursor

immae:
This permits to select a message based on his id to /reply, edit, delete or "react" to a message. In "mouse" mode, clicking on a message pastes the id of the line in the input. In "cursor" mode, "M" permits to put the id of the line, "R" adds "/reply <id> ", and "D" deletes the line ("s///").

This PR is dependent on
https://github.com/wee-slack/wee-slack/pull/622

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,6 +267,24 @@ To enable tab completion of emojis, copy or symlink the `weemoji.json` file to y
 /set weechat.completion.default_template "%(nicks)|%(irc_channels)|%(emoji)"
 ```
 
+#### Cursor and mouse mode
+
+The cursor mode and mouse mode can be used to interact with older messages, for editing, deleting, reacting and replying to a message. The default behavior when right-clicking on a message is to paste its id in the input. It can be used in `/reply`, `s/` substitution/deletion and in `+:emoji:` commands instead of a message number.
+
+In cursor mode, the `M` key achieves the same result (memo: the default for weechat is to paste the message with `m`, `M` simply copies the id).
+In addition, `R` will prepare a `/reply id` and `D` will delete the message (provided it’s yours).
+
+Please see weechat’s documentation about [how to use the cursor mode](https://weechat.org/files/doc/stable/weechat_user.en.html#key_bindings_cursor_context) or [adapt the bindings](https://weechat.org/files/doc/stable/weechat_user.en.html#command_weechat_key) to your preference.
+
+Default:
+```
+/key bindctxt mouse @chat(python.*.slack.com.*):button2 hsignal:slack_mouse
+/key bindctxt cursor @chat(python.*.slack.com.*):M hsignal:slack_cursor_message
+/key bindctxt cursor @chat(python.*.slack.com.*):D hsignal:slack_cursor_delete
+/key bindctxt cursor @chat(python.*.slack.com.*):R hsignal:slack_cursor_reply
+```
+hsignals `slack_mouse` and `slack_cursor_message` currently have the same meaning but may be subject to evolutions.
+
 Removing a team
 ---------------
 
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -716,15 +716,24 @@ def buffer_input_callback(signal, buffer_ptr, data):
     if not channel:
         return w.WEECHAT_RC_ERROR
 
-    reaction = re.match("^(\d*)(\+|-):(.*):\s*$", data)
-    substitute = re.match("^(\d*)s/", data)
+    def get_id(message_id):
+        if not message_id:
+            return 1
+        elif message_id[0] == "$":
+            return message_id[1:]
+        else:
+            return int(message_id)
+
+    message_id_regex = "(\d*|\$[0-9a-fA-F]{3,})"
+    reaction = re.match("^{}(\+|-):(.*):\s*$".format(message_id_regex), data)
+    substitute = re.match("^{}s/".format(message_id_regex), data)
     if reaction:
         if reaction.group(2) == "+":
-            channel.send_add_reaction(int(reaction.group(1) or 1), reaction.group(3))
+            channel.send_add_reaction(get_id(reaction.group(1)), reaction.group(3))
         elif reaction.group(2) == "-":
-            channel.send_remove_reaction(int(reaction.group(1) or 1), reaction.group(3))
+            channel.send_remove_reaction(get_id(reaction.group(1)), reaction.group(3))
     elif substitute:
-        msgno = int(substitute.group(1) or 1)
+        msg_id = get_id(substitute.group(1))
         try:
             old, new, flags = re.split(r'(?<!\\)/', data)[1:]
         except ValueError:
@@ -734,7 +743,7 @@ def buffer_input_callback(signal, buffer_ptr, data):
             # rid of escapes.
             new = new.replace(r'\/', '/')
             old = old.replace(r'\/', '/')
-            channel.edit_nth_previous_message(msgno, old, new, flags)
+            channel.edit_nth_previous_message(msg_id, old, new, flags)
     else:
         if data.startswith(('//', ' ')):
             data = data[1:]
@@ -1248,22 +1257,31 @@ class SlackTeam(object):
 
 
 class SlackChannelCommon(object):
-    def send_add_reaction(self, msg_number, reaction):
-        self.send_change_reaction("reactions.add", msg_number, reaction)
+    def send_add_reaction(self, msg_id, reaction):
+        self.send_change_reaction("reactions.add", msg_id, reaction)
 
-    def send_remove_reaction(self, msg_number, reaction):
-        self.send_change_reaction("reactions.remove", msg_number, reaction)
+    def send_remove_reaction(self, msg_id, reaction):
+        self.send_change_reaction("reactions.remove", msg_id, reaction)
 
-    def send_change_reaction(self, method, msg_number, reaction):
-        if 0 < msg_number < len(self.messages):
+    def send_change_reaction(self, method, msg_id, reaction):
+        if type(msg_id) is not int:
+            if msg_id in self.hashed_messages:
+                timestamp = str(self.hashed_messages[msg_id].ts)
+            else:
+                return
+        elif 0 < msg_id <= len(self.messages):
             keys = self.main_message_keys_reversed()
-            timestamp = next(islice(keys, msg_number - 1, None))
-            data = {"channel": self.identifier, "timestamp": timestamp, "name": reaction}
-            s = SlackRequest(self.team.token, method, data)
-            self.eventrouter.receive(s)
+            timestamp = next(islice(keys, msg_id - 1, None))
+        else:
+            return
+        data = {"channel": self.identifier, "timestamp": timestamp, "name": reaction}
+        s = SlackRequest(self.team.token, method, data)
+        self.eventrouter.receive(s)
 
-    def edit_nth_previous_message(self, n, old, new, flags):
-        message = self.my_last_message(n)
+    def edit_nth_previous_message(self, msg_id, old, new, flags):
+        message = self.my_last_message(msg_id)
+        if message is None:
+            return
         if new == "" and old == "":
             s = SlackRequest(self.team.token, "chat.delete", {"channel": self.identifier, "ts": message['ts']}, team_hash=self.team.team_hash, channel_identifier=self.identifier)
             self.eventrouter.receive(s)
@@ -1276,13 +1294,18 @@ class SlackChannelCommon(object):
                 s = SlackRequest(self.team.token, "chat.update", {"channel": self.identifier, "ts": message['ts'], "text": new_message}, team_hash=self.team.team_hash, channel_identifier=self.identifier)
                 self.eventrouter.receive(s)
 
-    def my_last_message(self, msgno):
-        for key in self.main_message_keys_reversed():
-            m = self.messages[key]
-            if "user" in m.message_json and "text" in m.message_json and m.message_json["user"] == self.team.myidentifier:
-                msgno -= 1
-                if msgno == 0:
-                    return m.message_json
+    def my_last_message(self, msg_id):
+        if type(msg_id) is not int:
+            m = self.hashed_messages.get(msg_id)
+            if m is not None and m.message_json.get("user") == self.team.myidentifier:
+                return m.message_json
+        else:
+            for key in self.main_message_keys_reversed():
+                m = self.messages[key]
+                if m.message_json.get("user") == self.team.myidentifier:
+                    msg_id -= 1
+                    if msg_id == 0:
+                        return m.message_json
 
     def change_message(self, ts, message_json=None, text=None):
         ts = SlackTS(ts)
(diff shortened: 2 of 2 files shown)


immae:
(TLDR: if you use vim, it’s similar to visual mode)

@mkaito : the cursor mode in weechat is a mode where you can move a cursor "freely" on the screen and interact with the part where it is (see https://weechat.org/files/doc/stable/weechat_user.en.html#key_bindings_cursor_context for some default keys)

Basically you run "/cursor" (there is a key for that but I don’t know), then you can use some keys to interact with the "element" where it is on the screen. Here I added "M" key to paste the message id, "D" to delete it and "R" to add a "/reply <id> " automatically.

wee_slack.py
```diff
@@ -704,15 +704,24 @@ def buffer_input_callback(signal, buffer_ptr, data):
...
-    substitute = re.match("^(\d*)s/", data)
+    def get_id(message_id):
+        if not message_id or len(message_id) == 0:
+            return 1
+        elif message_id[0:2] == "0x":
+            return message_id
+        else:
+            return int(message_id)
+
+    message_id = "(\d*|0x[0-9a-fA-F]{12})"
```
yeled: My IDs are not 12 chars long. they are the format `0x6c3d3900`

dipusone: mine too.

ismaelbouyaf: I’m waiting for the dependent PR to be accepted before doing any change here, feel free to modify it locally in the meantime :)

immae: I’m waiting for the dependent PR to be accepted before doing any change here (but the maintainer seems busy with other things at the moment), feel free to modify it locally in the meantime :)

ismaelbouyaf:
The PR is now up to date, as promised. Would you do a review, @trygveaa ? Thanks :)

immae:
The PR is now up to date, as promised. Would you do a review, @trygveaa ? Thanks :)

trygveaa:
Have you thought about what was said on #weechat regarding not exposing internal id's to users, but rather change the input mode or something like that?

Also, can you add some documentation for this?

trygveaa:
@yeled: Does it matter if the message you're replying to is scrolled off the screen?

immae:
I made a change to use `$<thread_id>` (which represents a message id) rather than exposing the internal pointer value to the user, as suggested on #weechat

wee_slack.py
```diff
@@ -1277,12 +1292,17 @@ def edit_nth_previous_message(self, n, old, new, flags):
...
+            m = self.hashed_messages[msgno[1:]]
             if "user" in m.message_json and "text" in m.message_json and m.message_json["user"] == self.team.myidentifier:
-                msgno -= 1
-                if msgno == 0:
-                    return m.message_json
+                return m.message_json
+        else:
+            for key in self.main_message_keys_reversed():
+                m = self.messages[key]
+                if "user" in m.message_json and "text" in m.message_json and m.message_json["user"] == self.team.myidentifier:
```
trygveaa: You can use `.get` instead of checking if `user` is present first. And I don't think we need to check text? (I know you just used the same check as is already here). So:

```suggestion
                if m.message_json.get("user") == self.team.myidentifier:
```

trygveaa:
Using the commands with an unknown message id fails with a stacktrace. These should fail gracefully.

README.md
```diff
@@ -267,6 +267,24 @@ To enable tab completion of emojis, copy or symlink the `weemoji.json` file to y
 /set weechat.completion.default_template "%(nicks)|%(irc_channels)|%(emoji)"
 ```
 
+#### Cursor and mouse mode
+
+The cursor mode and mouse mode can be used to interact with older messages, for editing, suppressing, reacting and replying to a message. The default behavior when right-clicking on a message is to paste its id in the input. It can be used in `/reply`, `s/` substitution/deletion and in `+:emoji:` commands instead of a message number.
```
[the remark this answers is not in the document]
immae: ✓

README.md
```diff
@@ -267,6 +267,24 @@ To enable tab completion of emojis, copy or symlink the `weemoji.json` file to y
...
+
+In cursor mode, the `M` key achieves the same result (memo: the default for weechat is to paste the message with `m`, `M` simply copies the id).
+In addition, `R` will prepare a `/reply id` and `D` will delete the message (provided it’s yours).
+
+Please see weechat’s documentation about how to use the cursor mode or adapt the bindings to your preference.
```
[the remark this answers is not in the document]
immae: ✓

wee_slack.py
```diff
@@ -716,15 +716,24 @@ def buffer_input_callback(signal, buffer_ptr, data):
     if not channel:
         return w.WEECHAT_RC_ERROR
 
-    reaction = re.match("^(\d*)(\+|-):(.*):\s*$", data)
-    substitute = re.match("^(\d*)s/", data)
+    def get_id(message_id):
+        if not message_id or len(message_id) == 0:
```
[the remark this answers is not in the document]

wee_slack.py
```diff
@@ -716,15 +716,24 @@ def buffer_input_callback(signal, buffer_ptr, data):
...
+            return 1
+        elif message_id[0] == "$":
+            return message_id
```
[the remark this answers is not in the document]

wee_slack.py
```diff
@@ -1255,15 +1264,21 @@ def send_remove_reaction(self, msg_number, reaction):
...
-            data = {"channel": self.identifier, "timestamp": timestamp, "name": reaction}
-            s = SlackRequest(self.team.token, method, data)
-            self.eventrouter.receive(s)
+        else:
+            return
+        data = {"channel": self.identifier, "timestamp": timestamp, "name": reaction}
+        s = SlackRequest(self.team.token, method, data)
+        self.eventrouter.receive(s)
 
     def edit_nth_previous_message(self, n, old, new, flags):
```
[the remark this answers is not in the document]

wee_slack.py
```diff
@@ -1255,15 +1264,21 @@ def send_remove_reaction(self, msg_number, reaction):
         self.send_change_reaction("reactions.remove", msg_number, reaction)
 
     def send_change_reaction(self, method, msg_number, reaction):
```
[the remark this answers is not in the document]

wee_slack.py
```diff
@@ -716,15 +716,24 @@ def buffer_input_callback(signal, buffer_ptr, data):
...
     if reaction:
         if reaction.group(2) == "+":
-            channel.send_add_reaction(int(reaction.group(1) or 1), reaction.group(3))
+            channel.send_add_reaction(get_id(reaction.group(1)), reaction.group(3))
         elif reaction.group(2) == "-":
-            channel.send_remove_reaction(int(reaction.group(1) or 1), reaction.group(3))
+            channel.send_remove_reaction(get_id(reaction.group(1)), reaction.group(3))
     elif substitute:
-        msgno = int(substitute.group(1) or 1)
+        msgno = get_id(substitute.group(1))
```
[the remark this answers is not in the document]

wee_slack.py
```diff
@@ -716,15 +716,24 @@ def buffer_input_callback(signal, buffer_ptr, data):
...
-    substitute = re.match("^(\d*)s/", data)
+    def get_id(message_id):
+        if not message_id or len(message_id) == 0:
+            return 1
+        elif message_id[0] == "$":
+            return message_id
+        else:
+            return int(message_id)
+
+    message_id = "(\d*|\$[0-9a-fA-F]{3,})"
```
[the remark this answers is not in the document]

wee_slack.py
```diff
@@ -3828,14 +3880,27 @@ def setup_hooks():
...
+    w.hook_hsignal("slack_cursor_message", "line_event_cb", "message")
+
     # Hooks to fix/implement
     # w.hook_signal('buffer_opened', "buffer_opened_cb", "")
     # w.hook_signal('window_scrolled', "scrolled_cb", "")
     # w.hook_timer(3000, 0, 0, "slack_connection_persistence_cb", "")
 
 ##### END NEW
 
-
```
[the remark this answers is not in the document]
immae: Oups, I missed them, sorry

wee_slack.py
immae: Changed on both occurences

wee_slack.py
```diff
@@ -3693,6 +3717,34 @@ def command_status(data, current_buffer, args):
...
+
+    e = EVENTROUTER
+    if hdata and buffer_pointer in e.weechat_controller.buffers:
+        line_timestamp = w.hdata_time(struct_hdata_line_data, hdata, 'date')
+        line_time_id = w.hdata_integer(struct_hdata_line_data, hdata, 'date_printed')
+        ts = SlackTS("{}.{}".format(int(line_timestamp), line_time_id))
+
+        channel = e.weechat_controller.buffers[buffer_pointer]
+        channel.hash_message(ts)
+        message_hash = "$" + channel.messages[ts].hash
```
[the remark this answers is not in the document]

wee_slack.py
```diff
@@ -3693,6 +3717,34 @@ def command_status(data, current_buffer, args):
...
+
+    struct_hdata_line = w.hdata_get('line')
+    struct_hdata_line_data = w.hdata_get('line_data')
+    hdata = w.hdata_pointer(struct_hdata_line, line_pointer, 'data')
+
+    e = EVENTROUTER
+    if hdata and buffer_pointer in e.weechat_controller.buffers:
+        line_timestamp = w.hdata_time(struct_hdata_line_data, hdata, 'date')
+        line_time_id = w.hdata_integer(struct_hdata_line_data, hdata, 'date_printed')
+        ts = SlackTS("{}.{}".format(int(line_timestamp), line_time_id))
```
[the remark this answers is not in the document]
immae: It works ✓

immae:
Thanks for your review @trygveaa !
I (think I) addressed all the requests, you can find the diff commit here: https://github.com/wee-slack/wee-slack/pull/626/commits/a9950f038d606764d09d6d7358af2195e7f26858

wee_slack.py
```diff
@@ -3693,6 +3727,33 @@ def command_status(data, current_buffer, args):
         s = SlackRequest(team.token, "users.profile.set", {"profile": profile}, team_hash=team.team_hash)
         EVENTROUTER.receive(s)
 
+@utf8_decode
+def line_event_cb(data, signal, hashtable):
+    line_pointer = hashtable["_chat_line"]
```
trygveaa: This is never used.

wee_slack.py
```diff
@@ -3693,6 +3727,33 @@ def command_status(data, current_buffer, args):
...
+    buffer_pointer = hashtable["_buffer"]
+    line_timestamp = hashtable["_chat_line_date"]
+    line_time_id = hashtable["_chat_line_date_printed"]
+
+    e = EVENTROUTER
+    if line_timestamp and line_time_id and buffer_pointer in e.weechat_controller.buffers:
+        ts = SlackTS("{}.{}".format(line_timestamp, line_time_id))
+
+        channel = e.weechat_controller.buffers[buffer_pointer]
+        message_hash = channel.hash_message(ts)
```
trygveaa: This fails with a stacktrace if you try to use cursor mode in a team buffer.

wee_slack.py
```diff
@@ -3693,6 +3727,33 @@ def command_status(data, current_buffer, args):
...
+        if message_hash is None:
+            return w.WEECHAT_RC_OK
+        else:
```
trygveaa: No point in having an else-block when the if-block returns.

wee_slack.py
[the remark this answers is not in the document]
trygveaa: Did you forget this?

wee_slack.py

wee_slack.py

wee_slack.py
immae: Fixed

immae:
All fixed

immae:
Sorry I missed one of your comments, it should be fixed now. Only trivial changes here (except for an `if` to check that we’re in a channel when handling clicks)

wee_slack.py
trygveaa: Just a small improvement, this can be simplified by doing the check in the first if, instead of a nested if with a separate return.

```diff
diff --git a/wee_slack.py b/wee_slack.py
index fcbc293..d1a526b 100644
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -3732,14 +3732,11 @@ def line_event_cb(data, signal, hashtable):
     buffer_pointer = hashtable["_buffer"]
     line_timestamp = hashtable["_chat_line_date"]
     line_time_id = hashtable["_chat_line_date_printed"]
+    channel = EVENTROUTER.weechat_controller.buffers.get(buffer_pointer)
 
-    e = EVENTROUTER
-    if line_timestamp and line_time_id and buffer_pointer in e.weechat_controller.buffers:
+    if line_timestamp and line_time_id and isinstance(channel, SlackChannelCommon):
         ts = SlackTS("{}.{}".format(line_timestamp, line_time_id))
 
-        channel = e.weechat_controller.buffers[buffer_pointer]
-        if not isinstance(channel, SlackChannelCommon):
-            return w.WEECHAT_RC_OK
         message_hash = channel.hash_message(ts)
         if message_hash is None:
             return w.WEECHAT_RC_OK
```

wee_slack.py
```diff
@@ -3693,6 +3727,33 @@ def command_status(data, current_buffer, args):
...
         EVENTROUTER.receive(s)
 
+@utf8_decode
+def line_event_cb(data, signal, hashtable):
+    buffer_pointer = hashtable["_buffer"]
+    line_timestamp = hashtable["_chat_line_date"]
+    line_time_id = hashtable["_chat_line_date_printed"]
+
+    e = EVENTROUTER
+    if line_timestamp and line_time_id and buffer_pointer in e.weechat_controller.buffers:
```
trygveaa: I saw now that `line_timestamp` and `line_time_id` is `-1` if you select e.g. a date change line, and that is truthy. When can these values be falsy?

trygveaa:
@immae: No problem. Thanks for sticking by for all my comments! I think the comments I added now are my last ones. Just take a look and those, and we're good.

wee_slack.py
immae: :+1:

wee_slack.py
immae: They represent date.
time_id is actually a "misuse" of date_printed (cf doc here for recall: https://github.com/wee-slack/wee-slack/blob/master/wee_slack.py#L3147 ) , which is a time (according to the documentation of weechat, if you search for `date_printed` in https://weechat.org/files/doc/stable/weechat_plugin_api.en.html#hdata_line_data)

So this value for non-wee-slack related times should instead be a timestamp that is about equal to nowadays time.

So we could check that it is both positive and less than say 1 million (ierk)

However, in this case, we have a subsequent check that will check that the SlackTS is indeed in the existing message lists, so we need only to make sure that `int(the_value)` (called in the SlackTS initiator) is not failing, which should only be a problem for None.

trygveaa: Yeah, I know they represent dates. I just noticed that it's `-1` when not set, so I wondered if the if worked as you indented. But if it's just to handle `None` to prevent errors in `SlackTS`, then it makes sense. I don't really think it will ever be None though, but I'm not sure, so let's keep it.

trygveaa:
Thanks! This is really nice.

PR #630, opened 2018-09-14
Reply to existing thread with a command

janza:
This allows for replying in threads existing threads using command:

```
/thread b4d Hello thread!
```

Useful with the feature to show thread messages in the parent channel, see #616

### DIFF
--- a/README.md
+++ b/README.md
@@ -241,6 +241,11 @@ Open an existing thread as a channel. The argument is the thread identifier, whi
 /thread af8
 ```
 
+Reply to an existing thread. The second argument is the thread identifier, which is printed in square brackets with every threaded message in a channel:
+```
+/thread af8 Hello thread!
+```
+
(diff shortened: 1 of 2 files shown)


trygveaa:
I think it makes more sense to use the `/reply` command instead of the `/thread` command for this. I.e. change the `/reply` command to allow a thread hash in addition to a number.

yeled:
Why don't /thread or /reply tab complete?

janza:
I'll try to change it. It might be tricky distinguishing between hashes (which might be just numbers) and line numbers.

trygveaa:
We can probably assume no one is going to count 100 messages back, so everything with three characters or more can be assumed to be a hash.

trygveaa:
@yeled: `/reply` currently only accepts numbers, so it's nothing to complete. `/thread` could probably complete message hashes, it's just not implemented by anyone. Actually, I don't think we have completion for any other commands than `/slack` (and the options for that command are static).

janza:
Ah, cool! then I'll close this one.

Issue #523, opened 2018-03-10
Constant events=POLLIN -> CPU load


jan-vandenberg:
I'm running wee-slack with legacy tokens and everything runs fine except my weechat process uses 2-5% of my CPU _constantly_ . I have other (non-Slack) weechat instances on the same machine that don't have this problem, so I guess its related to the websocket? 

Because when I strace my process, this comes flying across the screen 100s of times:    

poll([{fd=5, events=POLLIN}, {fd=7, events=POLLIN}, {fd=0, events=POLLIN}, {fd=18, events=POLLIN}, {fd=10, events=POLLIN}, {fd=47, events=POLLIN}, {fd=15, events=POLLIN}], 7, 8) = 0 (Timeout)
fcntl(5, F_GETFD)                       = 0
fcntl(7, F_GETFD)                       = 0
fcntl(0, F_GETFD)                       = 0
fcntl(18, F_GETFD)                      = 0
fcntl(10, F_GETFD)                      = 0
fcntl(47, F_GETFD)                      = 0
fcntl(15, F_GETFD)                      = 0 

Any tips how I can debug this further?

tfheen:
I would check which of those fds, if any, are the websocket fd.  `lsof -p pid-of-weechat` should tell you. We just hook into weechat's hook_fd, so maybe that's doing polling?

tfheen:
Does https://err.no/p/2018-03-10-V1Vioqnb6ys/stdin make it better for you?  (Feel free to adjust the wanted_interval to something higher and see how that works.)

jan-vandenberg:
Yes! That patch indeed does seem to impact the CPU load/strain quite a bit!
I've tested some different wanted_interval numbers (up to 10000) and this even further reduces the CPU strain (and makes strace very quiet), however it also makes wee-slack unusable: messages/history don't appear. 
I've returned to the default wanted_interval values from the diff and will test this a little bit longer.
Thanks!

tfheen:
Numbers above 1000 probably won't make stuff any better, we have other timers that fire at 1s intervals so you'll hit those.  If 100 is ok and seems to work without side effects, I'll test that too and look at merging it.

jaywilliams:
This definitely lowers the overall CPU usage. I'm trying it on my slow netbook computer, and the CPU usage dropped from 25% when idling to 5% idling.

jaywilliams:
[...]

PR #631, opened 2018-09-23
Use an adaptive eventrouter timer

tfheen:
If there's more work, schedule ourselves more often, if there's less,
back off a bit to lessen the CPU load.

Fixes: #523


yardenac:
I've been using this patch for a few months. It certainly helps, although IME most performance problems are related to starting up slack connections, so a more sustainable solution would space those out, as discussed in https://github.com/wee-slack/wee-slack/issues/611

PR #634, opened 2018-10-01
Add support for `/slack upload`-ing remote URLs

apetresc:
At the moment, `/slack upload` only supports local filenames, which is a bit limiting, especailly for those of us who use weechat over SSH. This simple patch adds support for something like:

`/slack upload https://github.com/Microsoft/MS-DOS/raw/master/msdos-logo.png`

which results in:

The only weakness I can see with this implementation is that it does the _download_ on the main thread (it still uses `hook_process()` for the upload). I don't anticipate this to be an issue since the types of remote files people generally upload are images or code files which will download very quickly, but let me know if this is a blocker and I need to spawn a background process for that too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -217,6 +217,7 @@ Upload a file to the current slack buffer:
 ```
 /slack upload [file_path]
 ```
+`file_path` can be a URL, in which case wee-slack will automatically download it and attach it as if it had been a local file.
 
 Run a Slack slash command. Simply prepend `/slack slash` to what you'd type in the official clients.:
 ```
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -12,8 +12,10 @@ import json
 import pickle
 import sha
 import os
+import os.path
 import re
 import urllib
+import urlparse
 import sys
 import traceback
 import collections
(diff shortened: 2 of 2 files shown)


apetresc:
Sorry to nag, but it's been about a month with no feedback at all on the PR (except for the ❤️ , thanks!) Just wanted to make sure there wasn't anything I was missing.

Is the feature explicitly not in scope, or has there just not been enough time for a review yet?

trygveaa:
Sorry, I just haven't had time to review it yet.

The blocking download will freeze all of weechat, so I don't think that's okay to do.

Though, I'm not sure why you want to upload images which are already available on the web. Can't you just send a normal message with the url?

apetresc:
@trygveaa Ah okay, I will refactor to make a non-blocking download call.

Uploading the file triggers the in-line display that simply linking to an image/text file/etc doesn't seem to, which is the main reason you'd want to do this.

trygveaa:
> Uploading the file triggers the in-line display that simply linking to an image/text file/etc doesn't seem to, which is the main reason you'd want to do this.

As in, the image is shown directly in the chat in Slacks official clients? Doesn't that happen for you? It does for me, as long as the image hasn't been shared recently, and isn't very large.

I see that for text files it doesn't happen though, so for those I see your point.

trygveaa:
Closing because of lack of update from author (I consider making it non-blocking a requirement for merging this).

PR #638, opened 2018-10-13
Fix typing and completion errors in core buffer

ericdwang:
Typing more than 8 characters or hitting tab after a non-space character
in the core buffer causes errors because the buffer is a SlackTeam
object instead of a SlackChannel.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -819,7 +819,7 @@ def typing_notification_cb(signal, sig_type, data):
         if typing_timer + 4 < now:
             current_buffer = w.current_buffer()
             channel = EVENTROUTER.weechat_controller.buffers.get(current_buffer, None)
-            if channel and channel.type != "thread":
+            if channel and hasattr(channel, "type") and channel.type != "thread":
                 identifier = channel.identifier
                 request = {"type": "typing", "channel": identifier}
                 channel.team.send_to_websocket(request, expect_reply=False)
@@ -887,7 +887,7 @@ def nick_completion_cb(data, completion_item, current_buffer, completion):
     current_buffer = w.current_buffer()
     current_channel = EVENTROUTER.weechat_controller.buffers.get(current_buffer, None)
 
-    if current_channel is None or current_channel.members is None:
+    if current_channel is None or getattr(current_channel, "members", None) is None:
         return w.WEECHAT_RC_OK
     for m in current_channel.members:
         u = current_channel.team.users.get(m, None)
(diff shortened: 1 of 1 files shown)


ericdwang:
Build failed for an unrelated reason, created a separate PR for the fix: #639. Not sure why it only started failing recently though, Travis might've changed something on their side.

wee_slack.py
```diff
@@ -819,7 +819,7 @@ def typing_notification_cb(signal, sig_type, data):
         if typing_timer + 4 < now:
             current_buffer = w.current_buffer()
             channel = EVENTROUTER.weechat_controller.buffers.get(current_buffer, None)
-            if channel and channel.type != "thread":
+            if channel and hasattr(channel, "type") and channel.type != "thread":
```
trygveaa: I'm thinking we can add `type = 'team'` to `SlackTeam` instead of changing this check.

wee_slack.py
```diff
@@ -887,7 +887,7 @@ def nick_completion_cb(data, completion_item, current_buffer, completion):
     current_buffer = w.current_buffer()
     current_channel = EVENTROUTER.weechat_controller.buffers.get(current_buffer, None)
 
-    if current_channel is None or current_channel.members is None:
+    if current_channel is None or getattr(current_channel, "members", None) is None:
```
trygveaa: It would be nice to have the `members` property on `SlackTeam` as well, but we don't have a list of all members on the team, so it would be misleading. So maybe best to just have it like this.

ericdwang: `members` could be constructed from `users` (e.g. `self.members = [user.identifier for user in self.users]`), but I'm not sure how useful nick completion would be in the core buffer anyway.

wee_slack.py
ericdwang: Sounds good!

wee_slack.py
```diff
@@ -1063,6 +1063,7 @@ def __init__(self, eventrouter, token, websocket_url, team_info, nick, myidentif
         # This highlight step must happen after we have set related server
         self.set_highlight_words(kwargs.get('highlight_words', ""))
         self.load_emoji_completions()
+        self.type = "channel"
```
trygveaa: Any reason you set type to `channel` instead of `team`?

wee_slack.py
trygveaa: Ah, forgot about that. Having users in the team buffer is nice for opening a query with someone. Also, it removes the need to check if the attribute is present.

Though, we shouldn't copy the list, as we would have to keep the lists in sync. But you can add this in `SlackTeam`:

```
    @property
    def members(self):
        return self.users
```

ericdwang: Even better!

trygveaa:
Thanks!

PR #641, opened 2018-10-16
Fix loading history with deleted attachment

### LINKED ISSUE #640: Fail to load buffer history after a message with deleted attachment
When loading the history of a specific buffer, if there exists a message with a deleted attachment, the message containing it, and all following messages wll fail to load. This is repeatable in any slack based buffer as i have found, as long as the attachment is deleted from the slack side. I would be happy to look into this bug but I am very new to this project and am unsure where to start, any guidance would be helpful.

Also: once the offending message is deleted, the buffer loads as normal

janza:
Deleted files don't have `url_private` nor `title` attributes. Their `mode` is set to `tombstone`.
Closes #640

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2992,8 +2992,16 @@ def unwrap_attachments(message_json, text_before):
 
 
 def unwrap_files(message_json, text_before):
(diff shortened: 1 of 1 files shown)


trygveaa:
Thanks!

PR #642, opened 2018-10-16
Show thread broadcast messages in parent channel

### LINKED ISSUE #620: Thread messages sent to parent channel are not printed in channel
When you send a thread message in the official client, you have to option to send the message to the parent channel as well. In wee-slack, these messages are only printed in the thread, not in the parent channel.

#491 is related, which regards the ability to send such messages from wee-slack.

janza:
Closes #620

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2173,6 +2173,19 @@ class SlackMessage(object):
                     w.color(config.thread_suffix_color),
                     self.hash or self.ts,
                     len(self.submessages))
+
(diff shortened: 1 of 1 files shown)


trygveaa:
This doesn't seem to work properly for me. When the channel history is loaded and a broadcast message is received, it's only printed in the thread, not in the channel. When you reload wee-slack, or receive a message in a channel which doesn't have the history loaded, it's printed in the channel, but if you open the thread, it's not shown there.

trygveaa:
Closing because of lack of response from the author.

PR #644, opened 2018-10-19
Gray out muted channels in buflist

ericdwang:
Similar to how Slack grays out muted channels in their own UI.


trygveaa:
Thanks!

PR #645, opened 2018-10-21
Add option to download files locally

### LINKED ISSUE #562: Add downloading of files uploaded to slack
It would be really great if files uploaded to slack (where "uploaded a file: [link]" message appears) message appears could be saved without necessity of copying the links and pasting them to browser with authenticated session.

janza:
Useful when running wee-slack without an open session in a browser/app.

Might also be interesting to replace links to downloaded files with something like: `file://${files_download_location}${file_name}`

Closes #562

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2605,6 +2605,23 @@ def process_message(message_json, eventrouter, store=True, **kwargs):
             channel.store_message(message, team)
         dbg("NORMAL REPLY {}".format(message_json))
 
+    download_location = config.get_string('files_download_location')
+    if not download_location:
+        return
+
+    for f in message_json.get('files', []):
+        if f.get('mode', '') == 'tombstone':
+            continue
+
+        file_name = '{}_{}.{}'.format(f['id'], f['title'], f['filetype'])
+        weechat.hook_process_hashtable(
+            "url:" + f['url_private'],
+            {
+                'file_out': os.path.join(download_location, file_name),
+                'httpheader': 'Authorization: Bearer ' + team.token
+            },
+            config.slack_timeout, "", "")
+
 
 def subprocess_thread_message(message_json, eventrouter, channel, team):
     # print ("THREADED: " + str(message_json))
@@ -3898,6 +3915,9 @@ class PluginConfig(object):
         'external_user_suffix': Setting(
             default='*',
             desc='The suffix appended to nicks to indicate external users.'),
+        'files_download_location': Setting(
+            default='',
+            desc='If set, file attachments will be downloaded to this location.'),
         'group_name_prefix': Setting(
             default='&',
             desc='The prefix of buffer names for groups (private channels).'),


wee_slack.py
```diff
@@ -2605,6 +2605,23 @@ def process_message(message_json, eventrouter, store=True, **kwargs):
             channel.store_message(message, team)
         dbg("NORMAL REPLY {}".format(message_json))
 
+    download_location = config.get_string('files_download_location')
```
trygveaa: Can you place this (everything you're adding in this function) in a separate function?

wee_slack.py
```diff
@@ -2605,6 +2605,23 @@ def process_message(message_json, eventrouter, store=True, **kwargs):
...
+    if not download_location:
+        return
+
+    for f in message_json.get('files', []):
+        if f.get('mode', '') == 'tombstone':
```
trygveaa: You don't need the second argument to `.get` here.

wee_slack.py
```diff
@@ -2605,6 +2605,23 @@ def process_message(message_json, eventrouter, store=True, **kwargs):
...
+            continue
+
+        file_name = '{}_{}.{}'.format(f['id'], f['title'], f['filetype'])
```
trygveaa: The id is just a random string, not very useful to the user. I guess you include it in case there are two files with the same title, but I think at least we should place the title first.

The title may include the file type, so we should only append the file type if it does not.

wee_slack.py
trygveaa: The comment was only for the specific line I commented on, i.e. `f.get('mode', '')`.

wee_slack.py
trygveaa: If the title has a file extension or not depends on the upload. If you upload an existing image and don't change the title, it will have a file extension in my experience.

If the id goes after the title as I suggested, I guess it doesn't matter that much if we remove the file extension or not.

I do wonder if we should drop the id though. It would be much nicer to just use the title which will be equal to the original file name if it isn't changed. Other programs usually append a number at the end if the file name already exists, so we could do the same.

trygveaa:
> I'm having a look at continuing this. Regarding having the call to process_message() from handle_history() not ending up downloading files, where is the main call to process_message() that should indeed download files? I otherwise only see a call from process_reply()... but process_reply() is even not called from anywhere!? There is something crucial that I haven't understood :)

Yes, there is some magic in how those functions are called. They are called when we receive a message on the websocket, and which of them is called depends on the message. The code for this is in [`EventRouter.handle_next`](https://github.com/wee-slack/wee-slack/blob/5146b1f9166bb99344b923eef859d9bd7127737d/wee_slack.py#L488). Specifically, the functions starting with `process_` are called here:

https://github.com/wee-slack/wee-slack/blob/5146b1f9166bb99344b923eef859d9bd7127737d/wee_slack.py#L559-L560

> About downloading of files in (some) process_message() function, would the subtype matter? Is it fine to opportunistically get files from message_json`?

You don't have to check the subtype. It is fine just to do it if `files` exist in the message.

wee_slack.py
quite: I see several with the title `Image from iOS`. If we rewrite the URL displayed in weeslack to file:// (or perhaps better configurable), then I think dropping the ID would be fine. Otherwise, you'd be hard pressed to find what you want among files with such titles :)

trygveaa: Hm, right, that not that useful. I think it would be easier to find it with a number than with the id though. With a number you can just take the latest. Looking through lots of id's would be hard.

Rewriting the link if the option is enabled seems like a good idea.

PR #646, opened 2018-10-23
Add option to disable IRC smart filter for join/leave messages

ericdwang:
Since join/leave messages in Slack are rarer compared to IRC (e.g. going
offline doesn't leave the channel), it'd be nice to enable the smart
filter for IRC while still showing the messages in wee-slack.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -3205,6 +3205,8 @@ def tag(tagset, user=None, thread=False, muted=False):
         # catchall ?
         "default": {"irc_privmsg", "notify_message", "log1"},
     }
+    if not config.irc_smart_filter:
+        tagsets["joinleave"].remove("irc_smart_filter")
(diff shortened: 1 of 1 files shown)


trygveaa:
You can set the filter to only operate on irc buffers (e.g. `/filter add irc_smart irc.* irc_smart_filter *`), so this option is not necessary.

I'm actually planning to remove the `irc_smart_filter` tag completely from wee-slack, because we don't implement the "smartness" it's supposed to have. We just add it to all join/leave messages. However, I wanted to wait until we properly mark join/leave messages as read before changing it, because it might be annoying for people to suddenly see join/leave messages which aren't marked as read (even though you still will be able to filter them out of course, you will just have to define a new filter).

ericdwang:
Ah, I didn't know you could specify the filter like that. Good to know, thanks!

PR #647, opened 2018-10-25
Simplify format regexes

cmatheson:
This uses regexes word boundaries instead of home-rolled word
boundaries.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2814,13 +2814,13 @@ def process_emoji_changed(message_json, eventrouter, **kwargs):
 
 ###### New module/global methods
 def render_formatting(text):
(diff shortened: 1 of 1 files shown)


auscompgeek:
If the bold regex works, then I'm not sure it actually works as expected. I don't think we want to match something like `a*b*c`, unless Slack does as well.

The italics regex definitely doesn't work though.

auscompgeek:
If the bold regex works, then I'm not sure it actually works as expected. I don't think we want to match something like `a*b*c`, unless Slack does as well.

trygveaa:
Closing because of lack of response from the author to @auscompgeek comments.

Issue #570, opened 2018-05-01
nicks with parentheses in don't send correctly


dustymabe:
For example the name `SteelOw(l)` tab completes correctly but then sends as `@SteelOw(`. If I try to escape the other parenthesis I see `@SteelOw(l\)`. I don't know how to send that nick using weeslack and make it highlight for the person.

nstielau:
One possible implementation path would just be telling me to remove the parenthesis from my nick. :smile:

dustymabe:
> One possible implementation path would just be telling me to remove the parenthesis from my nick. smile

:) it is a bug though..

Also here is some more information about versions:

```
SCRIPT_NAME = "slack"
SCRIPT_AUTHOR = "Ryan Huber <rhuber@gmail.com>"
SCRIPT_VERSION = "2.0.0"
SCRIPT_LICENSE = "MIT"
SCRIPT_DESC = "Extends weechat for typing notification/search/etc on slack.com"
```

```
$ weechat --version
2.0.1

```

Issue #545, opened 2018-03-28
Bug with UTF-8 characters in names + @-notification


wohali:
Straight from my log:

```
11:34 <wohali> sorry that was for @Liê
11:34 <wohali> that is not rendering right here, I hope that notifies the right person. Must be a Unicode issue.
11:34 <jan> @wohali the name is `@Liên`
11:35 <jan> we only see `@Liê` from you
11:35 <wohali> I am typing Liên but it comes out as Liê
11:35 <wohali> trying @Liê again
11:35 <wohali> yeah, when I add the @ it breaks
11:36 <wohali> I'll file a bug
```

In the edit line, it shows as `@Liên` but as soon as I hit Enter, it transmits as `@Liê`.

tfheen:
Does ```diff --git a/wee_slack.py b/wee_slack.py
index 14d56f0..c4ee3eb 100644
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2664,7 +2664,7 @@ def linkify_text(message, team, channel):
         .replace('>', '&gt;')
         .split(' '))
     for item in enumerate(message):
-        targets = re.match('^\s*([@#])([\w.-]+[\w. -])(\W*)', item[1])
+        targets = re.match('^\s*([@#])([\w.-]+[\w. -])(\W*)', item[1], re.UNICODE)
         if targets and targets.groups()[0] == '@':
             named = targets.groups()
             if named[1] in ["group", "channel", "here"]:
```
make this better?

wohali:
Yes! That's the fix. Thank you!

sdelafond:
Is there any plan to merge this ?

PR #649, opened 2018-10-29
Fix for #570, also incorporates #545

pasja:
Please review the unicode changes, as I am not 100% sure about those.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2861,7 +2861,7 @@ def render(message_json, team, force=False):
         message_json["_rendered_text"] = text
         return text
 
-
+@utf8_decode
 def linkify_text(message, team, channel):
     # The get_username_map function is a bit heavy, but this whole
     # function is only called on message send..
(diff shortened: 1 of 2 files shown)


pasja:
Thanks for the thorough review, I really appreciated it!
I incorporated the suggested changes, with that the PR is much smaller.

Please check it and let me know if it still needs some love.

wee_slack.py
```diff
@@ -2880,7 +2880,7 @@ def linkify_text(message, team, channel):
         .replace('>', '&gt;')
         .split(' '))
     for item in enumerate(message):
-        targets = re.match('^\s*([@#])([\w.-]+[\w. -])(\W*)', item[1])
+        targets = re.match('^\s*([@#])([\w\(.-]+[\w\). -])(\W*)', item[1], re.UNICODE)
```
trygveaa: I tried to understand why this regex is written like this to figure out if your parenthesis are in the right places, but it doesn't make sense to me. Before your additions the two character classes (`[\w.-]` and `[\w. -]`) are equal except for a space. However, the message is split on spaces, so what is matched here will never contain spaces.

Either way, by placing the end parenthesis in the last class you define that nicks with parenthesis have to end with an end parenthesis. However, nicks in Slack are allowed to have parenthesis in the middle, so you should place it together with the start parenthesis. You can probably drop the last character class (`[\w. -]`), otherwise you will have to add both parenthesis in both classes.

pasja: Ok, I changed the regexp as you suggested and wrote a test for that case as well.

PR #650, opened 2018-10-29
Add linkarchive slack command to create a link to a given message

immae:
This PR makes a small addition thanks to the new "line message search" feature: it permits to paste in the input the full link to a given message, obtained either by its thread id or by clicking / cursoring it.

### DIFF
--- a/README.md
+++ b/README.md
@@ -223,6 +223,13 @@ Run a Slack slash command. Simply prepend `/slack slash` to what you'd type in t
 /slack slash /desiredcommand arg1 arg2 arg3
 ```
 
+Paste in the input the slack link to some message:
+```
+/slack linkarchive [$]message_id
+```
+This command is intended to be used with the mouse or cursor mode (see
+below section).
+
 To send a command as a normal message instead of performing the action, prefix it with a slash or a space, like so:
 ```
 //slack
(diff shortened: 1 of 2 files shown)


wee_slack.py
```diff
@@ -3651,6 +3651,16 @@ def command_openweb(data, current_buffer, args):
...
 
+@slack_buffer_required
+def command_linkarchive(data, current_buffer, args):
+    channel = EVENTROUTER.weechat_controller.buffers[current_buffer]
+    message_id = args.split(' ', 1)[1]
+    if message_id[0] == "$":
+        message_id = message_id[1:]
+    if isinstance(channel, SlackChannelCommon) and message_id in channel.hashed_messages:
+        ts = channel.hashed_messages[message_id].ts
+        url = "https://{}/archives/{}/p{}{:0>6}".format(channel.team.domain, channel.identifier, ts.majorstr(), ts.minorstr())
```
trygveaa: I'm trying to figure out if it's better to use the [chat.getPermalink](https://api.slack.com/methods/chat.getPermalink) method to create this link, rather than hard code it.

On the one side, we would be sure that the link is correct, even if it changes. And links to thread messages should have some extra parameters, which this includes.

On the other side, the link is probably not going to change, we can include the thread parameters ourselves, and it would be nice to having to do a request.

immae: The documentation says
> This method allows hundreds of requests per minute. Use it more or less as often as required, but please obey HTTP 429 Too Many Requests status codes and wait to retry.
So I guess we can use it as much as necessary and shouldn’t think too much about it?
But yes, that means having one call each time you want to generate a link (which should be okay, I guess?)

trygveaa: My concern isn't about rate limiting, it's about slow requests. Normally they will go fast enough, but what should we do if the request takes a long time for some reason? A blocking request will freeze all of weechat, so that's not acceptable. With an async request, it might look like nothing is happening, until the input field is suddenly populated later. Unless we show some loading, but ugh.

trygveaa: Blocking means that it will freeze all of weechat completely if the network hangs for some reason. Even if Slack always responds fast on that request, one might have some random network issues causing requests hanging. We can never allow that to happen, so a blocking is off the table.

Let's just go for hard coding it, like you have done. But let's add the query parameters described on that page for threads.

README.md
```diff
@@ -223,6 +223,13 @@ Run a Slack slash command. Simply prepend `/slack slash` to what you'd type in t
 /slack slash /desiredcommand arg1 arg2 arg3
 ```
 
+Paste in the input the slack link to some message:
+```
+/slack linkarchive [$]thread_id
```
trygveaa: Should probably write `message_id` instead, so it doesn't seem like this can only be used for thread messages.

wee_slack.py
```diff
@@ -3659,6 +3659,19 @@ def command_openweb(data, current_buffer, args):
...
+def command_linkarchive(data, current_buffer, args):
+    channel = EVENTROUTER.weechat_controller.buffers[current_buffer]
+    message_id = args.split(' ', 1)[1]
+    if message_id[0] == "$":
+        message_id = message_id[1:]
+    if isinstance(channel, SlackChannelCommon) and message_id in channel.hashed_messages:
+        message = channel.hashed_messages[message_id]
+        ts = message.ts
+        url = "https://{}/archives/{}/p{}{:0>6}".format(channel.team.domain, channel.identifier, ts.majorstr(), ts.minorstr())
+        if isinstance(channel, SlackThreadChannel):
```
trygveaa: You should rather check if the message is a thread message here, otherwise links for thread messages printed in the channel (with the thread_messages_in_channel option) won't get these parameters.

README.md
immae: Fixed, I also reordered alphabetically the `/key` lines

wee_slack.py
immae: Fixed too. I considered using the ThreadMessage and didn’t think about this possible case so I kept the ThreadChannel initially.
I’m impressed by all the subtelties you know about slack!

trygveaa: Hehe, thanks :)

trygveaa:
Thanks!

PR #651, opened 2018-10-29
Fix status call

ismaelbouyaf:
This fixes the status call in case the user has no emoji or no status text defined


trygveaa:
Thanks!

PR #652, opened 2018-10-29
Add open thread command with cursor

immae:
This adds an "open thread" command using the cursor.

It also adds support for `/thread $id` syntax


trygveaa:
If you do this on a message without a thread, a buffer for a non-existing thread will open. I see that this is a problem with the existing `/thread` command, and not really with these changes, but the problem is much more likely to occur with this addition.

trygveaa:
Hm, right, the buffer works even if the thread is created afterwards. I guess it's not a problem then, even though the thread doesn't technically exist yet.

trygveaa:
Thanks!

PR #653, opened 2018-10-31
Include threads when marking latest message as read

ericdwang:
This is the same as #549, which I closed originally to work on a more complicated solution that tackled some other thread-related things. With #616 I feel that solution isn't necessary anymore though, and this simple change actually pairs nicely with enabling `thread_messages_in_channel` (but it'll still fix channels not being marked as read regardless).


mkaito:
Oh nice, this has been annoying :+1:

trygveaa:
Thanks! And sorry for the late reply. I was also thinking about doing something more complex, but this is probably good enough. It's definitely better than the current behavior.

PR #654, opened 2018-10-31
Load `weemoji.json` with UTF-8

vojtechkral:
I had to do this to make `wee-slack` work on my machine. Not sure why it was needed.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -3826,7 +3826,7 @@ def create_slack_debug_buffer():
 def load_emoji():
     try:
         DIR = w.info_get("weechat_dir", "")
-        with open('{}/weemoji.json'.format(DIR), 'r') as ef:
+        with open('{}/weemoji.json'.format(DIR), 'r', encoding='utf8') as ef:
             return json.loads(ef.read())["emoji"]
     except Exception as e:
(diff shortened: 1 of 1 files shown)


auscompgeek:
I'm guessing your weechat is using Python 3 rather than Python 2.

This will break Python 2 compatibility, but a `from io import open` should fix that I think.

trygveaa:
@vojtechkral: Does wee-slack really work for you with this change? As @auscompgeek said, this won't work with python2, and I'm pretty sure the script doesn't work with python3 for various reasons.

Additionally, the `weemoji.json` file is ascii, so I don't really see how this change could help.

vojtechkral:
It now seems to work regardless of whether I have that line modified or not...
When I first installed wee-slack, I kept getting an error that `weemoji.json` cannot be read, followed by a  `UnicodeDecodeError` in the `dbg` function.
I have no idea what happened there...

trygveaa:
Ah, are you using a language with non-ascii characters so the error message you got about weemoji.json might have contained non-ascii characters? There may be a bug in the dbg function regarding that.

vojtechkral:
@trygveaa Yes, I am using a non-ascii language. If the error message comes from `perror()` or the like, it would probably contain non-ascii stuff...

PR #657, opened 2018-11-05
Add apostrophe to the regex with a corresponding test

### LINKED ISSUE #618: Cannot ping name with apostrophe
Hi.
When I try to ping someone who has an aspostrophe in his name like `@O'Connor`, it does not work, as if wee-slack is removing the characters after the apostrophe: it shows `@O'`.
Thanks to fix this issue.

pasja:
* fixes #618


trygveaa:
Thanks!

PR #660, opened 2018-11-08
Add an option to set color of (edited) suffix

er-1:
This option allows a user to set a custom color to the (edited) suffix on messages that have been edited.
The default color is set to '095', as it was the former choice.


trygveaa:
Thanks!

PR #665, opened 2019-01-15
Two fixes of topic handling

liskin:
Fixes:

* `wrong arguments for function "buffer_set"` when joining a channel
  from the web client

* resolve references (e.g. `<@JHDJFHKDW>`) in topic

I'm still not really happy with the code, seems to me that
using `update_from_message_json` to fill channel attributes is really
dangerous and we should avoid it. I'm not sure what will break if I just
drop the call. Does anyone here know or shall I just drop it and try
running that for a while?

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1461,6 +1461,7 @@ class SlackChannel(SlackChannelCommon):
             topic = self.topic['value']
             if topic == "":
                 topic = self.slack_purpose['value']
+            topic = unhtmlescape(unfurl_refs(topic, ignore_alt_text=False))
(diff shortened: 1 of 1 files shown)


trygveaa:
Ah, I didn't know/realize `update_from_message_json` sets the topic. As you say, it's a bad function.

Yes, we should remove it. But we should check what it sets first, so we know if we have to replace it with something. I see that it's only called from `process_channel_joined`, so we should take a look at what's sent into it there, and see if we have to set any of those properties specifically.

trygveaa:
Thanks!

PR #666, opened 2019-01-17
Work further on download files, #645

quite:


quite:
This adds to #645, solving the requested changes and discussion.

Rewriting the URL is not done, should open an issue for that after closing #562

wee_slack.py
```diff
@@ -5,6 +5,7 @@
 from collections import OrderedDict
 from functools import wraps
 from itertools import islice
+from itertools import count
```
trygveaa: These two imports can be combined on one line.

wee_slack.py
```diff
@@ -2644,6 +2647,39 @@ def process_message(message_json, eventrouter, store=True, **kwargs):
             channel.store_message(message, team)
         dbg("NORMAL REPLY {}".format(message_json))
 
+    if download:
+        download_files(message_json, **kwargs)
+
+
+def download_files(message_json, **kwargs):
+    team = kwargs["team"]
+    download_location = config.get_string('files_download_location')
```
trygveaa: You should use `config.files_download_location` instead.

Note that you have to add `get_files_download_location = get_string` to `PluginConfig` to make that work.

wee_slack.py
```diff
@@ -4008,6 +4044,9 @@ class PluginConfig(object):
         'external_user_suffix': Setting(
             default='*',
             desc='The suffix appended to nicks to indicate external users.'),
+        'files_download_location': Setting(
+            default='',
+            desc='If set, file attachments will be downloaded to this location.'),
```
trygveaa: I think we should say "will be automatically downloaded", to make it clear that they are downloaded without requiring any action.

wee_slack.py
```diff
@@ -2644,6 +2647,39 @@ def process_message(message_json, eventrouter, store=True, **kwargs):
...
+        yield path
+        main, ext = os.path.splitext(path)
+        for i in count(start=1, step=1):
+            yield main + "-{}".format(i) + ext
+
+    for f in message_json.get('files', []):
+        if f.get('mode') == 'tombstone':
+            continue
+
+        filename = '{}_{}.{}'.format(team.preferred_name, f['title'], f['filetype'])
```
trygveaa: I still think we shouldn't append the file type if it's already there, so we don't end up with `.png.png`. E.g.:

```
        filetype = '' if f['title'].endswith(f['filetype']) else '.' + f['filetype']
```

wee_slack.py
```diff
@@ -2644,6 +2647,39 @@ def process_message(message_json, eventrouter, store=True, **kwargs):
...
+        for fileout in fileout_iter(os.path.join(download_location, filename)):
+            if os.path.isfile(fileout):
+                continue
+            weechat.hook_process_hashtable(
+                "url:" + f['url_private'],
+                {
+                    'file_out': fileout,
+                    'httpheader': 'Authorization: Bearer ' + team.token
+                },
+                config.slack_timeout, "", "")
```
trygveaa: For me, weechat isn't creating the directory if it doesn't exist, so it seems like we have to create the directory ourselves.

wee_slack.py
quite: Done.

quite:
I've resolved your comments. No sure if I did it the right way, regarding commiting, pushing and resolving. I'm currently wired to Gerrit, where you always maintain your whole patch set in the shape you want it (force-pushing, except change history *is* tracked!).

wee_slack.py
```diff
@@ -2617,6 +2619,45 @@ def process_message(message_json, eventrouter, store=True, **kwargs):
...
+def download_files(message_json, **kwargs):
+    team = kwargs["team"]
+    download_location = config.files_download_location
+    if not download_location:
+        return
+    if not os.path.exists(download_location):
+        try:
+            os.makedirs(download_location)
+        except:
+            dbg("failed to makedirs({})".format(download_location))
```
trygveaa: We should print this to the core buffer (`w.prnt('', 'ERROR: ...')`) instead so it's more apparent that something is failing, and preferably with a bit more descriptive message, e.g. "Failed to create directory at \`files_download_location\`". It should also include the error message from the exception.

wee_slack.py
```diff
@@ -2617,6 +2619,45 @@ def process_message(message_json, eventrouter, store=True, **kwargs):
...
+
+    def fileout_iter(path):
+        yield path
+        main, ext = os.path.splitext(path)
+        for i in count(start=1, step=1):
```
trygveaa: `step` is not necessary here.

trygveaa:
Thanks!

PR #670, opened 2019-01-31
[WIP] render snippets

dbarrosop:
This prints the preview of the snippet alongside its URL.

It's not extremely useful because the preview is only a few lines long. Ideally this would download the entire snippet and render it but I couldn't figure out in a reasonable amount of time how to do it. Some guidance/help here would be appreciated.


trygveaa:
Or you can reuse parts of the `download_files` method to download the snippet content, since shared files and snippets are accessed the same way.

https://github.com/wee-slack/wee-slack/blob/55dfe7442183912339d5e08740a133fd333268b2/wee_slack.py#L2651-L2685

PR #674, opened 2019-02-11
Fix /slack linkarchive command because of an API change

pasja:
* for details please see: https://api.slack.com/methods/chat.getPermalink


trygveaa:
Oops, this was actually caused by a mistake on my part, not an API change. Thanks for fixing it!

PR #675, opened 2019-02-14
Fix 'slackbot' name change to 'Slackbot' breaking never_away

V13Axel:
The 'slackbot' channel has been renamed to 'Slackbot'. 

The channel name change breaks `slack_never_away_cb`, and this PR fixes it.


trygveaa:
Thanks!

PR #677, opened 2019-02-21
Add ability to open last thread in channel without specifying message id

tomasv:
Hey 👋 

Great plugin!

I had this idea and managed to hack something together: most of the time only the last thread in the channel is the most intriguing one and in those cases there's no need to type out a random id to access it.

This adds the ability to just type:
```
/thread
```
and the last thread in the current channel should be opened if found.


tomasv:
@trygveaa updated.

PR #680, opened 2019-03-09
Add @user-groups support

Aerex:
Add @user-groups with tab-completion. @user-groups will be unfurl into format `<!subteam^{ID}|handle> message` before sending message via `linktext` method

* Implementation includes unit test


torhve:
I like this, but shouldn't this PR also add support for notification to usergroups that you are in?

Aerex:
@torhve That is a good point. I will start add that into the PR.

wee_slack.py
```diff
@@ -917,6 +916,23 @@ def emoji_completion_cb(data, completion_item, current_buffer, completion):
         w.hook_completion_list_add(completion, ":" + e + ":", 0, w.WEECHAT_LIST_POS_SORT)
     return w.WEECHAT_RC_OK
 
+@utf8_decode
+def usergroups_completion_cb(data, completion_item, current_buffer, completion):
+    """
+    Adds all :-prefixed usergroups to completion list
```
trygveaa: `:-prefixed` isn't right for usergroups.

wee_slack.py
```diff
@@ -917,6 +916,23 @@ def emoji_completion_cb(data, completion_item, current_buffer, completion):
...
+    """
+
+    current_channel = EVENTROUTER.weechat_controller.buffers.get(current_buffer, None)
+
+    if current_channel is None:
+        return w.WEECHAT_RC_OK
+    subteam_ids = current_channel.team.subteams.viewkeys()
+    for subteam_id in subteam_ids:
+        subteam = current_channel.team.subteams.get(subteam_id, None)
```
trygveaa: To loop through a dict getting the value you can use:
```
for subteam in current_channel.team.subteams.values():
```

wee_slack.py
```diff
@@ -1021,14 +1037,34 @@ def should_try(self):
...
+       self.team_id = originating_team_id 
+
+   def __repr__(self):
+       return "Name:{} Identifier:{}".format(self.name, self.identifier)
+
+   def __eq__(self, compare_str):
+       if compare_str == self.subteam_id :
+           return True
+       else:
+           return False
```
trygveaa: Just do:
```
return compare_str == self.subteam_id
```

wee_slack.py
```diff
@@ -1062,6 +1099,14 @@ def __init__(self, eventrouter, token, websocket_url, team_info, nick, myidentif
         for c in self.channels.keys():
             channels[c].set_related_server(self)
             channels[c].check_should_open()
+
+        # Build mapper between subteam handle and their identifiers 
+        self.subteam_handles_to_id = {}
+        for s in self.subteams.keys():
+            subteam = self.subteams.get(s, None)
+            if subteam:
+                self.subteam_handles_to_id[subteam.handle] = subteam.identifier
```
trygveaa: This should be a method, like `get_username_map`, in case `self.subteams` changes.

wee_slack.py
```diff
@@ -917,6 +916,23 @@ def emoji_completion_cb(data, completion_item, current_buffer, completion):
...
+
+    current_channel = EVENTROUTER.weechat_controller.buffers.get(current_buffer, None)
+
+    if current_channel is None:
+        return w.WEECHAT_RC_OK
+    subteam_ids = current_channel.team.subteams.viewkeys()
+    for subteam_id in subteam_ids:
+        subteam = current_channel.team.subteams.get(subteam_id, None)
+        if subteam:
+            w.hook_completion_list_add(completion, "@" + subteam.handle, 0, w.WEECHAT_LIST_POS_SORT)
```
trygveaa: Even though user groups are not nicks, I think we should set `nick_completion` to 1, because they are used like nicks. This will cause a colon to be inserted after the user group if it is completed on the start of the line.

wee_slack.py
Aerex: typo error I will fixed that.

wee_slack.py
Aerex: Yeah not sure why I did that. I will fix that

trygveaa: You mean remove if/else from that too? Sure, go ahead.

Though, when I think about it, these methods don't really make sense. `__eq__` should check if it equals another instance of the same class, not compare with a string. I wonder if they are used anywhere.

Aerex:
@trygveaa I have added [subteam_created](https://api.slack.com/events/subteam_created), [subteam_updated](https://api.slack.com/events/subteam_updated), [subteam_self_added](https://api.slack.com/events/subteam_self_added), and [subteam_self_removed](https://api.slack.com/events/subteam_self_removed) event as process handlers. The [subteam_members_changed](https://api.slack.com/events/subteam_members_changed) event didn't seem useful to me compared to the others but if you feel like we should add let me know. 

Also, I have included a usergroups command that will print out the users for a given usergroup given the handler as an argument. Usage is described in the documentation.  

I was going to add some test for handling the `usersgroups.users.list` but it would be more involved that I thought it would be worth. You don't seem to have any tests for the other request handlers. 

Let know if you have any comments, questions, or concerns. Thanks

wee_slack.py
```diff
@@ -2957,6 +3100,8 @@ def linkify_text(message, team, channel):
             named = targets.groups()
             if named[1] in ["group", "channel", "here"]:
                 message[item[0]] = "<!{}>".format(named[1])
+            elif named[1] in list(channel.team.subteam_handles_to_id.viewkeys()):
+                message[item[0]] = "<!subteam^{}|{}>".format(channel.team.subteam_handles_to_id[named[1]], named[1])
```
trygveaa: If you write e.g. colon after the group, it will be lost because `named[2]` is not included. I see this was an issue with group/channel/here too, I've fixed that in master now.

wee_slack.py
```diff
@@ -3488,6 +3635,34 @@ def command_users(data, current_buffer, args):
...
+        subteam = team.subteams[subteam_identifer]
+        s = SlackRequest(team.token, "usergroups.users.list", { "usergroup": subteam.identifier }, team_hash=team.team_hash)
+        e.receive(s)
+    elif not args:
+        team.buffer_prnt("Usergroups:")
+        for subteam in team.subteams.values():
+            team.buffer_prnt("    {:<25}(@{})".format(subteam.name, subteam.handle))
+    else:
+        w.prnt('', 'ERROR: Unknown usergroup handler: {}'.format(args))
+        w.prnt('', 'Usage: /slack usergroups <handler>')
```
trygveaa: Don't print usage info when the command syntax was correct. The first message is sufficient.

wee_slack.py
```diff
@@ -3241,6 +3386,8 @@ def tag(tagset, user=None, thread=False, muted=False):
         # messages in the team/server buffer, e.g. "new channel created"
         "team_info": {"no_highlight", "log3"},
         "team_message": {"irc_privmsg", "notify_message", "log1"},
+        # messages from a subgroup/usergroup 
+        "usergroup":{"irc_privmsg", "notify_message", "log1"},
```
trygveaa: I can't see that you use this tag anywhere?

docs/Commands.md
```diff
@@ -173,3 +173,19 @@ Uploads a file to the current buffer.
...
+List the usergroups in the current team.
+
+```
+/slack usergroups <handler>
+```
+
+List the users in a usergroup using handler (omit @)
+
+
+
```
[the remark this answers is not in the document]
Aerex: I did run the script however I did not notice that it was added when I staged my files.

wee_slack.py
```diff
@@ -2639,6 +2725,63 @@ def process_message(message_json, eventrouter, store=True, download=True, **kwar
...
+        team.subteam_handles_to_id[new_subteam_info['handle']] = new_subteam_info.get('id')
+        template = '{name} has updated its handle to @{handle} in team {team}'
+        message = template.format(name=current_subteam_info.name, handle=new_subteam_info['handle'],
+                team=team.preferred_name)
+        messages.append(message)
+
+    team.subteams[new_subteam_info.get('id')] = SlackSubteam(team.identifier, **new_subteam_info)
+
+    for message in messages:
+        team.buffer_prnt(message, message=True)
```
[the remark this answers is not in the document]
Aerex: Fair enough. I wasn't certain myself so I added as much as I could just in case. I remove of the usergroups notification
events except when the usergroup handle changes. A user can enable or disable that option in the configuration. I think
it useful if the handle changes so you can keep using without interruptions.

Aerex:
@trygveaa Thanks for the review. I added your suggestions to my pull request. I added some two comments from your
responses. I will reiterate one comment here. In regards to the notification, I removed all of the notification
usergroup except for when the handle has changed. You can enable or disable the notification in the new setting 
**notify_usergroup_updated**. Let me know if you have any questions or concerns. Thanks

wee_slack.py
```diff
@@ -2973,6 +3046,8 @@ def linkify_text(message, team):
             named = targets.groups()
             if named[1] in ["group", "channel", "here"]:
                 message[item[0]] = "<!{}>{}".format(named[1], named[2])
+            elif named[1] in list(usergroups.viewkeys()):
+                message[item[0]] = "<!subteam^{}|{}>".format(usergroups[named[1]], named[1])
```
trygveaa: Did you see my comment about including `named[2]` here?

> If you write e.g. colon after the group, it will be lost because named[2] is not included. I see this was an issue with group/channel/here too, I've fixed that in master now.

wee_slack.py
```diff
@@ -3026,6 +3102,12 @@ def unfurl_ref(ref, ignore_alt_text, auto_link_display):
                 display_text = "#{}".format(ref.split('|')[1])
             elif id.startswith("@U"):
                 display_text = ref.split('|')[1]
+            elif id.startswith("!subteam"):
+                if ref.split('|')[1].startswith('@'):
+                    handle = ref.split('|')[1][1:]
+                else:
+                    handle = ref.split('|')[1]
```
trygveaa: Hm, so the handle in `<!subteam^U2147483697|@group>` doesn't include `@` for you?

wee_slack.py
```diff
@@ -927,6 +930,21 @@ def emoji_completion_cb(data, completion_item, current_buffer, completion):
...
+    """
+    Adds all @-prefixed usergroups to completion list
+    """
+
+    current_channel = EVENTROUTER.weechat_controller.buffers.get(current_buffer, None)
+
+    if current_channel is None:
+        return w.WEECHAT_RC_OK
+    for subteam in current_channel.team.subteams.values():
+        if subteam:
```
trygveaa: What is the reason for this if?

wee_slack.py
```diff
@@ -2649,6 +2699,28 @@ def process_message(message_json, eventrouter, store=True, download=True, **kwar
...
+    team = kwargs['team']
+    subteam_json_info = subteam_json['subteam']
+    subteam = SlackSubteam(team.identifier, **subteam_json_info)
+    team.subteams[subteam_json_info['id']] = subteam
+
+def process_subteam_updated(subteam_json, eventrouter, **kwargs):
+    team = kwargs['team']
+    usergroups = team.generate_usergroup_map()
+    new_subteam_info = subteam_json['subteam']
+    messages = []
```
trygveaa: This isn't used anymore.

wee_slack.py
```diff
@@ -2649,6 +2699,28 @@ def process_message(message_json, eventrouter, store=True, download=True, **kwar
...
+
+    current_subteam_info = team.subteams[new_subteam_info.get('id')]
+
+    if config.notify_usergroup_handle_updated and current_subteam_info.handle != new_subteam_info['handle']:
+        usergroups[new_subteam_info['handle']] = new_subteam_info.get('id')
+        template = '{name} has updated its handle to @{handle} in team {team}'
```
trygveaa: Since we don't use the user group name in wee-slack anywhere, I think it would be more convenient to write `User group @{old_handle} has ...`.

_pytest/test_processsubteamupdated.py
```diff
@@ -0,0 +1,30 @@
...
+
+    t = eventrouter.teams.keys()[0]
+
+    assert len(eventrouter.teams[t].subteams) == 1
+
+    socket = mock_websocket
+    eventrouter.teams[t].ws = socket
+    datafiles = glob.glob("_pytest/data/websocket/1483975206.59-subteam_updated.json")
+
+    with patch('wee_slack.SlackTeam.buffer_prnt') as fake_buffer_prnt:
```
trygveaa: Using patch here isn't necessary anymore.

wee_slack.py
Aerex: In a channel I found some of the usergroups printing the `@` symbol twice like `@@dev`. I have only seen this in old
messages. These two lines account for a possible extra `@` sign and removes them. Maybe it just me.

wee_slack.py
Aerex: I did but I must overlooked something when I was testing.

Aerex:
@trygveaa I added a comment in regards to this comment 
> Hm, so the handle in <!subteam^U2147483697|@group> doesn't include @ for you?

wee_slack.py
trygveaa: Ah, the web-client includes @ in the alt-text when creating a message, while wee-slack does not. I changed linkify_text to include @ in wee-slack as well.

trygveaa:
Thanks!

PR #683, opened 2019-03-19
Upload file to thread instead of channel when in thread buffer

### LINKED ISSUE #672: Uploading to a thread uploads to the channel instead
I opened a thread with `/slack thread [id]`, then from the thread buffer I issued `/slack upload /path/to/file.png`. The file uploaded, but it ended up in the channel containing the thread, rather than the thread itself.

ericdwang:
Currently files are always uploaded to the main channel, even when
`/slack upload` is run in a thread buffer. This happens because the
`thread_ts` field isn't included in the upload request.

Fixes #672


trygveaa:
Thanks!

PR #684, opened 2019-03-22
Correctly handle group join/leave events

auscompgeek:
This ensures that changes to private channel membership are rendered correctly and update the user list.


trygveaa:
Thanks!

PR #689, opened 2019-05-16
Fix install instructions

yanshiyason:
I couldn't get the plugin to autoload, and had to make these changes for it to work.
I am on OSX, maybe behavior is different on Linux..

### DIFF
--- a/README.md
+++ b/README.md
@@ -63,7 +63,9 @@ pkg_add py-websocket-client weechat-python
 ```
 cd ~/.weechat/python
 wget https://raw.githubusercontent.com/wee-slack/wee-slack/master/wee_slack.py
-ln -s ../wee_slack.py autoload
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -63,7 +63,9 @@ pkg_add py-websocket-client weechat-python
 ```
 cd ~/.weechat/python
 wget https://raw.githubusercontent.com/wee-slack/wee-slack/master/wee_slack.py
-ln -s ../wee_slack.py autoload
+
+cd ~/.weechat/python/autoload
```
auscompgeek: Why not just `cd autoload`?

trygveaa:
Could the problem have been that the autoload directory didn't exist? I asked another mac user, and the instructions worked for him as long as the directory exists.

I've changed the instructions to include a command to create the directories. Can you test the new instructions as see if they work for you?

PR #690, opened 2019-05-26
Allow %h (weechat home) replacement in download_location

quite:


docs/Options.md
```diff
@@ -92,7 +92,7 @@ slack` after changing it to take effect.
 
 **Default:** ``
 
-**Description:** If set, file attachments will be automatically downloaded to this location.
+**Description:** If set, file attachments will be automatically downloaded to this location. "%h" will be replaced by WeeChat home, "~/.weechat" by default.
```
trygveaa: This file is generated, so you need to change the description in [wee_slack.py](https://github.com/wee-slack/wee-slack/blob/7bff4f2856a2cc8126fa8d2fecbd07e0422f2fb4/wee_slack.py#L4344-L4347), and run `./generate_docs.py` instead of changing this file manually.

wee_slack.py
```diff
@@ -2795,6 +2795,8 @@ def download_files(message_json, **kwargs):
     download_location = config.files_download_location
     if not download_location:
         return
+    download_location = download_location.replace("%h", weechat.info_get("weechat_dir", ""))
```
trygveaa: Use the weechat wrapper `w` instead of `weechat`.

PR #691, opened 2019-05-27
Fix /topic command output

luk1337:
* This change makes sure the /topic command outputs
  'Topic for #random is ""' instead of
  'Topic for #random is "{'value': ''}"'.


trygveaa:
Thanks!

PR #700, opened 2019-06-12
Expose more information in /whois

auscompgeek:
Also hide any unset fields.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2310,6 +2310,7 @@ class SlackUser(object):
             setattr(self, key, value)
 
         self.name = nick_from_profile(self.profile, kwargs["name"])
+        self.username = kwargs["name"]
         self.update_color()
 
     def __repr__(self):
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -3604,14 +3605,25 @@ def whois_command_cb(data, current_buffer, command):
...
-        team.buffer_prnt("[{}]: Phone: {}".format(user, u.profile.get('phone', '')))
+
+        team.buffer_prnt("[{}]: username: {}".format(user, u.username))
+        team.buffer_prnt("[{}]: id: {}".format(user, u.identifier))
+
+        print_profile('title')
+        print_profile('email')
+        print_profile('phone')
+        print_profile('skype')
+        print_profile('fields')
```
trygveaa: What is fields?

auscompgeek: I couldn't remember what this was, so I had to go back through the Slack API docs...

From [users.profile.get](https://api.slack.com/methods/users.profile.get):

> For a description of the `fields` key, see the users.profile.set method.

...

[users.profile.set](https://api.slack.com/methods/users.profile.set) has a description of `fields`, but it's quite lengthy so I won't paste it here.

Looking again though, I'm not even sure we get the custom profile fields in the first place.

trygveaa: We get the profile with `rtm.start`, so that's why we never have to call `users.profile.get`.

Okay, after reading the docs, I see that this is custom fields each team can add. By default it only includes skype (which is the same as the skype key directly under profile). The fields are defined with an id and a value, e.g. if i add "some skype handle" as skype in my profile, fields for my profile will be this:
```
        "fields": {
          "Xf0D7H375J": {
            "value": "some skype handle",
            "alt": ""
          }
        },
```

To get what the id represent, we have to call [team.profile.get](https://api.slack.com/methods/team.profile.get).

If we want to show these fields in whois, we should do it properly, i.e. look up the ids and print them nicely. Otherwise, we should omit fields.

auscompgeek: I concur. I'll hide the fields for now and file a followup issue.

wee_slack.py
```diff
@@ -3604,14 +3605,25 @@ def whois_command_cb(data, current_buffer, command):
...
-        team.buffer_prnt("[{}]: Email: {}".format(user, u.profile.get('email', '')))
-        team.buffer_prnt("[{}]: Phone: {}".format(user, u.profile.get('phone', '')))
+
+        team.buffer_prnt("[{}]: username: {}".format(user, u.username))
+        team.buffer_prnt("[{}]: id: {}".format(user, u.identifier))
+
+        print_profile('title')
+        print_profile('email')
+        print_profile('phone')
+        # TODO: display other custom fields
```
trygveaa: I don't think I want both a TODO and an issue for missing features. It should suffice with the issue.

PR #702, opened 2019-06-24
Add logic to determine which users to subscribe to for presence

ericdwang:
Recently in one of my Slack teams I noticed my own presence was never getting updated! It always looked like I was away in all channels, and my nick color also wasn't `weechat.color.chat_nick_self`. Looking into it, I found my user id wasn't being included in the `presence_sub` request. That seemed odd to me, and I found some users' presence in other channels weren't getting updated as well. So I added this logic to determine users instead of using the order from `rtm.start`:

1. Current user
2. Users in DMs
3. Users in MPDMs
4. Users in private groups
5. Users in public channels

That's what made sense to me, but I'm open to changing it!

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -8,7 +8,7 @@ from __future__ import print_function, unicode_literals
 from collections import OrderedDict
 from functools import wraps
 from io import StringIO
-from itertools import islice, count
+from itertools import islice, count, chain
 
 import errno
 import textwrap
@@ -62,6 +62,7 @@ SCRIPT_DESC = "Extends weechat for typing notification/search/etc on slack.com"
 
 BACKLOG_SIZE = 200
 SCROLLBACK_SIZE = 500
+PRESENCE_SUBSCRIBE_SIZE = 750
 
 RECORD_DIR = "/tmp/weeslack-debug"
 
@@ -1422,11 +1423,37 @@ class SlackTeam(object):
                 c.update_nicklist(user.id)
 
     def subscribe_users_presence(self):
-        # FIXME: There is a limitation in the API to the size of the
-        # json we can send.
-        # We should try to be smarter to fetch the users whom we want to
-        # subscribe to.
-        users = list(self.users.keys())[0:750]
+        """
+        Subscribe to the presence of users in this order (restricted by the
+        API JSON size limit):
+
+        1. Current user
+        2. Users in DMs
+        3. Users in MPDMs
+        4. Users in private groups
+        5. Users in public channels
+        """
+        dm_users, mpdm_users, group_users, channel_users = set(), set(), set(), set()
+        for channel in self.channels.values():
+            if getattr(channel, "is_open", False):
+                if channel.type == "im":
+                    dm_users.add(channel.user)
+                elif channel.type == "mpim":
+                    mpdm_users.update(channel.members)
+                elif channel.type == "group":
+                    group_users.update(channel.members)
+            elif getattr(channel, "is_member", False):
+                if channel.type == "channel":
+                    channel_users.update(channel.members)
+
+        users, all_users = [self.myidentifier], {self.myidentifier}
+        for user in chain(dm_users, mpdm_users, group_users, channel_users):
+            if len(users) >= PRESENCE_SUBSCRIBE_SIZE:
+                break
+            if user not in all_users:
+                users.append(user)
+                all_users.add(user)
+
         self.send_to_websocket({
             "type": "presence_sub",
             "ids": users,


wee_slack.py
```diff
@@ -1422,11 +1423,37 @@ def update_member_presence(self, user, presence):
...
+            elif getattr(channel, "is_member", False):
+                if channel.type == "channel":
+                    channel_users.update(channel.members)
+
+        users, all_users = [self.myidentifier], {self.myidentifier}
+        for user in chain(dm_users, mpdm_users, group_users, channel_users):
+            if len(users) >= PRESENCE_SUBSCRIBE_SIZE:
+                break
+            if user not in all_users:
+                users.append(user)
```
auscompgeek: Instead of building a set and a list at the same time, why don't we just convert the set to a list at the end?

wee_slack.py
```diff
@@ -1422,11 +1423,37 @@ def update_member_presence(self, user, presence):
...
+        for channel in self.channels.values():
+            if getattr(channel, "is_open", False):
+                if channel.type == "im":
+                    dm_users.add(channel.user)
+                elif channel.type == "mpim":
+                    mpdm_users.update(channel.members)
+                elif channel.type == "group":
+                    group_users.update(channel.members)
+            elif getattr(channel, "is_member", False):
+                if channel.type == "channel":
```
auscompgeek: Is there a way to distinguish the mandatory channel (the one initially called #general)? Seems prudent to skip that, otherwise we just end up building a set of every user.

trygveaa: Yeah, with the `is_general` attribute.

trygveaa: We want to include the members from this channel as long at it's fewer than 750 though.

auscompgeek: If the number of members in #general are < 750, then barring any shared channels and guests, it stands to reason that the entire workspace would have <= 750 users as well, so we could just subscribe to all users without having to walk.

trygveaa: Sure, but say that the entire workspace has 1000 users, but the total of all channels and dms except for the general channel is 500. Then we want to include some users from that channel.

The best might be to take members from all other channels first, and then from the general channel last. I.e. exclude the general channel from `channel_users`, but include it last in the chain.

auscompgeek: Perhaps. But at the same time, it seems like "presence won't be available for those in #general that you don't otherwise share channels with" is simpler to understand than "presence will be available for an arbitrary random selection of users in #general, but you can't tell who".

I dunno.

auscompgeek: I would probably prefer special casing smaller workspaces anyway - I've seem teams where there's a _very_ disproportionate ratio of channels to users.

wee_slack.py
ericdwang: In an earlier version of the code this was creating a whole list of all users and slicing at the end so the order mattered then. But good catch, that's not needed now!

wee_slack.py
ericdwang: I think having a special case for <= 750 users makes sense, no need for all this work then.

Also I didn't know #general was mandatory! In that case it also makes sense to look at it last apart from the other channels.

ericdwang: Re-reading the comments: I think I may actually prefer just excluding #general completely. In my workspace with > 750 users, I only share channels with 177 of them! And since #general is limited to announcements from certain users I rarely look at it and the presences of everyone there.

trygveaa: Yeah, okay. You're right that having presence for just a random part of the users in general would be a bit weird, so I agree that subscribing to all users when there are less than 750 and skipping general if there are more is good.

wee_slack.py
```diff
@@ -1422,16 +1423,49 @@ def update_member_presence(self, user, presence):
...
+            if getattr(channel, "is_open", False):
+                if channel.type == "im":
+                    dm_users.add(channel.user)
+                elif channel.type == "mpim":
+                    mpdm_users.update(channel.members)
+                elif channel.type == "group":
+                    group_users.update(channel.members)
+            elif getattr(channel, "is_member", False):
+                if channel.type == "channel" and not channel.is_general:
+                    channel_users.update(channel.members)
```
trygveaa: This can be shortened to:

```
        users = defaultdict(set)
        for channel in self.channels.values():
            if (getattr(channel, "is_open", False) or getattr(channel, "is_member", False)
                 and not getattr(channel, "is_general", False)):
                users[channel.type].update(channel.get_members())
```

Then use `users['im']` instead of `dm_users` etc. `defaultdict` comes from `collections`.

wee_slack.py
```diff
@@ -1422,16 +1423,49 @@ def update_member_presence(self, user, presence):
...
+
+        return list(islice(chain(
+            {self.myidentifier},
+            dm_users,
+            mpdm_users,
+            group_users,
+            channel_users,
+            self.users.keys(),
+        ), PRESENCE_SUBSCRIBE_SIZE))
```
trygveaa: This doesn't take into account that the same user could be in multiple mpdms/channels, so some users may be repeated.

wee_slack.py
ericdwang: Nice, though I think it'd be fine to just define a normal dict with channel types because they're known, e.g. `users = dict.fromkeys(("im", "mpim", "group", "channel", "shared"), set())`

wee_slack.py
ericdwang: Tried to get too clever with this, woops.

wee_slack.py
```diff
@@ -1422,16 +1423,46 @@ def update_member_presence(self, user, presence):
...
+        5. Users in public channels (excluding the "general" channel)
+        """
+        users = {
+            channel_type: set() for channel_type in
+            ("im", "mpim", "group", "channel", "shared")
+        }
+        for channel in self.channels.values():
+            if (
+                getattr(channel, "is_open", False)
+                or getattr(channel, "is_member", False)
```
trygveaa: `and` has a higher presence than `or`, so you need parenthesis around the first two conditions.

auscompgeek: I think it's probably more sensible to only check `is_general` if it might be a public channel. I say it's mostly fine as is, but maybe add parens anyway to make the intent clear to a casual reader.

ericdwang: Only public channels have `is_member` and `is_general` so I agree it makes sense to group those together instead. It should also be fine to assume `is_general` exists if `is_member` exists so `getattr(channel, "is_general", False)` isn't necessary.

trygveaa:
I tested this again now on a team with many users, and saw that it only subscribed to 36 users. The reason is that most of the channels initially don't contain any other members than myself. Members is only populated after `channels.info` is called in `create_buffer`, which mostly happens after `subscribe_users_presence` has been called.

And sorry for the long wait, was on summer vacation.

PR #719, opened 2019-09-11
Update README.md

oksiquatzel:
It tooked me a while to figure out why weechat says that websocket module was not found in wee-slack plugin. 
pip list sayed websocket-client was installed.
after installing pip3 and websocket-client with pip3, everything worked fine. 
The current ubuntu 19.04 `python-websocket` package has always installed the python2.7 version.
The Dependencies should reflect this major change in weechat


trygveaa:
Thanks!

Issue #667, opened 2019-01-28
Support polls / vote


luchmhor:
I noticed that slack allows to create polls. I have not found out how to create a poll using wee-slack, neither how to vote on existing polls.
Is it doable?

trygveaa:
You can create a poll by running:
```
/slack slash /poll "Some poll text" "Option 1" "Option 2"
```

I don't think you can vote in polls though. I looked through Slacks API documentation, but I didn't find any method to trigger message buttons. So if we want to implement it, we may have to check what the web client does, and see if we can do the same (I'm generally not a fan of doing things outside of the API though).

tfheen:
https://api.slack.com/interactive-messages#responding seems relevant. Not sure how much we're into well-documented territory here, though.

trygveaa:
Yeah, I saw that. I haven't read it in detail, but it only seems to describe how you create a message with buttons, and how you can respond when a user clicked a button you created. It doesn't say anything about triggering a button as far as I can see.

aereaux:
To add to this: previously, the votes in the poll were shown, but recently it started only showing a message about there being a poll.  Is there anything that can be done about that?

auscompgeek:
@aereaux you'll have to give more details; what poll app are you referring to?

Either way, I hate the idea of these poll apps. Reactions work just fine for polls, and is even suggested by Slack for that purpose.

aereaux:
I'm fairly sure it's called Polly.  That's the username that is shown.  The change happened sometime on Friday.

I agree about these poll apps, but I don't think it would be easy to get the rest of my team to switch.

trygveaa:
With #729 polls are now shown again.

Responding to polls is still not possible unfortunately, and I still don't know if it's possible. I checked the web client, and it uses an undocumented API endpoint which only works with the tokens the web client use, and not with oauth tokens, so we can't use that.

Issue #726, opened 2019-10-18
Zoom Slack integration does not show channel's link


gianarb:
Hello

At work we use zoom and we have the slack integration enabled.
When somebody starts a new meeting I see a message from the Zoom Bot but I don't see the actual link:

```
Zoom :] John has started a meeting
```

What I see in Slack Client is:

![](https://s3-us-west-2.amazonaws.com/slack-files2/images/2019-07-29/708237923684_0e8412fea57944ba4049_1600.png)

trygveaa:
Can you post the json for the message as received from the api? You can get this by running `/set plugins.var.python.slack.record_events true` and reloading wee-slack to enable dumping of all the api responses to `/tmp/weeslack-debug/websocket`. Then, when the message is received it will be stored in a json file in that directory. Search the directory for zoom to find it, and post it here.

auscompgeek:
It looks like the Zoom app uses the new [message blocks](https://api.slack.com/block-kit). Good luck to whoever decides to tackle this...

```json
{
  "type": "message",
  "subtype": "bot_message",
  "text": "davo has started a meeting",
  "suppress_notification": false,
  "bot_id": "BJNCM6SLC",
  "team": "TREDACTED",
  "bot_profile": {
    "id": "BJNCM6SLC",
    "deleted": false,
    "name": "Zoom",
    "updated": 1557835614,
    "app_id": "A5GE9BMQC",
    "icons": {
      "image_36": "https://slack-files2.s3-us-west-2.amazonaws.com/avatars/2017-09-21/245295805989_06e77af1bfb8e3c81d4d_36.png",
      "image_48": "https://slack-files2.s3-us-west-2.amazonaws.com/avatars/2017-09-21/245295805989_06e77af1bfb8e3c81d4d_48.png",
      "image_72": "https://slack-files2.s3-us-west-2.amazonaws.com/avatars/2017-09-21/245295805989_06e77af1bfb8e3c81d4d_72.png"
    },
    "team_id": "TREDACTED"
  },
  "blocks": [
    {
      "type": "call",
      "call_id": "RPKGPH83F",
      "block_id": "QLp",
      "api_decoration_available": true,
      "call": {
        "v1": {
          "id": "RPKGPH83F",
          "app_id": "A5GE9BMQC",
          "app_icon_urls": {
[...]

PR #729, opened 2019-11-03
Add implementation of slack layout blocks

aereaux:
See https://api.slack.com/reference/block-kit/blocks .

This is used by, among other things, the Polly poll bot, and fixes the issue I brought up in #667.

Is there a function I can use for rendering message formatting?  Right now the poll options include `*text*` and dates: `!date^1572642900^{date_short_pretty} at {time} (Today at 21:15 PM)`

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -3390,7 +3390,7 @@ def unfurl_blocks(message_json):
             elif block["type"] == "context":
                 block_text.append("|".join(i["text"] for i in block["elements"]))
             else:
-                block_text.append(' {}<<Unsupported block type "{}">>{}'.format(w.color(config.color_reaction_suffix), block["type"], w.color("reset")))
+                block_text.append(' {}<<Unsupported block type "{}">>{}'.format(w.color(config.color_deleted), block["type"], w.color("reset")))
                 dbg('Unsupported block: "{}"'.format(json.dumps(block)), level=4)
         except Exception as e:
             dbg("Failed to unfurl block ({}): {}".format(repr(e), json.dumps(block)), level=4)
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -3379,7 +3379,7 @@ def unfurl_blocks(message_json):
     for block in message_json["blocks"]:
         try:
             if block["type"] == "section":
-                if "text" in block: block_text.append(block["text"]["text"])
+                if "text" in block: block_text += unfurl_texts([block["text"]])
                 if "fields" in block: block_text += unfurl_texts(block["fields"])
             elif block["type"] == "actions":
                 block_text.append("|".join(i["text"]["text"] for i in block["elements"]))
@@ -3389,7 +3389,7 @@ def unfurl_blocks(message_json):
                 block_text.append("|".join(i["text"] for i in block["elements"]))
             else:
                 block_text.append(json.dumps(block))
-        except KeyError as e:
+        except Exception as e:
             block_text.append(json.dumps(block) + repr(e))
     return "\n".join(block_text)
 
@@ -3397,7 +3397,11 @@ def unfurl_blocks(message_json):
 def unfurl_texts(texts):
     texts_ret = []
     for text in texts:
-        texts_ret.append(text["text"]) #TODO: markdown, etc. parse?
+        if text["type"] == "mrkdwn":
+            ftext = render_formatting(text["text"])
+        else:
+            ftext = text["text"]
+        texts_ret.append(ftext)
     return texts_ret
 
 
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -3384,11 +3384,13 @@ def unfurl_blocks(message_json):
             elif block["type"] == "actions":
                 block_text.append("|".join(i["text"]["text"] for i in block["elements"]))
             elif block["type"] == "call":
-                block_text.append(". Join via ").append(block["call"]["v1"]["join_url"])
+                block_text.append(". Join via " + block["call"]["v1"]["join_url"])
             elif block["type"] == "divider":
                 block_text.append("\n")
             elif block["type"] == "context":
                 block_text.append("|".join(i["text"] for i in block["elements"]))
+            elif block["type"] == "rich_text":
+                continue
             else:
                 block_text.append(' {}<<Unsupported block type "{}">>{}'.format(w.color(config.color_deleted), block["type"], w.color("reset")))
                 dbg('Unsupported block: "{}"'.format(json.dumps(block)), level=4)
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -3383,6 +3383,8 @@ def unfurl_blocks(message_json):
                 if "fields" in block: block_text += unfurl_texts(block["fields"])
             elif block["type"] == "actions":
                 block_text.append("|".join(i["text"]["text"] for i in block["elements"]))
+            elif block["type"] == "call":
+                block_text.append(". Join via ").append(block["call"]["v1"]["join_url"])
             elif block["type"] == "divider":
                 block_text.append("\n")
             elif block["type"] == "context":
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2474,11 +2474,11 @@ class SlackMessage(object):
         else:
             text = ""
 
-        if "blocks" in self.message_json:
-            text += unfurl_blocks(self.message_json)
-
         if self.message_json.get('mrkdwn', True):
             text = render_formatting(text)
+
+        if "blocks" in self.message_json:
+            text += unfurl_blocks(self.message_json)
 
         text = unfurl_refs(text)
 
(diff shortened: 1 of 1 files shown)


gianarb:
I think it is the same used by zoom #726 from my understanding

aereaux:
> I think it is the same used by zoom #726 from my understanding

If you could try out this PR and let me know how it looks, that might be great.  I'm not really sure the best way to try to match slack's styling.

gianarb:
@aereaux I tried it with zoom and it works! I am happy with what I see, it is totally `not pretty`, but it is not your fault. I see too many information in my case:

2019-11-07-11-13-50_2122x273

When in practice I think I would like to just see the `join_url`, but I think it depends on app by app. I am more than happy with this PR!

gianarb:
looking at the code I think it will be nice to add a case `type=call` that returns the concatenation of:

```
name+join_url
```

aereaux:
OK, I think this is mostly in a state I'm happy with.  The only problem is that I'm not sure if there's a better way of handling errors.

gianarb:
@aereaux do you know why I can't see here the commit I sent you for the block call?
Thanks

aereaux:
Sorry, I think I accidentally force-pushed it away.  I've readded it.

gianarb:
No problem @aereaux I have added support for `rich_text` coming from the new editor

https://github.com/aereaux/wee-slack/pull/3

aereaux:
@gianarb What do you think of my most recent changes.  It gets rid of the debug output in the message text.  I don't feel quite right just dropping the content of the rich_text message.

gianarb:
I installed it right now! I would really like to get this PR merged!

gianarb:
@aereaux I didn't even the possibility to figure it out yet but since I installed this version the `/zoom` command stopped to work properly and the message does not contain the join URL anymore. Kind of confused because I lost the version that worked for me previously. :disappointed:

BrianGallew:
Something, something, never force push.  I'm sad that you lost the history because I was hoping to try this out myself.

trygveaa:
How did you loose your changes? If you still have the same git clone, you may be able to get the commit back by using the reflog.

gianarb:
well :D my setup was pretty bad, I think at some point I just copy pasted some file out from the repository. I will take some time today to get the raw message back in order to figure out why the link does not show up for me anymore @BrianGallew my issue is only with zoom at the moment, the rest work great, if you have a bit more knowledge than myself (zero) with python you should try as well!

gianarb:
@trygveaa @BrianGallew just for my sanity I opened a PR based on this one with a couple of fixes 

https://github.com/wee-slack/wee-slack/pull/742

@BrianGallew this works as expected (at least for me)

wee_slack.py
```diff
@@ -2428,6 +2428,9 @@ def render(self, force=False):
         else:
             text = ""
 
+        if "blocks" in self.message_json:
+            text += unfurl_blocks(self.message_json)
```
trygveaa: This should go after the next two lines, so it's not overwritten by those.

wee_slack.py
```diff
@@ -3311,6 +3314,39 @@ def linkify_word(match):
...
+                block_text.append("|".join(i["text"]["text"] for i in block["elements"]))
+            elif block["type"] == "call":
+                block_text.append(". Join via ").append(block["call"]["v1"]["join_url"])
+            elif block["type"] == "divider":
+                block_text.append("\n")
+            elif block["type"] == "context":
+                block_text.append("|".join(i["text"] for i in block["elements"]))
+            else:
+                raise NotImplementedError("Block type not implemented", block["type"])
+        except Exception as e:
```
trygveaa: What exceptions can be end up here? I don't it's a good idea to catch general exceptions as you may hide bugs in your implementation.

wee_slack.py
```diff
@@ -3311,6 +3314,39 @@ def linkify_word(match):
...
+            elif block["type"] == "actions":
+                block_text.append("|".join(i["text"]["text"] for i in block["elements"]))
+            elif block["type"] == "call":
+                block_text.append(". Join via ").append(block["call"]["v1"]["join_url"])
+            elif block["type"] == "divider":
+                block_text.append("\n")
+            elif block["type"] == "context":
+                block_text.append("|".join(i["text"] for i in block["elements"]))
+            else:
+                raise NotImplementedError("Block type not implemented", block["type"])
```
trygveaa: Instead of raising an exception here, I think we should render `<Unsupported block type "type">` in the message, and log the block to the debug buffer.

wee_slack.py
aereaux: Thanks for reviewing this!  The problem here is that if there is an error in my implementation, it will just not display the message, and you may or may not see the error (I think it goes in the weechat buffer).  This way it will at least display the message and then log the error.  It may be less of a concern now that my implementation has somewhat fewer errors, but it was really annoying to not have messages show up.  What do you think the best solution is here?

wee_slack.py
aereaux: OK, I added a solution for this.  I use the same coloring as for reactions.  I'm not sure if this is the best solution.

wee_slack.py
trygveaa: Right, I forgot that because of the exception handling in `handle_next` errors are just silently ignored when you have debug_mode off. I've been meaning to change that.

Okay, we can keep the exception handler. Though, at some point, I'll probably change the exception handler in `handle_next` to print an error saying you should report the issue to github when exceptions occur. If that is after we have after verified that it matches Slacks documentation for blocks I think we can remove the exception handler here.

I can't find any documentation for the `call` type though. Do you, or @gianarb know if that exists? I assume the implementation here just matches what Zoom sends?

wee_slack.py
trygveaa: Hm, don't think we should reuse the color for reactions. But it feels a bit unnecessary to add a separate config for it, since it hopefully won't be shown that often.

I added a config for deleted messages now, `color_deleted`. What do you think about using that?

wee_slack.py
```diff
@@ -3311,6 +3314,40 @@ def linkify_word(match):
...
+    block_text = []
+    for block in message_json["blocks"]:
+        try:
+            if block["type"] == "section":
+                if "text" in block: block_text += unfurl_texts([block["text"]])
+                if "fields" in block: block_text += unfurl_texts(block["fields"])
+            elif block["type"] == "actions":
+                block_text.append("|".join(i["text"]["text"] for i in block["elements"]))
+            elif block["type"] == "call":
+                block_text.append(". Join via ").append(block["call"]["v1"]["join_url"])
```
trygveaa: Calling `.append` twice like this doesn't work since `.append` returns `None`. @gianarb has fixed this in https://github.com/wee-slack/wee-slack/pull/742/commits/4ec9aa586c480f8606795b9f261d84a8bb6a2497.

That also drops the `rich_text` type which we want to do, so we don't print `Unsupported block type` after each message sent from the web client. I saw you said you don't feel quite right just dropping the content, but Slack has said the `text` field will still contain the message[0], and we don't have anything to do with the field unless we use it to render the message (and then we would have to drop the text field, since we don't want the message twice).

Can you pull in that commit to this PR (except for the last part which changes the else-block)? Or can/should I pull it into this PR?

[0]: From https://api.slack.com/changelog/2019-09-what-they-see-is-what-you-get-and-more-and-less:
> The text field will continue to contain a textual and markup-laden representation of the user's typed message.
> Messages are still messages. These ones are still written by users. You'll still find the essence of their communication in the text field of message objects.

wee_slack.py
aereaux: Sounds good.

wee_slack.py
aereaux: Yeah, I think the ideal solution would be to render as much as we can (not sure how that would work), and then add a message (perhaps in message_deleted color) to see more in the debug log.

I think slack's documentation is really lacking.  I haven't seen anything about call blocks.

wee_slack.py
```diff
@@ -3384,13 +3384,15 @@ def unfurl_blocks(message_json):
...
+                block_text.append(". Join via " + block["call"]["v1"]["join_url"])
             elif block["type"] == "divider":
                 block_text.append("\n")
             elif block["type"] == "context":
                 block_text.append("|".join(i["text"] for i in block["elements"]))
+            elif block["type"] == "rich_text":
+                continue
             else:
-                block_text.append(' {}<<Unsupported block type "{}">>{}'.format(w.color(config.color_deleted), block["type"], w.color("reset")))
+                block_text.append(' {}<<Unsupported block type "{}">>{}'.format(w.color(config.color_reaction_suffix), block["type"], w.color("reset")))
```
trygveaa: I don't think you meant to include this change here.

aereaux: Yep, sorry about that.

trygveaa:
Thanks!

trygveaa:
@aereaux, @gianarb: I did some tweaking of the formatting, added checking of the action type, used unfurling of text objects in context and action blocks, and added support for the image type. Feel free to comment on my changes if you want:https://github.com/wee-slack/wee-slack/compare/c61e15dbbfc09c520b944a63c97ba07c2261ee30...1fe556f16751889d6ca92d7ecea408eea6d3d4fa

trygveaa:
I've now added support for showing dates correctly.

Issue #629, opened 2018-09-11
History is re-loaded every once in a while


matthijskooijman:
It seems that every once in a while, wee-slack forgets about a channel's (or private chat's) history. This means that when I switch to the buffer, wee-slack first has to get the channel history before I can see new messages.

This is a problem because:
 - Getting the channel history takes a while, so I have to wait
 - I have a script that forwards highlights to another machine to show them on my desktop, so old highlights get shown again.

I realize that lazily loading history is the default, but I have `plugins.var.python.slack.background_load_all_history` set to `true`, which I would understand to mean to automatically load all history on load, and then also keep it in memory.

It could be that my highlight script is not processing input correctly. I'm doing this (in Python):

        w.hook_print("", "notify_message",   "", 1, "on_msg", "")
        w.hook_print("", "notify_private",   "", 1, "on_msg", "private")
        w.hook_print("", "notify_highlight", "", 1, "on_msg", "")

I looked through the weechat API docs, but could not see a way to distinguish between new messages and older messages obtained from history. It could be that the slowness I'm seeing when loading the history is partly because of my highlight script that has to process all these highlights, so if there is some way to make this distinction, that would be great (but that would not fix the actual subject of this issue, namely that history is not kept in memory).

Given this behaviour, I'm wondering if highlights are correctly processed by this script when a buffer's history is dropped, before I open the buffer again?

All this is observed using 0eaf9ed4057f4adbfeb57289f7cc12774027a9aa (about a month old, slightly older than v2.1.0) but the commit log does not show any commits that would obviously fix this.

This issue is tangentially related to #627 (which asks for dropping the entire history handling and just use normal weechat logging). Fixing that would workaround this issue, I think.

matthijskooijman:
Great, the new-marking is indeed fixed now; my notifications don't trigger erronously anymore. This also greatly speeds up the history loading.

Shall we leave this issue open for the other bug, where background loading does not happen on a websocket reconnect?

trygveaa:
> Shall we leave this issue open for the other bug, where background loading does not happen on a websocket reconnect?

Sure, there isn't any other open issue for it as far as I know.

Zolotkey:
I am having a similar issue, with merged buffers of 2-3 channels. It will update all the channels in real time which is what I want. ex: I am in channel2..
```
#channel1 <user> blah
#channel2 <user2> text
#channel1 <user3> text
```

[...]

PR #732, opened 2019-11-08
Add check_should_open call in the reconnect path.

aereaux:
This ensures that the history gets reloaded if
background_load_all_history is set to true.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -432,6 +432,8 @@ class EventRouter(object):
                 team.set_disconnected()
             if not team.connected:
                 team.connect()
+                for c in team.channels.keys():
+                    team.channels[c].check_should_open()
                 dbg("reconnecting {}".format(team))
 
     def receive_ws_callback(self, team_hash):


aereaux:
This fixes #629.

trygveaa:
I tested this, but since it runs right after `connect` it doesn't work in all cases. If `ws_url` (in `connect`) is not set, it will queue up a request to connect, so when the added `check_should_open` it hasn't reconnected yet.

Additionally, I'm sometimes getting this error when `check_should_open` runs as added in this PR:
```
AttributeError: 'SlackChannel' object has no attribute 'is_member'
```
But I'm not sure if that's caused by this PR, I will have to test more.

trygveaa:
The `is_member` error is because channels created after wee-slack has been loaded didn't have this attribute set. `is_member` is only read in `check_should_open` which was only called when the script initially loads, so this wasn't a problem before. I've added `is_member` to channels created after the script is loaded in commit 8a7663399a1ca723f008fd7ca3eb05aa662aab38, so this problem is fixed now.

I'll have to test the first thing I mentioned in the previous commit more.

pasja:
I have been using this patch over the last week with 8a76633 and looks good to me, I have not discovered any hindrance.
I have two spaces, a premium and a free one and both of them are ok.

trygveaa:
I am experiencing problems with rate limiting of requests sporadically when it reconnects. When that happens, it basically hammers the slack servers and never recovers (until you reload the script).

The issue isn't really introduced in this PR, but the extra requests this PR introduces causes it to happen more often, so I think it would be best if the rate limiting issue was fixed before this is merged.

aereaux:
Another problem that I've found with this PR is that if wee-slack has problems connecting to the slack API (where I get `Failed connecting to slack team with token <token>, retrying.`), then after it successfully reconnects I get a `getting channel history...` at the end of each channel (after the rest of the history).  I'll see if I can do a little investigation to figure out why this is, but if anyone else has any ideas, let me know.

trygveaa:
@aereaux: I'm now working on fixing this in a different way, and are changing a lot of things, including the handling of `getting channel history`, so I don't think it's worth investigating that. If you want, you can test https://github.com/trygveaa/wee-slack/tree/history-after-reconnect (it's not finished yet though).

aereaux:
Will do, and thanks for working on this.  Any known issues I should watch out for?

trygveaa:
There might be some issues with threads, but I don't have anything particular you should watch out for.

trygveaa:
I fixed this in a different way, so I'm closing this. History now reloads automatically when `background_load_all_history` is set to true.

PR #742, opened 2020-01-09
manage unsupported block and rich text

gianarb:
I added support for reach_text block otherwise at the end of all
messages wee-slack was printing the raw JSON block.

I have also fixed the `call` block to print the right join url

I don't know if `dbg("DEBUG: {}").format(json.dumps(block))` this is the right way to do it but my plan here was to log in debug mode blocks that are not yet supported. If we prefer to print the raw JSON I can roll it back.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -3324,13 +3324,15 @@ def unfurl_blocks(message_json):
             elif block["type"] == "actions":
                 block_text.append("|".join(i["text"]["text"] for i in block["elements"]))
             elif block["type"] == "call":
-                block_text.append(". Join via ").append(block["call"]["v1"]["join_url"])
+                block_text.append(". Join via " + block["call"]["v1"]["join_url"])
             elif block["type"] == "divider":
                 block_text.append("\n")
             elif block["type"] == "context":
                 block_text.append("|".join(i["text"] for i in block["elements"]))
+            elif block["type"] == "rich_text":
+                continue
             else:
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -3324,13 +3324,15 @@ def unfurl_blocks(message_json):
...
+                block_text.append(". Join via " + block["call"]["v1"]["join_url"])
             elif block["type"] == "divider":
                 block_text.append("\n")
             elif block["type"] == "context":
                 block_text.append("|".join(i["text"] for i in block["elements"]))
+            elif block["type"] == "rich_text":
+                continue
             else:
-                block_text.append(json.dumps(block))
+                dbg("DEBUG: {}").format(json.dumps(block))
```
trygveaa: The first `)` should be after `format` instead of after `dbg`. The way it is now, you are calling `format` on the return value of `dbg`, which is `None`. That is why you would get the raw json block without the `rich_text` condition (it would raise an exception and go to the exception handler below).

Though I don't think a debug message like this with just the json contents is useful. It should also include a text saying something about unknown block type. I also think we should render `<Unsupported block type "type">` in the message (I mentioned this in the other PR too, [here](https://github.com/wee-slack/wee-slack/pull/729#discussion_r380413554)).

trygveaa:
The extra commit here is now included in #729 (except for the change in the else-block), so I'll close this.

PR #752, opened 2020-02-21
Enable render_emoji_as_string = 'both'

zware:
This allows setting `render_emoji_as_string` to `'both'`, showing both representations of emoji (`<emoji character> (:text-representation:)`), which allows one to at least see the text representation when their terminal doesn't render the emoji character.


zware:
Thanks for merging!

PR #753, opened 2020-02-29
Add ability to broadcast a thread message to the rest of the channel.

### LINKED ISSUE #491: Send thread message to parent channel
This is a feature present in the slack for web client & others. Where a message can go to both the thread & parent channel. Very useful for announcing the conclusion of the thread.

aereaux:
This adds an optional -broadcast parameter to /reply and a /slack
broadcast command.
This should fix #491

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -4004,39 +4004,35 @@ command_thread.completion = '%(threads)'
 
 @slack_buffer_required
 @utf8_decode
-def command_broadcast(data, current_buffer, args):
-    """
-    /slack broadcast <text>
-    When in a thread buffer, send a message while broadcasting to the rest of
-    the channel.
-    """
-    channel = EVENTROUTER.weechat_controller.buffers[current_buffer]
-
-    channel.send_message(args, request_dict_ext={'reply_broadcast': True})
-    return w.WEECHAT_RC_OK_EAT
-
-
-@slack_buffer_required
-@utf8_decode
 def command_reply(data, current_buffer, args):
     """
-    /reply [-broadcast] <count/message_id> <text>
-    Reply in a thread on the message. Specify either the message id
-    or a count upwards to the message from the last message.
+    /reply [-alsochannel] <count/message_id> <text>
+    Reply in a thread on the message. Specify either the message id, a count
+    upwards to the message from the last message, or, if you are already in a
+    thread, nothing.
     """
     channel = EVENTROUTER.weechat_controller.buffers[current_buffer]
+    parts = args.split(None, 1)
+    if parts[0] == "-alsochannel":
+        args = parts[1]
+        broadcast = True
+    else:
+        broadcast = False
+
     try:
-        msg_id, text = args.split(None, 1)
-        if msg_id == "-broadcast":
-            msg_id, text = text.split(None, 1)
-            broadcast = True
+        if isinstance(channel, SlackThreadChannel):
+            text = args
         else:
-            broadcast = False
+            msg_id, text = args.split(None, 1)
     except ValueError:
-        w.prnt('', 'Usage: /reply [-broadcast] <count/id> <message>')
+        w.prnt('', 'Usage: /reply [-alsochannel] <count/id> <message>')
         return w.WEECHAT_RC_OK_EAT
 
-    msg = get_msg_from_id(channel, msg_id)
+    if isinstance(channel, SlackThreadChannel):
+        msg = channel.parent_message
+    else:
+        msg = get_msg_from_id(channel, msg_id)
+
     if msg:
         parent_id = str(msg.ts)
     elif msg_id.isdigit() and int(msg_id) >= 1:
@@ -4049,7 +4045,7 @@ def command_reply(data, current_buffer, args):
     channel.send_message(text, request_dict_ext={'thread_ts': parent_id, 'reply_broadcast': broadcast})
     return w.WEECHAT_RC_OK_EAT
 
-command_reply.completion = '-broadcast %(threads)||%(threads)'
+command_reply.completion = '-alsochannel %(threads)||%(threads)'
 
 
 @slack_buffer_required
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -4006,10 +4006,17 @@ command_thread.completion = '%(threads)'
 @utf8_decode
 def command_reply(data, current_buffer, args):
     """
+    When in a channel buffer:
     /reply [-alsochannel] <count/message_id> <text>
-    Reply in a thread on the message. Specify either the message id, a count
-    upwards to the message from the last message, or, if you are already in a
-    thread, nothing.
+    Reply in a thread on the message. Specify either the message id or a count
+    upwards to the message from the last message.
+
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -4000,6 +4001,21 @@ def command_thread(data, current_buffer, args):
 
 command_thread.completion = '%(threads)'
 
+
+@slack_buffer_required
+@utf8_decode
+def command_broadcast(data, current_buffer, args):
+    """
+    /broadcast <text>
```
trygveaa: You should write `/slack broadcast <text>` here, as it won't work as just `/broadcast`.

wee_slack.py
```diff
@@ -4011,8 +4027,13 @@ def command_reply(data, current_buffer, args):
...
     try:
         msg_id, text = args.split(None, 1)
+        if msg_id == "-broadcast":
+            msg_id, text = text.split(None, 1)
+            broadcast = True
+        else:
+            broadcast = False
     except ValueError:
-        w.prnt('', 'Usage: /reply <count/id> <message>')
+        w.prnt('', 'Usage: /reply [-broadcast] <count/id> <message>')
```
trygveaa: You should add `[-broadcast]` to the docstring some lines above as well (I wasn't allowed to make a comment at that line).

wee_slack.py
```diff
@@ -4025,7 +4046,7 @@ def command_reply(data, current_buffer, args):
         w.prnt('', 'ERROR: Invalid id given, must be a number greater than 0 or an existing id')
         return w.WEECHAT_RC_OK_EAT
 
-    channel.send_message(text, request_dict_ext={'thread_ts': parent_id})
+    channel.send_message(text, request_dict_ext={'thread_ts': parent_id, 'reply_broadcast': broadcast})
     return w.WEECHAT_RC_OK_EAT
 
 command_reply.completion = '%(threads)'
```
trygveaa: You should update the completion pattern to include `-broadcast`.

trygveaa:
When I'm sending broadcast messages, they are not printed in the channel buffer.

I guess we don't get a normal message on the websocket for the ones we send ourselves, but a reply message instead.

aereaux:
Huh.  I have thread_messages_in_channel enabled, so I didn't notice that.  Is there a good way to fix that?  I can't think of one off the top of my head, but I'm not too familiar with the codebase.

tfheen:
I haven't tried the patch, but I think this is really cool to get in. Can we have `-channel`or similar as the name instead (or in addition to) of `-broadcast`? I think broadcasting might cause confusion with `@here` and `@channel`. The UI in the Slack client says "also send to channel".

trygveaa:
@tfheen: Hm, maybe that's clearer, or perhaps `-also-send-to-channel` (or is it too long, even with tab completion?). Though, then it will be a different name from `/slack broadcast`. Maybe instead of adding a separate command, we should allow `/reply` without an id when you're in a thread channel?

So:
From a channel: `/reply -also-send-to-channel $123 message`
From a thread: `/reply -also-send-to-channel message`
?

trygveaa:
@aereaux: I'l try to look into it.

tfheen:
I like the `/reply -channel <message>` idea, so +1 from me on that. I'd just do `-channel` (or maybe `-alsochannel`) and not mimic the UI of slack completely. Weechat seems to just run words together rather than using dashes from what I've seen.

trygveaa:
I like `-alsochannel`. I think `-channel` might be bit ambiguous if it is only goes to the channel or to the thread and the channel.

@aereaux: Can you change the reply argument from `-broadcast` to `-alsochannel`, remove command_broadcast and change reply to work without the id (so it sends to the current buffer)?

trygveaa:
`/reply` without an id should probably only work in thread channels, so people don't send to the channel by accident when forgetting to include the id.

aereaux:
OK, I think I implemented the reply logic as requested.

wee_slack.py
```diff
@@ -4000,22 +4001,38 @@ def command_thread(data, current_buffer, args):
...
     except ValueError:
-        w.prnt('', 'Usage: /reply <count/id> <message>')
+        w.prnt('', 'Usage: /reply [-alsochannel] <count/id> <message>')
         return w.WEECHAT_RC_OK_EAT
 
-    msg = get_msg_from_id(channel, msg_id)
+    if isinstance(channel, SlackThreadChannel):
+        msg = channel.parent_message
+    else:
+        msg = get_msg_from_id(channel, msg_id)
```
trygveaa: I think this is a bit more readable if you combine the two checks for thread channel, like this:

```
    if isinstance(channel, SlackThreadChannel):
        text = args
        msg = channel.parent_message
    else:
        try:
            msg_id, text = args.split(None, 1)
        except ValueError:
            w.prnt('', 'Usage: /reply [-alsochannel] <count/id> <message>')
            return w.WEECHAT_RC_OK_EAT
        msg = get_msg_from_id(channel, msg_id)
```

wee_slack.py
```diff
@@ -4000,22 +4001,38 @@ def command_thread(data, current_buffer, args):
...
 
+
 @slack_buffer_required
 @utf8_decode
 def command_reply(data, current_buffer, args):
     """
-    /reply <count/message_id> <text>
-    Reply in a thread on the message. Specify either the message id
-    or a count upwards to the message from the last message.
+    /reply [-alsochannel] <count/message_id> <text>
```
trygveaa: Since the count/id is optional now, this should be changed to ` [count/message_id]`. Same with the usage-line below.

wee_slack.py
```diff
@@ -4000,22 +4001,38 @@ def command_thread(data, current_buffer, args):
...
+    Reply in a thread on the message. Specify either the message id, a count
+    upwards to the message from the last message, or, if you are already in a
+    thread, nothing.
```
trygveaa: Maybe `-alsochannel` is self explanatory, but I think we should add a description of it, e.g. "If you specify -alsochannel, the message will be posted to both the thread and the channel.".

wee_slack.py
aereaux: How's this?  I think it makes it more explicit which form is allowed when.

trygveaa: Yeah, that would be good, except that it looks a bit weird in the help text inside weechat, because the code which generates the help text from the docstring expects the command first.

With this `/help reply` outputs this:

```
23:16 [python/slack]  /reply  in a channel buffer:
23:16 
23:16 /reply [-alsochannel] <count/message_id> <text>
23:16 Reply in a thread on the message. Specify either the message id or a count
23:16 upwards to the message from the last message.
23:16 
23:16 When in a thread buffer:
23:16 /reply [-alsochannel] <text>
23:16 Reply to the current thread.  This can be used to send the reply to the
23:16 rest of the channel.
```

One possible fix for it would be keeping `/reply [-alsochannel] [count/message_id] <text>` as the first line, and then include what you wrote below.

trygveaa:
Thanks!

I fixed the issue regarding the messages not appearing in the channel buffer that I mentioned above.

PR #758, opened 2020-03-15
Allow subscribing to threads

aereaux:
It currently seems to work on threads that are automatically subscribed
(such as by saying something in them or by subscribing in the slack web UI).

Does anyone know what the difference between thread_message and message_replied is?  They both have some processing of thread messages.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2508,8 +2508,8 @@ class SlackMessage(object):
 
         if self.number_of_replies():
             self.channel.hash_message(self.ts)
-            text += " " + colorize_string(get_thread_color(self.hash), "[ Thread: {} Replies: {} Subscribed: {} ]".format(
-                    self.hash, self.number_of_replies(), self.subscribed))
+            text += " " + colorize_string(get_thread_color(self.hash), "[ Thread: {} Replies: {}{} ]".format(
+                    self.hash, self.number_of_replies(), " Subscribed" if self.subscribed else ""))
 
         text = replace_string_with_emoji(text)
 
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2432,6 +2432,7 @@ class SlackMessage(object):
             senders = self.get_sender()
             self.sender, self.sender_plain = senders[0], senders[1]
         self.ts = SlackTS(message_json['ts'])
+        self.subscribed = message_json.get("subscribed", False)
 
     def __hash__(self):
         return hash(self.ts)
@@ -2488,8 +2489,8 @@ class SlackMessage(object):
 
         if self.number_of_replies():
             self.channel.hash_message(self.ts)
-            text += " " + colorize_string(get_thread_color(self.hash), "[ Thread: {} Replies: {} ]".format(
-                    self.hash, self.number_of_replies()))
+            text += " " + colorize_string(get_thread_color(self.hash), "[ Thread: {} Replies: {} Subscribed: {} ]".format(
+                    self.hash, self.number_of_replies(), self.subscribed))
 
         text = replace_string_with_emoji(text)
 
@@ -3008,6 +3009,7 @@ def download_files(message_json, team):
 
 
 def subprocess_thread_message(message_json, eventrouter, team, channel, history_message):
+    dbg("THREAD MESSAGE {}".format(message_json))
     parent_ts = message_json.get('thread_ts')
     if parent_ts:
         parent_message = channel.messages.get(SlackTS(parent_ts))
@@ -3021,6 +3023,8 @@ def subprocess_thread_message(message_json, eventrouter, team, channel, history_
 
             if parent_message.thread_channel and parent_message.thread_channel.active:
                 parent_message.thread_channel.buffer_prnt(message.sender, parent_message.thread_channel.render(message), message.ts, tag_nick=message.sender_plain)
+            elif message.ts > channel.last_read and parent_message.subscribed: #TODO: What is the difference between this and message_replied, at least for alerting on threads?
+                parent_message.notify_thread(action="subscribed", sender_id=message_json["user"]) #TODO: Update thread last read mark?
             elif message.ts > channel.last_read and message.has_mention():
                 parent_message.notify_thread(action="mention", sender_id=message_json["user"])
 
@@ -3069,6 +3073,7 @@ subprocess_group_topic = subprocess_channel_topic
 
 
 def subprocess_message_replied(message_json, eventrouter, team, channel, history_message):
+    dbg("MESSAGE REPLIED {}".format(message_json))
     parent_ts = message_json["message"].get("thread_ts")
     parent_message = channel.messages.get(SlackTS(parent_ts))
     # Thread exists but is not open yet
@@ -3214,6 +3219,16 @@ def process_subteam_updated(subteam_json, eventrouter, team, channel, metadata):
 
 def process_emoji_changed(message_json, eventrouter, team, channel, metadata):
     team.load_emoji_completions()
+
+
+def process_thread_subscribed(message_json, eventrouter, team, channel, metadata):
+    dbg("THREAD SUBSCRIBED {}".format(message_json))
+    team.channels[message_json["subscription"]["channel"]].messages.get(SlackTS(message_json["subscription"]["thread_ts"])).subscribed = True
+
+
+def process_thread_unsubscribed(message_json, eventrouter, team, channel, metadata):
+    dbg("THREAD UNSUBSCRIBED {}".format(message_json))
+    team.channels[message_json["subscription"]["channel"]].messages.get(SlackTS(message_json["subscription"]["thread_ts"])).subscribed = False
 
 
 ###### New module/global methods
@@ -4099,6 +4114,31 @@ def command_thread(data, current_buffer, args):
     return w.WEECHAT_RC_OK_EAT
 
 command_thread.completion = '%(threads)'
+
+
+@slack_buffer_required
+@utf8_decode
+def command_subscribe(data, current_buffer, args):
+    """
+    /slack subscribe <thread>
+    Subscribe to a thread, so that you are alerted to new messages.  When in a
+    thread buffer, you can omit the thread id.
+    """
+    channel = EVENTROUTER.weechat_controller.buffers[current_buffer]
+
+    if not args:
+        w.prnt('', 'Usage: /slack subscribe <thread>')
+        return w.WEECHAT_RC_ERROR
+
+    if args:
+        msg = get_msg_from_id(channel, args)
+
+    if not msg and isinstance(channel, SlackThreadChannel):
+        msg = channel.parent_message
+
+    # TODO: Looks like there is no RTM api for this?
+
+command_subscribe.completion = '%(threads)'
 
 
 @slack_buffer_required
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2241,8 +2241,15 @@ class SlackThreadChannel(SlackChannelCommon):
         self.members = self.parent_message.channel.members
         self.team = self.parent_message.team
         self.last_line_from = None
-        self.last_read = self.parent_message.message_json.get("last_read", SlackTS())
         self.new_messages = False
+
+    @property
+    def last_read(self):
+        return self.parent_message.last_read
+
+    @last_read.setter
+    def last_read(self, ts):
+        self.parent_message.last_read = ts
 
     @property
     def identifier(self):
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -2971,6 +2973,8 @@ def subprocess_thread_message(message_json, eventrouter, team, channel, history_
 
             if parent_message.thread_channel and parent_message.thread_channel.active:
                 parent_message.thread_channel.buffer_prnt(message.sender, parent_message.thread_channel.render(message), message.ts, tag_nick=message.sender_plain)
+            elif message.ts > channel.last_read and parent_message.subscribed: #TODO: What is the difference between this and message_replied, at least for alerting on threads?
+                parent_message.notify_thread(action="subscribed", sender_id=message_json["user"]) #TODO: Update thread last read mark?
```
trygveaa: last read shouldn't be updated here as this will run when the message is received, and we shouldn't update last read until you read the message.

wee_slack.py
```diff
@@ -4036,6 +4051,31 @@ def command_thread(data, current_buffer, args):
...
+        w.prnt('', 'Usage: /slack subscribe <thread>')
+        return w.WEECHAT_RC_ERROR
+
+    if args:
+        msg = get_msg_from_id(channel, args)
+
+    if not msg and isinstance(channel, SlackThreadChannel):
+        msg = channel.parent_message
+
+    # TODO: Looks like there is no RTM api for this?
```
trygveaa: Yeah, I couldn't find any either. I tried the api the web client uses, but that just says `not_allowed_token_type` if I try it with the tokens we use for wee-slack.

wee_slack.py
aereaux: Yep, this is just a reminder that it might be a good idea to update the thread last mark (wherever that should be done).

wee_slack.py
```diff
@@ -113,7 +113,7 @@
         "history": None,
         "join": None,
         "leave": None,
-        "mark": None,
+        "mark": "subscriptions.thread.mark",
```
trygveaa: Does this work for you? For me it just returns:

```
{
  "ok": false,
  "error": "not_allowed_token_type"
}
```

aereaux: Where do you see this?  I thought I had it working, but then I backed out some changes before I pushed, and it doesn't seem to work now.

aereaux:
There's also another bug that this causes to trigger more often (because I have more buffers open).  If a thread buffer is open when the client disconnects and reconnects, when a new message is sent to the thread, you get "Error: a buffer with same name (lab-life.slack.com.#<channel>.<thread>) already exists".  The only way I've found to get it to work again is by reloading wee-slack.

wee_slack.py
trygveaa: I see it if I call it manually, or if I add this to your branch and navigate away from a thread:

```
def handle_subscriptionsthreadmark(json, eventrouter, team, channel, metadata):
    print(json)
```

Additionally, the thread is not marked as read. Did you confirm that the thread was marked as read when you thought you had it working?

aereaux: I thought I did, but my testing procedure (checking in the slack web client) may have not been entirely accurate.

trygveaa:
> There's also another bug that this causes to trigger more often (because I have more buffers open). If a thread buffer is open when the client disconnects and reconnects, when a new message is sent to the thread, you get "Error: a buffer with same name (lab-life.slack.com.#.) already exists". The only way I've found to get it to work again is by reloading wee-slack.

Yeah, that's an unfortunate bug. It's been on my todo list to fix for a while.

wee_slack.py
aereaux: OK, I get the same result.  This may not be possible at the moment, then.

trygveaa: It turns out that if you use a token pulled out of the web client ([as described here](https://github.com/yuya373/emacs-slack#how-to-get-token)), calling `subscriptions.thread.mark` works.

trygveaa: I don't think we should require users to use those tokens, but maybe we should provide the functionality for users that do?

aereaux: Would we want to unconditionally try these APIs in case they work?  Or should we try to check to see if the user is using a more privileged API token?  This would also allow us to use the thread subscribe API.

trygveaa: Hm, I'm not sure. Just add it unconditionally for now, we can add a condition later if we want.

aereaux: OK, I added that and a couple of other things that rely on the special tokens.

wee_slack.py
aereaux: This now works if one has a special token.

aereaux:
OK, I think this is more or less ready for review.  I'm happy with the features included, and I'm using it as my daily driver, but it still might have some bugs in it.  I also kind of think it's a little messy, but I'm not sure how I'd be able to clean it up without a larger refactoring.

wee_slack.py
```diff
@@ -2240,6 +2240,8 @@ def __init__(self, eventrouter, parent_message):
         self.members = self.parent_message.channel.members
         self.team = self.parent_message.team
         self.last_line_from = None
+        self.last_read = self.parent_message.message_json.get("last_read", SlackTS())
```
trygveaa: This attribute on the parent message is only updated when the channel history is reloaded as far as I can see. That means that if you close the thread buffer and open it again, the last read indicator will be where it was the last time the channel history was loaded.

wee_slack.py
```diff
@@ -3021,6 +3042,8 @@ def subprocess_thread_message(message_json, eventrouter, team, channel, history_
 
             if parent_message.thread_channel and parent_message.thread_channel.active:
                 parent_message.thread_channel.buffer_prnt(message.sender, parent_message.thread_channel.render(message), message.ts, tag_nick=message.sender_plain)
+            elif message.ts > message.message_json.get("last_read", SlackTS()) and parent_message.subscribed:
```
trygveaa: Shouldn't this be `message.ts > parent_message.message_json.get("last_read", SlackTS())`?

wee_slack.py
```diff
@@ -3021,6 +3042,8 @@ def subprocess_thread_message(message_json, eventrouter, team, channel, history_
...
+                parent_message.notify_thread(action="subscribed", sender_id=message_json["user"])
```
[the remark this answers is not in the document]
aereaux: My thinking was that once this is stable, we could probably remove the other notify_thread calls, as (I think) all of them should be covered under subscribed, and then one can unsubscribe from threads that one is no longer interested in.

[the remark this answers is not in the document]
trygveaa: Right, that makes sense.

wee_slack.py
```diff
@@ -1597,6 +1597,20 @@ def calc_hash(ts):
...
+            if self.channel_buffer:
+                w.buffer_set(self.channel_buffer, "unread", "")
+                w.buffer_set(self.channel_buffer, "hotlist", "-1")
+            if not ts:
+                ts = next(reversed(self.messages), SlackTS())
+            if ts > self.last_read:
+                self.last_read = ts
+            if update_remote:
+                s = SlackRequest(self.team, SLACK_API_TRANSLATOR[self.type]["mark"],
+                        {"channel": self.identifier, "ts": ts, **post_data}, channel=self)
```
trygveaa: We still support Python 2, so you can't use `**` to merge dictionaries.

wee_slack.py
```diff
@@ -2266,32 +2267,37 @@ def refresh(self):
         self.rename()
 
     def mark_read(self, ts=None, update_remote=True, force=False):
-        if self.channel_buffer:
-            w.buffer_set(self.channel_buffer, "unread", "")
-            w.buffer_set(self.channel_buffer, "hotlist", "-1")
+        super().mark_read(ts=ts, update_remote=update_remote, force=force, post_data={"thread_ts": self.parent_message.ts})
```
trygveaa: We still support Python 2, so you have to use `super(SlackThreadChannel, self)`.

wee_slack.py
```diff
@@ -4101,6 +4130,56 @@ def command_thread(data, current_buffer, args):
...
+    the thread id.
+    """
+    return subscribe_helper(current_buffer, args, 'Usage: /slack unsubscribe <thread>', "subscriptions.thread.remove")
+
+command_unsubscribe.completion = '%(threads)'
+
+
+def handle_subscriptionsthreadmark(json, eventrouter, team, channel, metadata):
+    dbg(repr(json))
+    if not json["ok"]: w.prnt('', 'Failed to set subscription status.  Check debug log for details.')
```
trygveaa: Just print the error directly, instead of sending users to the debug log, like it's done in `handle_reactionsadd`. And I haven't seen any error messages which contains anything other than `ok` and `error`, so you can remove `dbg(repr(json))`.

wee_slack.py
```diff
@@ -4101,6 +4130,56 @@ def command_thread(data, current_buffer, args):
...
+    if args:
+        msg = get_msg_from_id(channel, args)
+    elif isinstance(channel, SlackThreadChannel):
+        msg = channel.parent_message
+    else:
+        w.prnt('', usage)
+        return w.WEECHAT_RC_ERROR
+
+    s = SlackRequest(team, api,
+            {"channel": channel.identifier, "thread_ts": msg.ts}, channel=channel)
```
trygveaa: You should print errors from this request. I.e. add `handle_subscriptionsthreadadd` and `handle_subscriptionsthreadremove` which does the equivalent of `handle_reactionsadd`.

aereaux:
Thanks for the reviews!  The other couple things a tad more difficult, so I'll see when I can get to them.

wee_slack.py
aereaux: OK, I think I solved this one.

aereaux:
OK, that should take care of everything except marking the channel as read if `thread_messages_in_channel` is set.  I'm also not totally confident in the way that threads are marked when `thread_messages_in_channel` is set, because I don't use that option for my daily use, but it should work.

wee_slack.py
```diff
@@ -3117,6 +3160,18 @@ def process_channel_marked(message_json, eventrouter, team, channel, metadata):
 process_mpim_marked = process_channel_marked
 
 
+def process_thread_marked(message_json, eventrouter, team, channel, metadata):
+    subscription = message_json.get("subscription", {})
+    ts = subscription.get("last_read")
+    thread_ts = subscription.get("thread_ts")
+    channel = team.channels.get(subscription.get("channel"))
+    if ts and thread_ts and channel:
+        thread_channel = channel.thread_channels.get(thread_ts)
```
trygveaa: You need to create a `SlackTS` of the `thread_ts` before you look it up in `thread_channels`.

trygveaa:
> OK, that should take care of everything

Nice. I ran into some issues, so posted a couple of more comments.

> except marking the channel as read if thread_messages_in_channel is set, because I'm not really sure how the logic would work there.

Yeah, that's okay. I think we'll drop it, at least for now. I think we would have to check the timestamps and only mark it as read if the only unread messages are from the thread that's being marked.

wee_slack.py
```diff
@@ -1962,6 +1959,12 @@ def get_typing_list(self):
                 del self.typing[user]
         return typing
 
+    def mark_read(self, ts=None, update_remote=True, force=False, post_data={}):
+        if config.thread_messages_in_channel and not update_remote:
```
trygveaa: The `not` shouldn't be here, it is when `update_remote` is true that we should mark thread channels (I used a bit awkward wording in the last comment, and said we should *not* do it if `update_remote` is false).

wee_slack.py
```diff
@@ -1962,6 +1959,12 @@ def get_typing_list(self):
...
+            for thread_channel in self.thread_channels.values():
+                thread_channel.mark_read(ts=ts, update_remote=update_remote, forece=force, post_data=post_data)
```
trygveaa: I realized now that this only marks the thread buffers you have opened. If you use `thread_messages_in_channel` the point is that you don't have to open thread buffers. So what we should do is go though `hashed_messages` and for all messages that is a parent message and is unread, we should mark it.

If you want, I can fix this.

trygveaa: Typo in `forece`.

aereaux: OK, I removed this for now.  I think that this might be better in another PR, because I think that's also how the other `thread_messages_in_channel` feature would need to be implemented as well.

trygveaa:
Thanks for implementing this! Great timing too, since Slack just changed the contents of the message_replied event, but since we don't need that event after these changes, we didn't have to fix it.

PR #763, opened 2020-04-06
Don't notify about threads when they're opened in the channel

zware:
This prevents messages being added to buffer 1 when a new thread is opened and the message is already expanded in the parent channel.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2579,6 +2579,9 @@ class SlackMessage(object):
     def notify_thread(self, action=None, sender_id=None):
         if config.auto_open_threads:
             self.open_thread()
+        elif config.thread_messages_in_channel:
+            # The usual notifications should occur in the channel, no need to notify elsewhere
+            pass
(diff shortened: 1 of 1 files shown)


zware:
I notice after pushing this that there have been other changes to threading since I wrote this patch that I haven't tested yet.  This may or may not be necessary anymore; will test and leave another comment here with findings :)

zware:
Ok, this does seem to still be necessary (to me :)).

bitdancer:
I agree, and have applied this to my local copy of wee_slack without problems.  Without it, and with #732 also applied locally, every time wee_slack reconnects the messages are repeated, which makes them functionally even less useful.

trygveaa:
I added an option to control this, `notify_subscribed_threads`. It defaults to `auto` which is the same behavior as after this PR, and it can also be set to `true` or `false` to enable/disable notifications regardless of the `auto_open_threads` and `thread_messages_in_channel` options.

PR #782, opened 2020-07-04
Set time zone to UTC for unfurl test

tfheen:


trygveaa:
Ah, didn't think about doing that. Thanks!

tfheen:
It was kinda the simple and dumb way to do it, but that's sometimes the way in tests… :-)

PR #784, opened 2020-07-24
feat: use pipenv to manage the development environment

sudoforge:
[Pipenv][0] makes it easier to maintain a consistent development
environment for projects using Python. You no longer need to manage
`pip` and `virtualenv` separately, nor deal with the drift and other
issues requirements.txt brings. It enables gaining insights into your
dependency graph, but most importantly - it enables deterministic
builds, which prevents the entire class of "it works on my machine"
problems.

[Pipenv][0] is similar to package managers from other ecosystems, such
as Node.js' `npm` or Rust's `cargo`. Additionally, [Pipenv][0] is an
[officially recommended package manager][1] by the Python project.

This patch brings [Pipenv][0] to the wee-slack project to simplify the
development and contribution experience. Additionally, this patch brings
basic contributing documentation to help onboarding.

[0]: https://github.com/pypa/pipenv
[1]: https://packaging.python.org/tutorials/managing-dependencies/#managing-dependencies

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,7 @@
 *.pyc
 .cache/
 *.sublime-*
+
+# Ignore pipenv's lockfile until support for multiple versions of python is
+# dropped
+/Pipfile.lock
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,8 @@ python:
 cache: pip
 
 install:
-  - pip install flake8 pytest websocket-client
+  - pip install pipenv
+  - pipenv install --dev --system
 
 before_script:
   # stop the build if there are Python syntax errors or undefined names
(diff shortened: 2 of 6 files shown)


sudoforge:
Ensuring I didn't break support for Python 2.7 required avoid pinning versions of dependencies (even major versions) as well as avoiding committing the lockfile. If you're open to it, I can add a new commit to this tree that removes support for Python 2.7 (which hit EOL in January 1 2020).

trygveaa:
This seems good. I haven't used pipenv before, but I've used many others including the others you mentioned. I'll take a closer look at it in a couple of weeks when I'm home from vacation.

sudoforge:
Sure thing - enjoy your break.

trygveaa:
Sorry for the delay, but I merged it now.

I removed the section about adding dependencies because people install this either just by copying the script, or by installing it from WeeChats repo, and neither of these methods install dependencies automatically, so I really want to keep dependencies to a minimum. So I don't think it's useful to document how to add new ones.

sudoforge:
That's fair; thanks for taking the time to test, review, and merge it! Hopefully it helps simplify the story of contributing to the project.

PR #788, opened 2020-08-11
Fix usage of dates with non ASCII characters for Python 2

### LINKED ISSUE #419: Rescent Messages Not Showing Up
I just recently set up wee-slack and noticed that there are some inconsistencies from the client and the website. It seems like the sync with slack stopped before it reached the current date.
For example in one of the channels I am in it only has the messages before August 4th (August 3rd was the last day it synced). There is 2 days worth of messages that aren't showing up.
Is there a way to force a sync between wee-slack and slack to fix this?

schodet:
When not using english locale, date could contains non ASCII characters.

Fixes #419

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -3926,7 +3926,7 @@ def resolve_ref(ref):
                     else:
                         token = token.replace('_pretty', '')
                 if token in token_to_format:
-                    return ref_datetime.strftime(token_to_format[token])
+                    return decode_from_utf8(ref_datetime.strftime(token_to_format[token]))
                 else:
                     return match.group(0)
 


trygveaa:
Thanks. I recommend you to upgrade to Python 3 though, such errors won't occur in Python 3.

trygveaa:
This is not related to #419 btw, as the code you fixed is much newer than that issue.

schodet:
About #419, at least, it has the same consequences, when a channel contains a date, recent messages could be dropped, depending on the date.

I use the debian package for weechat, and it surprisingly still uses python2, this will change in next release.

Thanks for merging.

trygveaa:
> About #419, at least, it has the same consequences, when a channel contains a date, recent messages could be dropped, depending on the date.

Yeah, but any exception in the rendering code could cause that.

> I use the debian package for weechat, and it surprisingly still uses python2, this will change in next release.

WeeChat has a Debian repo with up to date builds using Python 3 if you want to use that: https://weechat.org/download/debian/

wee-slack will probably support Python 2 until the next Debian version is released, but I won't guarantee it.

schodet:
Thanks for the tip :)

PR #794, opened 2020-09-09
Don't display the typing indicator for muted conversations

tzvetkoff:
Well, the commit message is pretty obvious.

I actually noticed this as my terminal highlights the tab since there's new console output.


trygveaa:
That makes sense, thanks!

I moved the `not self.muted` condition to the `show_typing` variable, otherwise you get a bug with channels that have set a different short_name.

Issue #800, opened 2020-10-08
Two threads have the same id


ptixed:
See the title, not much else I can add. Obviously that leads to some other issues. Like when i try to open the thread by id it opens first one and I can open second one only when there is some new activity on it. Are those ids generated by slack or wee-slack?

trygveaa:
The versions are generated by wee-slack, and they should be unique. Which version of wee-slack are you using?

ptixed:
I am using current master version

PR #801, opened 2020-10-08
fix possible hash collision for thread ids

ptixed:
#800 

notice that https://github.com/wee-slack/wee-slack/blob/master/wee_slack.py#L40
```
    str = unicode
```
so in expression
```
    while any(x.startswith(short_hash) for x in self if isinstance(x, str)):
``` 
`x` was actually checked if of type `unicode`

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -2287,7 +2287,7 @@ class SlackChannelHashedMessages(dict):
         full_hash = sha1_hex(str(key))
         short_hash = full_hash[:hash_len]
 
-        while any(x.startswith(short_hash) for x in self if isinstance(x, str)):
+        while any(x.startswith(short_hash) for x in self if type(x).__name__ == "str"):
             hash_len += 1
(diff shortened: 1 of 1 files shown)


trygveaa:
Ah, you're using Python 2. I guess I didn't test against that when I changed some things with the id generation in May.

I fixed it in a slightly different way.

trygveaa:
I would recommend upgrading to Python 3 though. Python 2 is not going to be supported forever. I'll probably support it until the next release of Debian (since current Debian stable uses Python 2 for WeeChat, unless you install from WeeChats repo). And you are more likely to run into bugs with Python 2 because I develop against Python 3 and don't test Python 2 as much.

ptixed:
Yeah, my machines use debian and ubuntu + python 2. Thank you for quick turnaround

pasja:
Might worth to mention that if buster-backports is available, then you could install weechat 2.8 which comes from python3 instead of python2.

PR #802, opened 2020-10-10
make message indexes relative to scroll position

### LINKED ISSUE #571: /reply 1 should reply to last _visible_ message
If I have a WeeChat filter set up on a channel, I'd expect `/reply 1` to reply to the last message that I can see. Otherwise, I end up replying to the last message posted in the channel, which means it's pretty much impossible to actually make replies correctly in channels I have filters set up for.

ptixed:
Message indexes should now be relative to what's currently on screen and not entire message buffer which is a more intuitive behavior. 
Probably little too early to merge as I have only done some preliminary tests and the change could affect many commands. 

I allowed myself to remove `reverse` parameter since it looks like it is never used, saved myself implementing some logic that would never be used anyway. 

fixes #571

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1691,15 +1691,23 @@ class SlackChannelCommon(object):
             return
         return message
 
-    def message_from_index(self, index, message_filter=None, reverse=True):
-        for ts in (reversed(self.visible_messages) if reverse else self.visible_messages):
-            message = self.messages[ts]
-            if not message_filter or message_filter(message):
-                index -= 1
-                if index == 0:
-                    return message
+    def message_from_index(self, index, message_filter=None):
 
-    def message_from_hash_or_index(self, hash_or_index=None, message_filter=None, reverse=True):
+        window = w.current_window()
+        scroll = w.hdata_pointer(hdata.window, window, 'scroll')
+        offset = w.hdata_integer(hdata.window_scroll, scroll, 'lines_after')
+
+        for ts in reversed(self.visible_messages):
+            if offset > 0:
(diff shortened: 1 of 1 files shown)


wee_slack.py
```diff
@@ -1691,15 +1691,23 @@ def message_from_hash(self, ts_hash, message_filter=None):
...
-                index -= 1
-                if index == 0:
-                    return message
-
-    def message_from_hash_or_index(self, hash_or_index=None, message_filter=None, reverse=True):
+    def message_from_index(self, index, message_filter=None):
+
+        window = w.current_window()
+        scroll = w.hdata_pointer(hdata.window, window, 'scroll')
+        offset = w.hdata_integer(hdata.window_scroll, scroll, 'lines_after')
```
trygveaa, line 1698: Using the scroll position from the current window makes it behave pretty weird if you have merged buffers. Since lines from multiple buffers are shown, the scroll amount won't match the message offset in one of the buffers.

trygveaa (commented):
> I allowed myself to remove `reverse` parameter since it looks like it is never used, saved myself implementing some logic that would never be used anyway.

Okay, that's fine. I don't remember why I included it, and yeah, I see that it's not used.

> fixes #571

This only fixes a part of that issue. If you define a filter to hide some lines, the will still be counted with this, while the issue is asking for them to be ignored.

I think the way to solve both that and the issue with merged buffers that I mention below is to iterate over the lines in the buffer rather than the messages in the channel, and for each line check if it should be counted or not until you reach the correct line.

PR #813, opened 2020-11-26
allow new threads to be opened via /thread <THREADID>

ashkulz:
This was broken in #774 as it checked if there was at least one reply, which is not the case for new threads.

### DIFF
--- a/docs/Commands.md
+++ b/docs/Commands.md
@@ -201,7 +201,7 @@ List the connected Slack teams.
 ### thread
 
 ```
-/thread [message_id]
+/thread [count/message_id]
 ```
 
 Open the thread for the message.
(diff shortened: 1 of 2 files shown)


trygveaa:
The reason I limited `/thread` to messages with at least one reply is that I want the count argument to only count threads because I think that's more useful, and especially that when you run `/thread` without arguments it opens the last thread.

Though, I do agree that if you provide a message id instead of a count, it should allow all normal messages (not joins etc. because that's not possible/visible in the official client, even though the backend seems to support it). I'll have to look into how to best solve this, we might need to have a separate filter for counts and hashes to support this.

ashkulz:
I don't update wee-slack all that often, and after upgrading this was broken so submitted the PR. The current documentation says `/thread [message_id]` and doesn't mention the index at all -- so to me it is a clear regression.

That said, I do see your point -- implemented a hacky fix, let me know what you think.

trygveaa:
Great, thanks! I think the fix is fine.

PR #815, opened 2020-12-10
Add support to disable teammate link previews

### LINKED ISSUE #381: Disable link preview
Is there a way to disable the url preview feature? I'd like to just see the url and be able to open it normally in my browser but what happens is it expands into a bunch of urls and some text that isn't easily discernible from the original text from person who shared the url. 

I tried disabling image/url preview in the slack interface but that works only for the web client. Wee-slack still expands the urls.

### LINKED ISSUE #526: Fully disable URL previews / unfurling

wohali:
This supersedes #529 and should finally close #526 .

To disable URL unfurling: `/set plugins.var.python.slack.link_previews "false"`, followed by a `/python reload` or a Slack channel close/reopen.

To re-enable, set to `"true"` and the same reload is required.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -3791,6 +3791,8 @@ def unwrap_attachments(message_json, text_before):
             # $author: (if rest of line is non-empty) $title ($title_link) OR $from_url
             # $author: (if no $author on previous line) $text
             # $fields
+            if 'original_url' in attachment and not config.link_previews:
+               continue
             t = []
             prepend_title_text = ''
             if 'author_name' in attachment:
@@ -5364,6 +5366,9 @@ class PluginConfig(object):
             default='200',
             desc='The number of messages to fetch for each channel when fetching'
             ' history, between 1 and 1000.'),
+        'link_previews': Setting(
+            default='true',
+            desc='Show previews of website content linked by teammates.'),
         'map_underline_to': Setting(
             default='_',
             desc='When sending underlined text to slack, use this formatting'


devasmith:
I just tried your patch and it looks like it's working for me.

wavexx:
Thank you so much!

wohali:
Now let's just hope someone can merge this, like @trygveaa . :wink:

ta:
I'm also running wohali/526-toggle-link-previews now and order has been restored - mostly... :-) It seems to affect the Slack/GitHub integration as the GitHub user now posts empty messages in my weechat.

Has anyone else experienced this as well?

trygveaa:
@wohali: Great, thanks for this. I assume the check for `user` is to try to only disable link previews, not other attachments? Seems like that has some problems, according to @ta. I see the same with messages posted using giphy.

I think it might be better to check if the attachment has the `original_url` field. From what I've seen, that is only present for link previews. What do you think?

wohali:
@trygveaa Let's try it! Change pushed.

wee_slack.py
```diff
@@ -3778,6 +3778,8 @@ def unhtmlescape(text):
 
 
 def unwrap_attachments(message_json, text_before):
+    if 'original_url' in message_json and not config.link_previews:
```
trygveaa, line 3781: `original_url` is a field in the attachment itself, not in the message, so you would need to move it into the for loop below.

wohali: Whoops, fixed.

wee_slack.py
```diff
@@ -3791,6 +3791,8 @@ def unwrap_attachments(message_json, text_before):
             # $author: (if rest of line is non-empty) $title ($title_link) OR $from_url
             # $author: (if no $author on previous line) $text
             # $fields
+            if 'original_url' in attachment and not config.link_previews:
+               return ''
```
trygveaa, line 3795: We should do `continue` instead of `return`, in case there are more attachments.

trygveaa:
Thanks!

PR #822, opened 2021-01-17
Don't linkify messages sent with /me

### LINKED ISSUE #704: Sending `/me -> wfh` results in a message that says `nuxi -&gt; wfh`
If you do `/me -> wfh` then what actually gets displayed in wee-slack and the browser (or at least in Firefox) is `nuxi -&gt; wfh`. Although the android slack app will display `nuxi -> wfh`.

The same problems exist with `<` and `&` which show up as `&lt;` and `&amp;`. I didn't try any other HTML special characters.

This only seems to affect `/me` messages, the normal chat messages are fine.

hionay:
If messages are linkified when they are sent by /me, they are shown as HTML special characters.

Fixes #704

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1615,11 +1615,11 @@ class SlackChannelCommon(object):
                 self.print_getting_history()
 
     def send_message(self, message, subtype=None, request_dict_ext={}):
-        message = linkify_text(message, self.team)
         if subtype == 'me_message':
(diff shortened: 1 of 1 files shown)


gcg (approved):
lgtm

trygveaa:
Thanks, however there is a problem with doing this, in that nicks you include in the message won't be turned into proper user mentions anymore, so they won't get a notification.

As far as I can see from the docs and testing, we have to linkify nicks/channels normally, but without escaping `<>&`. I made a commit that does that.

PR #831, opened 2021-03-31
Update Docker image to use python3 package

shawntoffel:
The current Dockerfile fails to build as the latest alpine images no longer include a plain `python` package. This change installs the `python3` package instead.


trygveaa:
Thanks!

PR #832, opened 2021-04-09
fixed incorrect `ln` path in readme

CosmosAtlas:
In the current readme file. The `ln` incorrectly points to the parent directory. This pull requests updates the readme file to point to the correct file at `./wee_slack.py`


CosmosAtlas:
nvm, the readme is correct.

PR #843, opened 2021-06-16
Add a localvar with channel starred value.

KenMacD:
There's no way to set/unset this from the script, but it can be used in autosort rules. For example:

    ${info:autosort_order,${is_starred},1,0,}


trygveaa:
We should update this field when a channel is starred or unstarred, so the field doesn't just match the channels that were starred when wee-slack was loaded.

KenMacD:
Makes sense, I can look at adding that. Does it look okay otherwise?

trygveaa:
Yes, apart from that it looks good.

Issue #806, opened 2020-10-19
Subscribe to a message, that's not yet a tread


virus-found:
Hi.
It'd be lovely to be able to subscribe to a message, from which a thread hasn't yet been created.
Now I have to use an official client (three-dot-menu click on such a message --> "Follow Message, you will be notified of any replies")

Command example (`$9b3` can be taken from /cursor command):
```
/slack subscribe $9b3
```
Now it gives an obvious error due to it not being supported.
```
Error: Invalid id given, must be an existing id or a number greater than 0 and less than the number of messages in the channel
```

trygveaa:
So to allow `/slack (un)subscribe` for all messages we just have to remove the check for if it's a thread message, however there are some issues.

First is that I'm not sure we have a good way to show which messages you are subscribed to without querying them individually, so we can't show "Subscribed" after them like we do for thread parent messages. Though the official client doesn't show this until you click the menu for a specific message either. Not sure what a good UX (that is technically feasable) would be for this in weechat.

The other is that it might be hard to know if you are subscribed to the message or not when a new thread is created. There are three situations:
1. You subscribe to a message and a thread is created while wee-slack is running (and hasn't been restarted). This is fine, we get a websocket message when a message is subscribed to so we can keep track of that.
2. You subscribe to a message, quit wee-slack, a thread is created, and you start wee-slack afterwards. This is also fine, when a thread has been created it has a subscribed status when we load the history from the API.
3. You subscribe to the message, reload wee-slack and then a thread is created (or you quit, start it later again, but the thread is created while wee-slack is running). This is problematic. When we load the history there is as far as I can see no information about being subscribed to a message if it's not a thread, and when a thread is created the websocket messages doesn't contain info if we are subscribed or not. I guess we would have to query the API each time a new thread is created to fix this.

PR #845, opened 2021-07-06
/slack subscribe message_id

virus-found:
#806 

This implements the `we just have to remove the check for if it's a thread message` part from https://github.com/wee-slack/wee-slack/issues/806#issuecomment-713671364
The issues haven't been taken care of.


trygveaa (changes requested):
Well, the reason this check is there is because of the issues. By just removing the check, the user would get no feedback if the subscribe was successful, and I suspect this will confuse users.

I can see that it's useful if you're aware of the issues though, so maybe we can allow it if we describe it in the help text of `/subscribe`.

PR #853, opened 2021-08-21
Validate token not empty, before initiate connection

inomoz:
I continuously received this error, when started weechat:
```
AttributeError: 'NoneType' object has no attribute 'token'
```

Because tokens list was `['', '']`. And weechat tried initiate connections without tokens.

This small change fix this error.


trygveaa:
Thanks! Sorry for the long wait.

There was a formatting error, so I fixed that.

If you got this error it seems like you have two commas after each other in the `slack_api_token` so you could remove the extra comma.

PR #854, opened 2021-08-25
Fix handling of (multiline) messages from non-ASCII nick

bemoody:
An error occurs when:
- running on Python 2.7 (weechat 2.3 on Debian oldstable)
- someone's nick contains non-ASCII characters

Any messages from that person are not displayed, and the core buffer shows the error message:
```
python: wrong arguments for function "string_remove_color" (script: slack)
```

Apparently the reason for this is that on Python 2, `weechat.string_remove_color` fails when passed a Unicode string containing non-ASCII characters.

Moreover, if the string might contain combining characters, CJK, or emoji, then the `len()` of the string is not the same as its width.  `weechat.strlen_screen` should give the right answer, or at least weechat's best guess.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -263,7 +263,8 @@ class WeechatWrapper(object):
     # number of spaces, so it aligns correctly.
     def prnt_date_tags(self, buffer, date, tags, message):
         prefix, _, _ = message.partition("\t")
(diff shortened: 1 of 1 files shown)


trygveaa:
Ah, more issues with Python 2 :/

But thanks for the fix! And sorry for the long wait.

And good catch with length vs. width of strings. Didn't think of that when I added this.

Issue #551, opened 2018-04-09
Direct message not opening when receiving first message


lzmartinico:
When I receive a new message from a team member I have never talked to, the direct message channel isn't automatically opened

trygveaa:
By "not automatically opened", do you mean that the buffer isn't created, or that weechat doesn't switch to the buffer?

What should happen is that a buffer is created in the background, and the number that buffer gets is shown in the hotlist.

lzmartinico:
Sorry, I meant no buffer got created, or at least no number appeared in the hotlist

tardypad:
I just noticed the same issue myself.
I have a couple of new people that sent me direct messages but no buffer has been created.
I just found it out while opening slack in the browser. It made it look like I ignored those people I guess :/

IIRC the following works: if my weechat is open and a new person talks to me, then the buffer is created.
I believe that maybe what doesn't work then is to have the new buffers created after starting weechat itself.

antoyo:
Same issue here.
Could you please fix this issue?

trygveaa:
I'm not able to reproduce the issue. Can you give me some more details? Like does it happen for messages from all users or only some? If only some, is there any commonalities between those? Have you talked to these people previously? Do the buffers appear after you reload the script, or are they still missing?

Also, try to [enable debug mode](https://github.com/wee-slack/wee-slack#development) and see if any errors are printed in the core or debug buffers.

antoyo:
I can reproduce as such:
```
/join Slackbot
/slack slash /remind in 1 minute Test
/close
/quit
```
Start weechat after a couple of minutes and the buffer Slackbot won't be opened.
I will edit this messages to post the other messages you mentionned.

antoyo:
Are you able to reproduce with the above steps?

trygveaa:
Yeah, thanks, I can reproduce it with that. But I don't have time to look into it right now.

PR #859, opened 2022-04-05
Check closed IM channels for unread messages

danielnelson:
After connecting, IM channels that are open cause a buffer to be created, and any that are not open are skipped.  This results in direct messages being ignored when they have unread messages.

With this patch each closed IM channel is checked for unread messages and if any are found a buffer is created.

closes: #551


lifepillar:
Is there any progress on this? Missing direct messages is a serious deficiency.

trygveaa:
I'm very sorry for the long delay here. Thanks for the fix!

Issue #811, opened 2020-11-16
Print button url


lzmartinico:
When displaying a button block, a URL should be printed if present, so that it can be opened in a web browser.

trygveaa:
I [can't see this field documented](https://api.slack.com/legacy/interactive-message-field-guide#action_fields), but if visiting the URL indeed triggers the action, then yes we should print it.

lzmartinico:
This is again related to the Jitsi integration (as in issue #810): the /jitsi command invokes a [button](https://github.com/jitsi/jitsi-slack/blob/880bfa00908be7bdef1f9976b2be99399864ee0c/slack_messages.go#L33) that contains the url for a meeting; so by clicking on the button it triggers the "action" of joining a meeting. This is a pretty straightforward case, but I am not sure if it would apply to other types of buttons.

Issue #810, opened 2020-11-16
Parse actions outside blocks


lzmartinico:
Looking at the attachment parsing logic, it seems like actions are only processed within block. However, the Slack Jitsi extension contains an action outside a block ([here]( https://github.com/jitsi/jitsi-slack/blob/880bfa00908be7bdef1f9976b2be99399864ee0c/slack_messages.go#L22)). I don't know enough about the Slack API to say if this should be allowed behaviour or not, but I would argue that since there is at least one plugin in the while that uses it, it should be covered

trygveaa:
As far as I can see, actions outside of blocks in attachments are [not documented](https://api.slack.com/reference/messaging/attachments), but if it works in the official clients, we should add it to wee-slack too.

trygveaa:
Oh, it's [documented here](https://api.slack.com/legacy/interactive-message-field-guide#field-guide__top-level-message-fields__attachment-fields).

PR #863, opened 2022-05-19
feat(display): Add URL when slack block contain URL

lermit:
When slack return a button message type and a url field is present we
should output the URL.

Fix: #811 and #810


trygveaa:
Sorry for the late response here. Thanks for the fix!

PR #879, opened 2022-12-18
Update Python script instructions in README.md

### LINKED ISSUE #847: Autoload instructions should mention alternate path for XDG directories
In the README section for `2. Download wee_slack.py to ~/.weechat/python`, the title and body should mention that if the user's weechat installation is using XDG directories, the destination directory would be `/home/crash/.local/share/weechat/python` (and `/home/crash/.local/share/weechat/python/autoload`)

Reference from Weechat's documentation on these paths: https://weechat.org/files/doc/stable/weechat_user.en.html#xdg_directories

XDG paths are used in current Arch linux packages for WeeChat (3.2-1)

hkcountryman:
Closes #847 

I believe the WeeChat docs and default directories have changed since README.md was created.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,18 +94,20 @@ If `--with-python@2` was used, you should use: `sudo /usr/local/opt/python@2/bin
 
 ### 2. Download wee\_slack.py
 
-In order to [load this Python script in WeeChat](https://weechat.org/files/doc/stable/weechat_scripting.en.html#load_script), it should be located under the `python/` directory in the WeeChat data files directory. By default, [WeeChat uses `~/.local/share/weechat/python/`](https://weechat.org/files/doc/stable/weechat_user.en.html#weechat_directories). (These directories will be created the first time you run WeeChat, or you can create them yourself with the first command below.) Download the script to that location:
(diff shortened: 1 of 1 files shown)


trygveaa:
Thanks!

README.md
```diff
@@ -92,15 +92,23 @@ Setup
...
+### 2. Download wee\_slack.py
 
-If you don't want wee\_slack to start automatically when weechat starts, you can skip the last command.
+In order to [load this Python script in WeeChat](https://weechat.org/files/doc/stable/weechat_scripting.en.html#load_script), it should be located under the `python/` directory in [the WeeChat data files directory](https://weechat.org/files/doc/stable/weechat_user.en.html#files_and_directories), `$XDG_DATA_HOME/weechat/python`, or `$HOME/.local/share/weechat/python/` if `$XDG_DATA_HOME` is not defined or is empty. (These directories will be created the first time you run WeeChat, or you can create them yourself with the first command below.) Download the script to that location:
 
 ```
-mkdir -p ~/.weechat/python/autoload
-cd ~/.weechat/python
+mkdir -p $XDG_DATA_HOME/weechat/python/autoload/
+cd $XDG_DATA_HOME/weechat/python/
```
trygveaa, line 101: I think I'll change this to `.local/share/weechat` though, as I think most systems don't have `$XDG_DATA_HOME` set (and thus uses `.local/share/weechat`), and I want these commands to work verbatim for most people.

Could use a variable with a default value, but that's extra syntax not everyone will understand, plus it won't work in fish and other shells with different syntax from bash, so I don't think I'll do that.

PR #884, opened 2023-02-20
Firefox-snap and chrome/chrome-beta support

bogdando:
* Add support for cookies decryption for Chrome/beta
* Add support for firefox-snap, and multiple default profiles paths
* If no token can be found, like no key localConfig_v2 in webappsstore2, use fetched d/d-s values with a user-provided token

Not tested on Darwin, leaving this as an exercise for you


extract_token_from_browser.py
```diff
@@ -1,64 +1,176 @@
...
+       cookie_ds_value = cipher.decrypt(cookie_ds_value[3:]).decode('utf8')
+
+if cookie_ds_value:
+    cookie_value = f"d={cookie_d_value};d-s={cookie_ds_value}"
 else:
     cookie_value = cookie_d_value
 
 local_storage_path = default_profile_path.joinpath("webappsstore.sqlite")
-con = sqlite3.connect(f"file:{local_storage_path}?immutable=1", uri=True)
 local_storage_query = "SELECT value FROM webappsstore2 WHERE key = 'localConfig_v2'"
```
trygveaa, line 154: I see you haven't changed this. Does this mean that it will always fail for Chrome? I assume Chrome doesn't have the same structure for local storage as Firefox?

extract_token_from_browser.py
```diff
@@ -1,64 +1,176 @@
...
+from Crypto.Protocol.KDF import PBKDF2
+
+
+class AESCipher:
+
+    def __init__(self, key):
+        self.key = key
+
+    def decrypt(self, text):
+        cipher = AES.new(self.key, AES.MODE_CBC, IV=(' ' * 16))
```
trygveaa, line 25: `IV` has to be `bytes`, not `str` for this to work for me.

extract_token_from_browser.py
```diff
@@ -1,64 +1,176 @@
...
 import json
-from pathlib import Path
+import os
+import secretstorage
 import sqlite3
 import sys
 
+from pathlib import Path
+from Crypto.Cipher import AES
+from Crypto.Protocol.KDF import PBKDF2
```
trygveaa, line 16: This means pycryptodome (or pycrypto) has to be installed. Might want to add a try/except around it so the script still works for fetching from Firefox even without this library.

extract_token_from_browser.py
```diff
@@ -1,64 +1,176 @@
...
-)
-cookie_d_value = con.execute(cookie_d_query).fetchone()[0]
-cookie_ds_query = (
-    "SELECT value FROM moz_cookies WHERE host = '.slack.com' AND name = 'd-s'"
-)
-cookie_ds_values = con.execute(cookie_ds_query).fetchone()
-con.close()
+if args.browser in ["chrome", "chrome-beta"]:
+    bus = secretstorage.dbus_init()
+    collection = secretstorage.get_default_collection(bus)
```
trygveaa, line 128: I don't have a service providing `org.freedesktop.secrets` so this fails for me with this message:

```
Traceback (most recent call last):
  File "/usr/lib/python3.10/site-packages/secretstorage/util.py", line 48, in send_and_get_reply
    raise DBusErrorResponse(resp_msg)
jeepney.wrappers.DBusErrorResponse: [org.freedesktop.DBus.Error.ServiceUnknown] ('The name org.freedesktop.secrets was not provided by any .service files',)

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/home/trygve/dev/wee-slack/./extract_token_from_browser.py", l
ine 130, in <module>
    collection = secretstorage.get_default_collection(bus)
  File "/usr/lib/python3.10/site-packages/secretstorage/collection.py", line 177, in get_default_collection
    return Collection(connection)
  File "/usr/lib/python3.10/site-packages/secretstorage/collection.py", line 45, in __init__
    self._collection.get_property('Label')
  File "/usr/lib/python3.10/site-packages/secretstorage/util.py", line 67, in get_property
    (signature, value), = self.send_and_get_reply(msg)
  File "/usr/lib/python3.10/site-packages/secretstorage/util.py", line 58, in send_and_get_reply
    raise SecretServiceNotAvailableException(data) from resp
secretstorage.exceptions.SecretServiceNotAvailableException: The name org.freedesktop.secrets was not provided by any .service files
```

Preferably, this should be handled gracefully, and we should try `peanuts` as the default password (as that's what Chrome uses by default).

extract_token_from_browser.py
```diff
@@ -1,64 +1,176 @@
...
 if sys.platform.startswith("linux"):
-    firefox_path = Path.home().joinpath(".mozilla/firefox")
+    iterations = 1
+    if args.browser == "firefox-snap":
+        browser_data = Path.home().joinpath(
+            "snap/firefox/common/.mozilla/firefox")
+    elif args.browser == "firefox":
+        browser_data = Path.home().joinpath(".mozilla/firefox")
+    else:
+        browser_data = Path.home().joinpath(".config/google-%s" % args.browser)
```
trygveaa, line 59: Can also add support for chromium with the path `.config/chromium`.

extract_token_from_browser.py
```diff
@@ -1,64 +1,176 @@
...
+        profile = "*.default*"
+
+    default_profile_path = max(
+        [ x for x in browser_data.glob(profile)], key=os.path.getctime)
+    if not default_profile_path:
+        print("Couldn't find the default profile for Firefox", file=sys.stderr)
+        sys.exit(1)
+else:
+    # b'v10' is for Chromium, but not Chrome, it seems?
+    prefix =  b'v11'
```
trygveaa, line 94: As you write, this is `v10` for Chromium (at least on my machine). We should check for both prefixes.

trygveaa (commented):
Thanks. I tested this on Chromium on my machine, and it works for me after some changes (see below comments). I just got the cookie though, not any tokens, see the below comment regarding if it can get tokens from Chrome.

> Not tested on Darwin, leaving this as an exercise for you

I don't have any mac, so can't test it :) As far as I know secretstorage won't work on mac though, so we should probably fail gracefully or print an error message there.

> Related https://github.com/n8henrie/pycookiecheat/issues/18

Hm, how is that issue related?

extract_token_from_browser.py
```diff
@@ -1,64 +1,176 @@
...
-con.close()
+teams = []
+try:
+    con = sqlite3.connect(f"file:{local_storage_path}?immutable=1", uri=True)
+    local_config_str = con.execute(local_storage_query).fetchone()[0]
+    local_config = json.loads(local_config_str)
+except Exception:
+    if not args.token:
+        print("Couldn't find the token for team, extract it manually "
+              "and specify it in --token", file=sys.stderr)
```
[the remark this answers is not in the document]
bogdando: good point, will do that

extract_token_from_browser.py
bogdando: it wasn't failing for Chrome, just returning empty results

extract_token_from_browser.py
bogdando: I havent' planned this PR for Chromium support, could be done as follow-ups if anyone willing to catch up?

extract_token_from_browser.py
bogdando: maybe later, whenever we open another one for Chromium support

extract_token_from_browser.py
trygveaa: Well, apparently this file exists, but it's empty for me at least. I don't know why Chrome has this empty file, but I'm pretty sure trying to read it is useless. Just getting the token and not the Cookies out of Chrome feels pretty pointless to me as you would still have to go to the browser to get out the tokens (and I'd argue that it's more work/less obvious how to get out the tokens than the cookie).

I checked a bit though and see that we can read out the token from Chromes local storage. It can be done like this (for Chromium, but Chrome is probably the same just with a different path):

```
import json
import plyvel

db = plyvel.DB('/home/<username>/.config/chromium/Default/Local Storage/leveldb')
local_storage_value = db.get(b'_https://app.slack.com\x00\x01localConfig_v2')
slack_config = json.loads(local_storage_value[1:])
```

This requires the [Plyvel](https://plyvel.readthedocs.io/en/latest/) library which we can conditionally import when running for Chrome like Pycryptodome. Note that Chrome has to be closed it to be able to read the library. We should gracefully handle the error when Chrome is open.

extract_token_from_browser.py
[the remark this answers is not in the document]
bogdando: Note that even with token extracted manually as ``window.prompt("Session token:", TS.boot_data.api_token)``, this script still provides value as it prepares the resulting command and saves a few extra steps :)

bogdando:
Works for me with chrome now, without manual extraction of tokens (it extracted data for all my slack teams, I was surprised to recall I have a bunch more of it). Many thanks, @trygveaa !

trygveaa:
Sorry, I completely forgot about this. I merged it now and pushed some additional fixes. Thanks!

PR #885, opened 2023-02-22
feat(huddle): Display link to join huddle when started

lermit:
This PR display a link when a huddle is started.

Have a nice day :)

Regards

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -4791,11 +4791,8 @@ def unwrap_huddle(message, message_json):
 
     if "room" in message_json:
         for block in message_json.get("blocks"):
-            print(block)
             for element in block.get("elements"):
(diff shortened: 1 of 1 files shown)


trygveaa:
Thanks! Seems to work good.

I see that after everyone has left the huddle the message is normally deleted, but I noticed that if you set a topic on the huddle (and possibly on other conditions, I didn't test that much), the huddle message won't be deleted, so it will still just say "A huddle started" even though no one is in it. I see that if you fetch the message history at this time, the room will have `date_end` set to the timestamp it was ended, so we should print a different text in that case. We don't get a message changed even for this though, but we get a `sh_room_update` event so looks like we have to look through the messages for the huddle message and update it based on this event.

I'm leaving that for a future improvement though, so merging this now.

PR #887, opened 2023-03-03
extract_token_from_browser: widen default profile check

frozencemetery:
On the Linux machines I've used, the default profile is suffixed .default, not .default-release.  I'm not sure what systems use .default-release, but preserving it for compatibility seems harmless.

This is actually not enough to get it working for me.  It bails later on: webappsstore.sqlite doesn't have anything with key "localConfig_v2", or anything that even looks like it.  System is firefox 110 on Debian testing.

### DIFF
--- a/README.md
+++ b/README.md
@@ -207,6 +207,8 @@ get the tokens and cookies for all the teams you're logged into:
 ./extract_token_from_browser.py firefox
 ```
 
+(Note this script requires the python3-snappy library.)
+
 #### Optional: Connecting to multiple teams
 
 You can run the register command multiple times to connect to multiple teams.
--- a/extract_token_from_browser.py
+++ b/extract_token_from_browser.py
@@ -4,6 +4,7 @@ import argparse
 from configparser import ConfigParser
 import json
 from pathlib import Path
+from snappy import snappy
 import sqlite3
 import sys
 
@@ -62,11 +63,22 @@ if cookie_ds_values:
 else:
     cookie_value = cookie_d_value
 
-local_storage_path = default_profile_path.joinpath("webappsstore.sqlite")
-con = sqlite3.connect(f"file:{local_storage_path}?immutable=1", uri=True)
-local_storage_query = "SELECT value FROM webappsstore2 WHERE key = 'localConfig_v2'"
-local_config_str = con.execute(local_storage_query).fetchone()[0]
+storage_path = default_profile_path.joinpath(
+    "storage/default/https+++app.slack.com/ls/data.sqlite"
+)
+con = sqlite3.connect(f"file:{storage_path}?immutable=1", uri=True)
+storage_query = "SELECT compression_type, conversion_type, value FROM data WHERE key = 'localConfig_v2'"
+is_compressed, conversion, payload = con.execute(storage_query).fetchone()
 con.close()
+
+if is_compressed:
+    payload = snappy.decompress(payload)
+
+if conversion == 1:
+    local_config_str = payload.decode("utf-8")
+else:
+    # untested; possibly Windows-only?
+    local_config_str = payload.decode("utf-16")
 
 local_config = json.loads(local_config_str)
 teams = [
--- a/extract_token_from_browser.py
+++ b/extract_token_from_browser.py
@@ -14,6 +14,9 @@ parser = argparse.ArgumentParser(
 parser.add_argument(
     "browser", help="Which browser to extract from", metavar="<browser>"
 )
+parser.add_argument(
+    "--profile-path", help="Skip detection and use this Firefox profile"
+)
 args = parser.parse_args()
 
 if args.browser != "firefox":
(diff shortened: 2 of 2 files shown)


trygveaa:
> On the Linux machines I've used, the default profile is suffixed .default, not .default-release. I'm not sure what systems use .default-release, but preserving it for compatibility seems harmless.

Since it didn't work further on, have you verified that the `cookies.sqlite` file in `.default` has the Slack cookie? I also have directory suffixed with `.default`, but the `cookies.sqlite` file in it is empty, while the `cookies.sqlite` file in the directory suffixed with `.default-release` has it for me.

I see that I have a file `~/.mozilla/firefox/profiles.ini` with some info about the default profile. It contains this for me:

```
[Install4F96D1932A9F858E]
Default=0z063xu1.default-release
Locked=1

[Profile1]
Name=default
IsRelative=1
Path=4yyzgkj6.default
Default=1

[Profile0]
Name=default-release
IsRelative=1
Path=0z063xu1.default-release

[General]
StartWithLastProfile=1
Version=2
```

Maybe we can use this to pick the correct directory. Do you also have this file? What does it contain for you?

> This is actually not enough to get it working for me. It bails later on: webappsstore.sqlite doesn't have anything with key "localConfig_v2", or anything that even looks like it. System is firefox 110 on Debian testing.

Hm, I'm on Arch Linux with Firefox installed from the Arch Linux repo, also in version 110, and this is where it stores it for me. `localConfig_v2` is the name Slack uses for the local storage object with the token, so that wouldn't be dependent on the OS, so it sounds like you have the wrong local storage file. Do you have any rows with `originKey = 'moc.kcals.ppa.:https:443'` (or any rows at all)?

(Also, I see that we're not checking the originKey of the local storage object, so we're looking for it under any domain, which we shouldn't do, but not related to the PR changes of course, just something I noticed now.)

frozencemetery:
> Since it didn't work further on, have you verified that the cookies.sqlite file in .default has the Slack cookie? I also have directory suffixed with .default, but the cookies.sqlite file in it is empty, while the cookies.sqlite file in the directory suffixed with .default-release has it for me.

Yes:

```python
>>> import sqlite3
>>> con = sqlite3.connect("file:/home/rharwood/.mozilla/firefox/uvrq3ve3.default/cookies.sqlite?immutable=1", uri=True)
>>> con.execute("SELECT value FROM moz_cookies WHERE host = '.slack.com' AND name = 'd'").fetchone()[0][:5]
'xoxd-'
```

> Maybe we can use this to pick the correct directory. Do you also have this file? What does it contain for you?

We could.  Mine has:

```ini
[Install4F96D1932A9F858E]
Default=uvrq3ve3.default
Locked=1

[Profile0]
Name=default
IsRelative=1
Path=uvrq3ve3.default
Default=1

[General]
StartWithLastProfile=1
Version=2

```

> Do you have any rows with originKey = 'moc.kcals.ppa.:https:443'

No:

```python
>>> con = sqlite3.connect("file:/home/rharwood/.mozilla/firefox/uvrq3ve3.default/webappsstore.sqlite?immutable=1", uri=True)
>>> con.execute("SELECT key FROM webappsstore2 where originKey = 'moc.kcals.ppa.:https:443'").fetchall()
[]
```

> (or any rows at all)?

Yes, plenty:

```python
>>> len(con.execute("SELECT key FROM webappsstore2").fetchall())
8092
```

grep suggests that the data we're might in storage/default/https+++app.slack.com/ls/data.sqlite (inside the profile directory).  "SELECT value FROM data WHERE key = 'localConfig_v2'" gives back a thing that looks mostly like json, but it's not encoded right somehow: json.loads() chokes on utf-8 errors.  But this works for me to get the token:

```python
>>> import sqlite3
>>> con = sqlite3.connect("file:/home/rharwood/.mozilla/firefox/uvrq3ve3.default/storage/default/https+++app.slack.com/ls/data.sqlite?immutable=1", uri=True)
>>> payload = con.execute("SELECT value FROM data WHERE key = 'localConfig_v2'").fetchone()[0]
>>> token = payload[payload.index(b"xoxc-"):]
>>> token = token[:token.index(b'"')]
>>> len(str(token))
111
```

I don't know whether this is a reasonable thing to do generally.  I'm guessing you don't have a file like that?

frozencemetery:
According to https://support.mozilla.org/gl/questions/1264072 the presence of a .default means that the firefox profile data I'm using predates version 67; I guess they don't change it on upgrade, but only create .default-release now.

trygveaa:
> grep suggests that the data we're might in storage/default/https+++app.slack.com/ls/data.sqlite (inside the profile directory). "SELECT value FROM data WHERE key = 'localConfig_v2'" gives back a thing that looks mostly like json, but it's not encoded right somehow: json.loads() chokes on utf-8 errors. But this works for me to get the token:

Ah, this is apparently the new place this data is stored, while `webappsstore.sqlite` is the old way. There's some description in https://searchfox.org/mozilla-central/source/dom/localstorage/ActorsParent.cpp:

```
/**
 * The database name for LocalStorage data in a per-origin directory.
 */
constexpr auto kDataFileName = u"data.sqlite"_ns;
```

```
/**
 * The legacy LocalStorage database.  Its contents are maintained as our
 * "shadow" database so that LSNG can be disabled without loss of user data.
 */
#define WEB_APPS_STORE_FILE_NAME u"webappsstore.sqlite"
```

I have both files with data. Though, I tried to create a new local storage value, and see that that was only stored to `data.sqlite`. And it looks like data is migrated to the new format when accessing the page. So we should be reading from this file instead of `webappsstore.sqlite`.

I see that it may be compressed with snappy which was why you saw partly binary data. We can decompress this with the [python-snappy library](https://pypi.org/project/python-snappy/). There is a `compression_type` column which is set to `1` if it's compressed and `0` if it isn't. There is also a `conversion_type` column. The data is UTF-8 if this is set to `1`, and I think it's UTF-16 if it's set to `0`, but I don't have any rows on any pages where it's set to `0` so I can't confirm that.

> According to https://support.mozilla.org/gl/questions/1264072 the presence of a .default means that the firefox profile data I'm using predates version 67; I guess they don't change it on upgrade, but only create .default-release now.

Ah, okay, then it would make sense to try `.default-release` first and then `.default`. But I think might be preferable to read it out of `profiles.ini`.

frozencemetery:
Awesome.  With those suggestions, I've updated this PR to something that works for me.

The parsing of profiles.ini is a bit sketchy since none of the section keys are fixed.  Based on your config, it seems that some profile entries can have `Default=1` set without actually being the default; that seems configured by the first section which starts with "Install".

trygveaa:
> Awesome. With those suggestions, I've updated this PR to something that works for me.

Great

> The parsing of profiles.ini is a bit sketchy since none of the section keys are fixed. Based on your config, it seems that some profile entries can have `Default=1` set without actually being the default; that seems configured by the first section which starts with "Install".

I read https://support.mozilla.org/en-US/kb/understanding-depth-profile-installation now. So `Default=1` under the profile is the default profile prior to Firefox 67, but with newer versions it uses the Default under the Install section instead.

Instead of looking through all sections, we should just look through the sections starting with `Install`, and if there's one such section, we should use the `Default` from that.

If there's zero or more than one Install section, I think we should print an error message saying the profile path has to be specified manually. And add a `--profile-path` option to allow specifying that.

frozencemetery:
Makes sense.  Updated.

trygveaa:
Sorry, I completely forgot about this. I merged it now. Thanks!

PR #894, opened 2023-07-07
docker: Update script path for wee_slack.py

j6k4m8:
I think the path has changed in the intervening years since this dockerfile was last edited; this new fix appears to work just fine in my tests.


trygveaa:
Ah, thanks. Yes, this is the new default path.

PR #896, opened 2023-07-27
feat: Filter channel list based on regular expression

bmasonrh:
It can be difficult to find a specific channel if there are thousands of them in your slack workspace.  This enhancement allows you to filter the channels based on a regular expression.  For example:

`/slack channels support`

returns
```
Channels that match "support":
    #customer-support    (not a member)
    #it-support-team     (not a member)
    #technnical-support  (member)
```


trygveaa:
Thanks!

PR #902, opened 2023-09-21
Make render_emoji_as_string = "both" work with rich_text blocks

melwitt:
Commit 74da30342e8d328fc085d2158e37afda3b908f2c added support for using rich_text blocks to format messages. With this change, the use of replace_string_with_emoji() for text type "emoji" can result in running the replacement twice as replace_string_with_emoji() is also called as one of the last steps in SlackMessage.render().

This works fine in the default case when render_emoji_as_string = "", however render_emoji_as_string = "both" results in output like:
  
  `😄 (😄 (:smile:))`

instead of the intended:

  `😄 (:smile:)`

This skips the second call of replace_string_with_emoji() if blocks were rendered earlier in the method.

### DIFF
--- a/_pytest/conftest.py
+++ b/_pytest/conftest.py
@@ -152,6 +152,8 @@ class FakeWeechat:
     def info_get(self, info_name, arguments):
         if info_name == "color_rgb2term":
             return arguments
+        elif info_name == "weechat_data_dir":
+            return "."
         else:
             return ""
 
--- a/_pytest/test_render_message.py
+++ b/_pytest/test_render_message.py
@@ -1,0 +1,100 @@
+import pytest
+
+import wee_slack
+
+
+@pytest.mark.parametrize(
+    "case",
+    [
+        {
+            "input_message": {
+                "blocks": [
+                    {
+                        "type": "rich_text",
+                        "block_id": "5Cg6",
+                        "elements": [
+                            {
+                                "type": "rich_text_section",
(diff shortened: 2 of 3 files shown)


trygveaa:
Ah, I hadn't noticed that. This patch makes emojis not rendered at all in e.g. reactions and attachments though.

melwitt:
> Ah, I hadn't noticed that. This patch makes emojis not rendered at all in e.g. reactions and attachments though.

And I didn't notice that 🤦 I'll look into it more. Thanks for reviewing.

trygveaa:
Probably easiest to just put `replace_string_with_emoji` around everything that is added to `text` except blocks. This code is sort of a mess and is rewritten in the `3.0.0` branch anyways, so I don't care that much about it being fixed in a nice/clean way here.

melwitt:
> Probably easiest to just put `replace_string_with_emoji` around everything that is added to `text` except blocks. This code is sort of a mess and is rewritten in the `3.0.0` branch anyways, so I don't care that much about it being fixed in a nice/clean way here.

Fair enough 😄 I'm going to try something else first and if there's something else I missed, I can just do the put `replace_string_with_emoji` everywhere instead.

trygveaa:
I ran black to fix the formatting and pushed it here. I also simplified the way the emojis in the test are written.

trygveaa:
Thanks!

PR #909, opened 2023-10-16
Handle Firefox containers in extract_token_from_browser.py

tohojo:
The extract_token_from_browser.py script does not handle Firefox containers
correctly. The container support works by namespacing cookies and local storage
with a 'userContextId'. In effect this means that when containers are used,
running the script will select the cookie from whichever container happens to be
returned first by the database query, which will be combined with the
non-containerised instance of the local storage.

Add correct handling by adding a new --container flag which specifies a
container either by numeric ID or by name, and including the correct
userContextId into the cookie query and local storage path (or excluding all
cookies from containers from the query if no --container flag is supplied).

With this the script can be run like:

python extract_token_from_browser.py --container Work firefox

to get the Slack auth token for the 'Work' container.

### DIFF
--- a/extract_token_from_browser.py
+++ b/extract_token_from_browser.py
@@ -39,10 +39,12 @@ def sqlite3_connect(path: StrPath):
         con.close()
 
 
-def get_cookies(cookies_path: StrPath, cookie_query: str) -> tuple[str, str | None]:
+def get_cookies(
+    cookies_path: StrPath, cookie_query: str, params: tuple
+) -> tuple[str, str | None]:
     with sqlite3_connect(cookies_path) as con:
-        cookie_d_value = con.execute(cookie_query.format("d")).fetchone()
-        cookie_ds_value = con.execute(cookie_query.format("ds")).fetchone()
+        cookie_d_value = con.execute(cookie_query.format("d"), params).fetchone()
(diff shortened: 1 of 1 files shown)


extract_token_from_browser.py
```diff
@@ -147,13 +150,35 @@ def get_cookies(cookies_path: StrPath, cookie_query: str) -> tuple[str, str | No
...
+                      file=sys.stderr)
+                sys.exit(1)
+
+        userctx = f"^userContextId={ctx_id}"
+    else:
+        userctx = ""
+
     cookie_query = (
-        "SELECT value FROM moz_cookies WHERE host = '.slack.com' " "AND name = '{}'"
+        f"SELECT value FROM moz_cookies WHERE originAttributes='{userctx}' AND host = '.slack.com' " "AND name = '{}'"
```
trygveaa, line 176: Since this is just invoked manually by a user on their local machine, I guess SQL injections aren't a concern, but I would prefer this to be an SQL parameter instead of including it directly in the string nevertheless.

trygveaa (changes requested):
Thanks! I just have a small thing I want you to change (see the below comment). And please run black to format the file.

extract_token_from_browser.py
tohojo: Sure, can do!

PR #915, opened 2023-12-29
Add `/slack reload_auth` to refresh auth tokens without reloading the whole script

devkev:
Large orgs can take a long time for wee-slack to (re)load (eg. 1 hour).  If the session auth length is also short (eg. 1 week or less), then having to fully restart wee-slack that often gets pretty disruptive (having to get the session token out of the browser is bad enough).

This patch adds a command that simply re-reads the token config from weechat, and then plumbs the new value into the SlackTeam object.  Thus when your session token has expired, and you are getting lots of `invalid_auth` errors in the wee-slack console, you can do

```
/set plugins.var.python.slack.slack_api_token <newtoken>
/slack reload_auth
```
or
```
/secure set slack_token <newtoken>
/slack reload_auth
```
and then the currently running wee-slack script will be able to auth successfully on the next attempt, no `/python reload slack` required.

If you're in multiple teams, then a little more care is needed, you have to update the token list in the weechat config carefully, and note the index of which one has changed (zero-based).  Then, from a weechat buffer in that team, do
```
/slack reload_auth <n>
```
where `<n>` is the index number of the appropriate entry in the token string.

Thoughts?  The interface around multiple teams and specifying `<n>` is a little clunky, and could probably be better, but I wanted to check that the approach even works in the first place (it seems to work ok for me), and I figure someone more familiar with wee-slack would have a better idea of how/if this aspect could be improved.  But this seems like a decent base to start from, at the very least.

### DIFF
--- a/wee_slack.py
+++ b/wee_slack.py
@@ -1674,6 +1674,9 @@ class SlackTeam(object):
     def generate_usergroup_map(self):
         return {s.handle: s.identifier for s in self.subteams.values()}
 
+    def set_token(self, token):
+        self.token = token
+
     def set_name(self):
         alias = config.server_aliases.get(self.subdomain)
         if alias:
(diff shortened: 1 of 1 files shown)


trygveaa:
Yes, I agree that having to reload the script is inconvenient, but as you saw it get's a bit complicated since all the tokens are stored in one option.

To improve the configuration (and other things), I'm working on a new version which uses a similar config setup as the IRC plugin. So it has one config section per Slack workspace and you set the token separately for each workspace. This version also reads the config values from WeeChat each time they are used, so it doesn't require you to reload the script on changes, it's enough to just set the option. I have pushed it to the [3.0.0 branch](https://github.com/wee-slack/wee-slack/tree/3.0.0). It's still missing some features, but should be usable.

So I'm a bit torn on including this or not. I don't really like having a command like that, and it's clunky as you said, but I don't have any other suggestions for improvement, apart from how I have done it in the 3.0.0 branch. But on the other hand it's useful until that branch is ready, and I'm not sure when that will be.

devkev:
That sounds like a much better way forward, I'm glad to hear about it.  Given that, I think it doesn't make sense to merge this, and then have to deal with eg. users being confused later when the command is gone.  If this PR is left here (until 3.0.0 is released) then anyone affected by this problem can apply it and use the dodgy command in the meantime.

Issue #910, opened 2023-10-19
ERROR: Failed connecting to Slack with token ...: channels: enterprise_is_restricted


yazgoo:
```
ERROR: Failed connecting to Slack with token <...>: channels: enterprise_is_restricted, presence: enterprise_is_restricted, channels: enterprise_is_restricted
```

Seems the same as

https://github.com/yuya373/emacs-slack/issues/543

trygveaa:
What kind of token do you use? (Which four characters does it start with)

github-olivier-abdesselam:
I have

```
python.slack.slack_api_token = xoxc-.....:xoxd-....
```

The error is:

```
ERROR: Failed connecting to Slack with token xoxc-<start of xoxc token>...<end of xoxd token>: channels: enterprise_is_restricted, presence: enterprise_is_restricted, channels: enterprise_is_restricted
```

yazgoo:
I have

```
python.slack.slack_api_token = xoxc-.....:xoxd-....
```

The error is:

```
ERROR: Failed connecting to Slack with token xoxc-<start of xoxc token>...<end of xoxd token>: channels: enterprise_is_restricted, presence: enterprise_is_restricted, channels: enterprise_is_restricted
```

trygveaa:
Hm, okay, so the workspace apparently restricts the usage of some API methods. Unfortunately it's impossible for me to know which API methods I can use in this case, and which I could potentially replace them with since I don't have any such restricted workspaces.

tokyo-kinney:
The updated Events API includes authorizations inside the event payload on most events.

{
    "token": "XXYYZZ",
    "is_ext_shared_channel": false,
    "team_id": "TXXXXXXXX",
    "api_app_id": "AXXXXXXXXX",
    "event": {...},
    "type": "event_callback",
    "event_id": "EvXXXXXXXX",
    "event_time": 1234567890,
    "authorizations": [
        {
            "enterprise_id": "E324567",
            "team_id": "T423567",
            "user_id": "W43567",
            "is_bot": false,
            "is_enterprise_install": true
        }
    ]
}
So maybe the team_id also needs to be included?

pierg75:
I had a go at debugging this. From the logs I can see three calls which got `enterprise_is_restricted`:

```
{"ok": false, "error": "enterprise_is_restricted", "wee_slack_process_method": "conversationslist"}

[...]

PR #916, opened 2023-12-29
WIP: fix enterprise

yazgoo:
aims at fixing 
https://github.com/wee-slack/wee-slack/issues/910


yazgoo:
what doesn't work yet

- [ ] only list  one channel initially, other channels appear when receiving messages
- [ ] getting exception below when switching buffer

``
python: stdout/stderr (slack):   File "/home/yazgoo/.weechat/python/autoload/wee_slack.py", line 5329, in set_own_presence_active
python: stdout/stderr (slack):     slackbot = team.get_channel_map()["Slackbot"]
python: stdout/stderr (slack):                ~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^
```

File: wee_slack.py [hunk 1]
```diff
@@ -2348,10 +2348,8 @@ def check_should_open(self, force=False):
             return
 
         if (
-            getattr(self, "is_open", False)
-            or self.unread_count_display
+            self.unread_count_display
             or self.type not in ["im", "mpim"]
-            and getattr(self, "is_member", False)
```
yazgoo, line 2354: attribute is not available

wee_slack.py
```diff
@@ -2579,7 +2577,7 @@ def get_members(self):
             # complete, so we have to fetch members for all channels.
             s = SlackRequest(
                 self.team,
-                "conversations.members",
+                "membership",
```
trygveaa, line 2580: Is `membership` a method that works in your Slack workspace? I just get `unknown_method` from it. You also haven't handled the response anywhere, so making this request won't have any effect.

wee_slack.py
```diff
@@ -7248,9 +7236,10 @@ def handle_getPresence(response_json, eventrouter, team, channel, metadata):
         callback=handle_prefs,
     )
     EVENTROUTER.receive(s)
+
     s = SlackRequest(
         None,
-        "users.getPresence",
+        "users.list",
```
trygveaa, line 7242: Not sure why you're using `users.list` here, as that method doesn't return the info we need here (`manual_away`)?

trygveaa (commented):
Ah, I actually started experiencing this problem too a short while ago so I've also worked on fixing it. I have a new version of wee-slack I've been working on a while, so I made the changes there rather than on master. It's a large rewrite of the current version and not all features are available yet, but it should be usable. I pushed the latest changes with the fix for this issue to the [3.0.0 branch](https://github.com/wee-slack/wee-slack/tree/3.0.0) now.

So I think I would rather spend time fixing the remaining things on that branch, rather than fixing this issue on master. If we were to continue here, we would at least have to make sure it doesn't break other workflows (like oauth tokens) before merging this. Nevertheless, I added a few questions to the changes you did.

PR #919, opened 2024-01-17
Set user user presence

pierg75:
The user can now set his presence with:

/slack away

and back with:

/slack back

It's not possible to set a custom status for now.

### DIFF
--- a/slack/commands.py
+++ b/slack/commands.py
@@ -287,6 +287,20 @@ def command_slack_reply(buffer: str, args: List[str], options: Options):
         run_async(slack_buffer.post_message(split_args[1], thread_ts, broadcast))
 
 
+@weechat_command("away|active")
+def command_slack_presence(buffer: str, args: List[str], options: Options):
+    slack_buffer = shared.buffers.get(buffer)
+    if slack_buffer is None:
+        return
+    new_presence = args[0]
+    if new_presence not in ("active", "away"):
+        print_error(
+            f'Error with command "/slack presence {args[0]}" (help on command: /help slack presence)'
+        )
+        return
+    run_async(slack_buffer.workspace.api.set_presence(new_presence))
+
+
 def print_uncaught_error(error: UncaughtError, detailed: bool, options: Options):
     weechat.prnt("", f"  {error.id} ({error.time}): {error.exception}")
     if detailed:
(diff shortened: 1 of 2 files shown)


trygveaa (changes requested):
I think I would prefer having a single `/slack presence` command instead of separate away and back commands since they both change the same thing. This would also give them a single help entry (when help texts are implemented). And I think we should use the term `active` like Slack does, instead of `back`. So `/slack presence away` and `/slack presence active`.

Can you also format the code with ruff?

slack/slack_api.py
```diff
@@ -412,3 +412,12 @@ async def reactions_change(
         if response["ok"] is False:
             raise SlackApiError(self.workspace, method, response, params)
         return response
+
+    async def set_away(self, presence: str):
```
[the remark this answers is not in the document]
pierg75: Thanks for the feedback, would this new version work?

slack/commands.py
```diff
@@ -287,6 +287,13 @@ def command_slack_reply(buffer: str, args: List[str], options: Options):
         run_async(slack_buffer.post_message(split_args[1], thread_ts, broadcast))
 
 
+@weechat_command("away|active")
+def command_slack_presence(buffer: str, args: List[str], options: Options):
+    slack_buffer = shared.buffers.get(buffer)
+    if slack_buffer:
+        run_async(slack_buffer.workspace.api.set_presence(args[0]))
```
trygveaa, line 294: You have to check that `args[0]` is either `active` or `away`, and otherwise print an error.

The error message should be in this format (because that's what WeeChat's built in commands print when you give an unrecognized argument):

```
Error with command "/slack presence asd" (help on command: /help slack presence)
```

trygveaa, line 294: So like this:

```suggestion
    if slack_buffer is None:
        return

    new_presence = args[0]
    if new_presence != "active" and new_presence != "away":
        print_error(
            f'Error with command "/slack presence {args[0]}" (help on command: /help slack presence)'
        )
        return

    run_async(slack_buffer.workspace.api.set_presence(new_presence))
```

pierg75: Thanks! Hopefully this is the right one.

trygveaa: I pushed a change to use the `new_presence` variable in the call to `set_presence`, otherwise the type narrowing doesn't (which is why the CI build failed).

trygveaa:
Thanks! And sorry for the delay in merging this.

pierg75:
> Thanks! And sorry for the delay in merging this.

No worries and thank you!

PR #926, opened 2024-05-03
Make logging "Channel Created" optional

oalbrigt:


trygveaa:
Sorry for the delay here.

I changed the default value to match the current behavior and added the option to `Options.md`. I also changed it to the `2.x` branch, as I'm moving development of the current version there (`master` is now the upcoming 3.0.0 version, which is a mostly complete rewrite).

Thanks!

PR #929, opened 2024-06-07
feat: Filter users list based on regular expression

bmasonrh:
It can be difficult to find a specific user when there are tens of thousands of users in your slack team.  This enhancement allows you to filter users based on a case-insensitive regular expression.  For example:

```
/slack users ^.Mason

Users that match "^.Mason"

    bmason  (active)
    jmason  (inactive)
    lmason (active)
    SMason  (inactive)
```

### DIFF
--- a/docs/Commands.md
+++ b/docs/Commands.md
@@ -240,8 +240,9 @@ If handle is given show the members in the usergroup
 ### users
 
 ```
-/slack users
+/slack users [regex]
 ```
 
 List the users in the current team.
+If regex is given show only users that match the case-insensitive regex.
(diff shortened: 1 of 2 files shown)


trygveaa:
Sorry for the delay here.

I pushed a fix for the behavior without any arguments, as `pat` would never be `None` so it printed `Users that match "":`. I also updated`Options.md` and changed it to the `2.x` branch, as I'm moving development of the current version there (`master` is now the upcoming 3.0.0 version, which is a mostly complete rewrite).

Thanks!

PR #931, opened 2024-07-15
conversation: show user real name in title

rjarry:
When opening a private conversation, show the real name of the user in the status bar instead of an empty status.


trygveaa:
Not sure about only showing the real name if a topic isn't set. What about always showing real name, e.g. `<real name> | <status> | <topic>`?

rjarry:
Yes, that works for me. I'll change that.

trygveaa:
Thanks!

PR #932, opened 2024-07-15
message: use nick color for user mentions

rjarry:
Instead of using the same color for all user mentions, reuse the mentioned user nick color.

### DIFF
--- a/slack/slack_message.py
+++ b/slack/slack_message.py
@@ -271,7 +271,9 @@ class PendingMessageItem:
 
             if self.display_type == "mention":
                 name = f"@{user.nick.format()}"
-                return with_color(shared.config.color.user_mention.value, name)
+                return with_color(
+                    user.nick.color or shared.config.color.user_mention.value, name
+                )
             elif self.display_type == "chat":
                 return user.nick.format(colorize=True)
             else:


trygveaa:
Hm, this means the `user_mention` option will basically never be used. Also, with the default colors it will show your own nick in white which is pretty similar to the default text color in most dark terminals. Maybe we should add an option to determine if the nick color or the `user_mention` option should be used?

rjarry:
I was trying to mimick what weechat is doing for IRC buffers. User mentions always use the user's nick color so that you can visually identify who's mentioned based on the other lines in the context. Maybe this is due to my specific configuration.

I was surprised that there was a setting to force all user mentions to a specific color. How about falling back to the nick color if `user_mention` is unset or forcibly set to an empty string?

rjarry:
Side note, the `slack.color.user_mention` setting is new (i.e. only in wee-slack 3.0.0 rework).

rjarry:
About your remark on your own nick being the same color than default text. You should have a highlight rule that makes that particular message highly visible anyway. So the mention color does not matter.

trygveaa:
> I was trying to mimick what weechat is doing for IRC buffers. User mentions always use the user's nick color so that you can visually identify who's mentioned based on the other lines in the context. Maybe this is due to my specific configuration.

WeeChat doesn't do that by default for IRC buffers. By default they're printed without any color. I assume you're using the `colorize_nicks.py` script, which does this.

> I was surprised that there was a setting to force all user mentions to a specific color.

The purpose of this is to be able to see that it's an actual slack mention, rather than just text as it will only highlight the user if it's a mention.

> How about falling back to the nick color if `user_mention` is unset or forcibly set to an empty string?

`user_mention` is a color option, so it's not possible to set it to an empty string. We can allow it to be `null` though, so perhaps we could use that to mean that the nick colors should be used.

> Side note, the `slack.color.user_mention` setting is new (i.e. only in wee-slack 3.0.0 rework).

Yeah, the old version doesn't do any coloring of mentions, so it's impossible to tell if it's a mention or just text.

> About your remark on your own nick being the same color than default text. You should have a highlight rule that makes that particular message highly visible anyway. So the mention color does not matter.

As mentioned above, the mention color matters IMO to know that it's a mention even if you have the highlight color in the prefix.

rjarry:
> WeeChat doesn't do that by default for IRC buffers. By default they're printed without any color. I assume you're using the `colorize_nicks.py` script, which does this.

Indeed I have this script. I had forgotten about it.

> `user_mention` is a color option, so it's not possible to set it to an empty string. We can allow it to be `null` though, so perhaps we could use that to mean that the nick colors should be used.

That would be nice. Otherwise, this new setting makes `colorize_nicks.py` not work on slack buffers.

I will dig around to see how I can change this setting to be nullable.

trygveaa:
> That would be nice. Otherwise, this new setting makes `colorize_nicks.py` not work on slack buffers.

I wouldn't think that script worked well on slack buffers either way? Any time a message is edited or reacted to, the color set by that script would be removed.

rjarry:
Indeed the script chokes after messages are edited.

trygveaa:
I added a new option `slack.color.user_mention_nick_color` to control if the nick color should be used as the mention color. If this option is on the `slack.color.user_mention` option will be removed.

PR #944, opened 2024-07-17
Add option to see the username as the nick

rodarima:
When `use_usernames` is true, the username is preferred over the full name or the display name.

Example: `JohnDoe` → `jdoe`

### DIFF
--- a/docs/Options.md
+++ b/docs/Options.md
@@ -286,3 +286,9 @@ slack` after changing it to take effect.
 
 **Description:** Use full names as the nicks for all users. When this is false (the default), display names will be used if set, with a fallback to the full name if display name is not set.
 
+### use_usernames
+
+**Default:** `false`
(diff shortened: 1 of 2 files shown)


trygveaa:
Hm, why do you want to use usernames? The usernames have been auto generated for slack users created in the last years, and it's not possible to set it when creating an account or change it later.

Is it because usernames are unique, unlike display names and full names?

rodarima:
> Hm, why do you want to use usernames?

Because they are shorter and in my organization they match the LDAP username.

trygveaa:
Ah, okay. I  updated Options.md and changed it to merge to the 2.x branch, as I'm moving development of the current version there (master is now the upcoming 3.0.0 version, which is a mostly complete rewrite).

Thanks! And sorry for the delay.

Issue #942, opened 2024-07-17
v3: Handle some `/` commands in slack buffers


rjarry:
Weechat `/` commands (e.g. `/join`, `/invite`, `/query`) should be implemented in slack buffers.

PR #945, opened 2024-07-21
Alias slack commands with IRC commands

rjarry:
Alias usual IRC commands (`/query`, `/join`, `/me`, etc.) with slack callbacks. Also add the missing `meMessage` command.

Closes: https://github.com/wee-slack/wee-slack/issues/942


slack/commands.py
```diff
@@ -813,6 +819,16 @@ def register_commands():
                 get_callback_name(command_cb),
                 cmd,
             )
+        if command.alias:
+            weechat.hook_command(
```
trygveaa, line 823: Hm, I think we perhaps should use `hook_command_run` instead. By using `hook_command` with the same names as irc uses, it complains that the command exists in multiple plugin if you run it from the core buffer, so it becomes impossible to run e.g. `/query -server libera nick` from the core buffer.

And then my thought has been to implement these commands as similarly as possible to the irc commands, so the arguments and options are the same.

I see that the `away` command is actually defined in the core plugin, and the irc plugin uses `hook_command_run`.

rjarry: I tried using `hook_command_run` but I keep getting errors:

```
[rjarry] /me goes back to sleep
slack: Error with command "/ /me goes back to sleep" (help on command: /help )
```

Also, if I understand correctly, hook_command_run() only works if overriding an existing command. What would happen if the IRC plugin isn't enabled?

For now, I have left things the way they were.

rjarry:
I had to change `chat.meMessage` to the undocumented [chat.command](https://github.com/ErikKalkoken/slackApiDoc/blob/master/chat.command.md) method. Apparently, only `chat.command` works with enterprise tokens:

```
13:46 =!= slack: Task('656dd77d-4ea8-465f-ba6b-219e006beef8', coroutine=command_slack_memessage) failed with: Error from Slack API method chat.meMessage with request {'channel': 'C069DHL58H4', 'text': 'goes back to sleep'} for workspace foobar: {'ok': False, 'error': 'enterprise_is_restricted'} (run `/slack debug error 8b6b1c05-c014-4a76-8b3f-284dc2f731c3` for the stack trace)
```

slack/commands.py
trygveaa: `hook_command` creates a new command which means it will get help text and completion and be tied to the plugin that created the command.

`hook_command_run` will just allow you to hook into any time anything starting with the command character is run by the user, regardless of the defined commands.

When something starting with a command character is run in WeeChat it will first look at all the `hook_command_run` hooks. These will be executed in the order defined by the priority ([see hook priority in the documentation](https://weechat.org/files/doc/stable/weechat_plugin_api.en.html#hook_priority)) regardless of if the commands defined. If any of them returns `WEECHAT_RC_OK_EAT` it will stop and not run any further hooks (neither `hook_command` or `hook_command_run`). As long as they return `WEECHAT_RC_OK` it will continue to the next `hook_command_run` hook.

When all the `hook_command_run` hooks have run and if all of them returned `WEECHAT_RC_OK` it will check if there is a defined command. If there isn't it will print an error that the command doesn't exist. If the command is defined only once in one plugin, it will run the hook for that command. If the command is defined in multiple plugins it will check if the current buffer is created by any of those plugins. If it is, it will run the hook from that plugin. If the plugin for the current buffer didn't define a command it will print an error that the command is defined in multiple plugins, but not in the current plugin.

Additionally, you can set a priority when defining a command. If you don't set this, you are not allowed to create the same command in the same plugin multiple times. If you do you can create it multiple times however, and they will be run in order of priority regardless of the buffer it was ran from (as long as the priority is different).

Note that I say plugin and not script as these things are tied to the plugin, which would be the python plugin in the case of python scripts. This means e.g. that multiple python scripts can't create the same command without specifying priority.

> I tried using hook_command_run but I keep getting errors:

I can't say for sure what the problem was without seeing your code, but note that the arguments sent to the callback is different for the two commands. With `hook_command_run` you get the whole command line, including the command.

> Also, if I understand correctly, hook_command_run() only works if overriding an existing command. What would happen if the IRC plugin isn't enabled?

`hook_command_run` works the same way regardless of if the command is defined or not, so it will still work even if the IRC plugin isn't enabled. However, the completion won't work and there won't be any help text.

After thinking some more while writing this, maybe what we should do is define it both with `hook_command` with a low priority and with `hook_command_run`. This will make the completion and help text work regardless of the IRC plugin, while sending the command to IRC from IRC buffers and to wee-slack from wee-slack buffers. And it will be sent to the IRC plugin if loaded when run from other buffers like the core buffer, and to wee-slack if the IRC plugin isn't loaded. I'll check if this works as I think it would.

For the away command it should only be defined with `hook_command_run` run though because this command is defined in core, and the IRC plugin also uses `hook_command_run` for this.

I can do these changes later though. I've kept this PR open for way to long, so I'll just do some testing and then I think I can merge it as is.

trygveaa:
Thanks! I refactored the memessage code a bit, and fixed a few issues (a type error, `command_cb` returning `WEECHAT_RC_OK_EAT` instead of `WEECHAT_RC_OK` and the alias commands missing completion).

And sorry for the very long delay in reviewing/merging this.

